PR #11, opened 2021-06-07
Add the ability to send data to specific participants

FeepsDev:
The title says it all

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -21,6 +21,7 @@ import LocalVideoTrack from '../track/LocalVideoTrack';
 import { TrackPublishOptions } from '../track/options';
 import { Track } from '../track/Track';
 import Participant from './Participant';
+import RemoteParticipant from './RemoteParticipant';
 
 const simulcastMinWidth = 200;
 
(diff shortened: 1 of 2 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -231,24 +232,31 @@ export default class LocalParticipant extends Participant {
...
    *
    * @param data Uint8Array of the payload. To send string data, use TextEncoder.encode
    * @param kind whether to send this as reliable or lossy.
    * For data that you need delivery guarantee (such as chat messages), use Reliable.
    * For data that should arrive as quickly as possible, but you are ok with dropped
    * packets, use Lossy.
+   * @param destination the participants who will receive the message
    */
-  publishData(data: Uint8Array, kind: DataPacket_Kind) {
+  publishData(data: Uint8Array, kind: DataPacket_Kind, destination: RemoteParticipant[] = []) {
```
davidzhao, line 244: It would be more flexibile if it could take in an sid array instead of RemoteParticipant, or perhaps either option.
```suggestion
  publishData(data: Uint8Array, kind: DataPacket_Kind, destination?: RemoteParticipant[] | string[]) {
```

src/room/participant/LocalParticipant.ts
```diff
@@ -231,24 +232,31 @@ export default class LocalParticipant extends Participant {
...
     if (data.length > 15_000) {
       throw new PublishDataError('data cannot be larger than 15k');
     }
 
+    const dest: string[] = [];
+    for(const rp of destination){
```
davidzhao, line 250:
```suggestion
    destination.forEach((val) => {
      if (val instanceof RemoteParticipant) {
        dest.push(val.sid)
      } else {
        dest.push(val)
      }
    })
```

davidzhao (approved):
lgtm!

PR #20, opened 2021-07-29
Set src to empty after detach

davidzhao:
In order to clean up Chrome's media players, srcObj/src needed to be set to empty. This will help keep memory usage down as well as avoid breakage when chrome [implements media player limits](https://chromium-review.googlesource.com/c/chromium/src/+/2816118) (again).

### DIFF
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -258,6 +258,7 @@ window.toggleVideo = async () => {
   if (videoTrack) {
     appendLog('turning video off');
     currentRoom.localParticipant.unpublishTrack(videoTrack);
+    videoTrack.detach();
     videoTrack = undefined;
     const video = getMyVideo();
     if (video) video.remove();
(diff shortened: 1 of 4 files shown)


hn8:
Is it a good idea to put empty media element back to a pool (with a size limit)? For example, Safari (stricter than Chrome) has the codepath to remember whether media element is created in user gesture or not. So it is nice to recycle those valuable audio elements to be reused later. The Safari edge case may happen when users only subscribe but not publish audio tracks (like broadcast or webinar).

davidzhao:
> Is it a good idea to put empty media element back to a pool (with a size limit)? For example, Safari (stricter than Chrome) remembers whether media element is created in user gesture or not. So it is nice to recycle those user-activated audio elements to be reused later. The Safari edge case may happen when users only subscribe but not publish audio tracks (like broadcast or webinar), so JS need to ask user gesture to create at least the first audio element, and as long as there is 1 audio playing, newer audio elements can play without user gesture,

You are right, Safari is pretty annoying with this. I just confirmed that this is currently broken if you have a user that hasn't called `getUserMedia`. We'll need to handle this case explicitly and I'll update the docs as well. Will create a separate PR to handle recycling of audio elements and the explicit trigger

frostbyte73 (approved):

PR #22, opened 2021-07-29
added handling for browser suppression of audio autoplay

### LINKED ISSUE #21: Audio doesn't playback on Safari unless getUserMedia is called
Safari requires explicit user consent to play audio on both Mac and iOS. Because our audio elements are created without explicit user interaction, they would not playback unless the user is also publishing their tracks. 

References:

* https://webkit.org/blog/7734/auto-play-policy-changes-for-macos/
* https://webkit.org/blog/6784/new-video-policies-for-ios/

davidzhao:
See README for details, fixes #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,30 @@ videoTrack.mute();
 room.localParticipant.unpublishTrack(videoTrack);
 ```
 
+### Audio playback
+
+Browsers can be restrictive regarding if audio could be played without user interaction. What each browser considers as user interaction can also be different (with Safari on iOS being the most restrictive). Some browser considers clicking on a button unrelated to audio as interaction, others require audio element's `play` function to be triggered by a onclick event.
+
+LiveKit will attempt to autoplay all audio tracks when you attach them to audio elements. However, if that fails, we'll notify you via `RoomEvent.AudioPlaybackStatusChanged`. `Room.canPlayAudio` will indicate if audio playback is permitted. (Note: LiveKit takes an optimistic approach so it's possible for this value to change from `true` to `false` when we encounter a browser error.
+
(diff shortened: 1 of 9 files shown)


frostbyte73 (approved):

README.md
```diff
@@ -107,6 +107,30 @@ videoTrack.mute();
 room.localParticipant.unpublishTrack(videoTrack);
 ```
 
+### Audio playback
+
+Browsers can be restrictive regarding if audio could be played without user interaction. What each browser considers as user interaction can also be different (with Safari on iOS being the most restrictive). Some browser considers clicking on a button unrelated to audio as interaction, others require audio element's `play` function to be triggered by a onclick event.
```
OmriSama, line 112: Would rephrase:

> Browsers can be restrictive with regards to audio playback that is not initiated by some user interaction.
> What each browser considers a 'user interaction' can vary by vendor (with Safari on iOS being the most restrictive).
> Some browsers consider clicking on a button unrelated to audio as interaction, others require an `<audio>` element's `play` function to be triggered by an `onClick` event.

hn8: Actually even the strictest Webkit doesn't require user gesture in play if the media element is created in user gesture first.
https://github.com/WebKit/WebKit/blob/main/Source/WebCore/html/HTMLMediaElement.cpp#L440

Also user activation could even propogate through postMessage or setTimeout and not just limited to onclick
https://whatpr.org/html/3851/interaction.html#tracking-user-activation

But it is inconsistent across browsers, Chrome only whitelists 4 gestures
https://github.com/chromium/chromium/blob/master/content/browser/web_contents/web_contents_impl.cc#L239

davidzhao: thanks all for the suggestions. updated.

PR #27, opened 2021-08-10
Reduce audio lag on safari with AudioContext magic

davidzhao:


dsa (approved):
LG

hn8:
AudioContext needs to be created or resumed in user gesture. Extra bonus: Firefox would not throttle setTimeout with active AudioContext

PR #29, opened 2021-08-11
LocalParticipant: Set server track mute status based on local track

### LINKED ISSUE #28: track cannot be muted before/immediately after publishing
Steps to reproduce:
1. Create your tracks using `createLocalTracks`
2. Mute the audio track that you have created using `audioTrack.mute()`, leave the vide track unmuted
3. Connect to your livekit server using `connect()` and pass the existing tracks
4. Connect another client to the server - observe that audio is correctly not transmitting
5. Mute your video track using `videoTrack.mute()`
6. Your audio will start transmitting.

(This will also happen if immediately muting a track after the connect has succeeded).

I believe this is being caused by the fact that the mute state does not appear to be sent to the engine when initially adding a track. So, when the video track gets muted `LocalParticipant.updateInfo` is being called which includes: https://github.com/livekit/client-sdk-js/blob/63e0be4f48b6650f936568ab31b6da6162be3197/src/room/participant/LocalParticipant.ts#L222-L234 This unmutes the local track.

This can be tested with the sample code by adding the mute step to this part of the code:
https://github.com/livekit/client-sdk-js/blob/63e0be4f48b6650f936568ab31b6da6162be3197/example/sample.ts#L225-L229

Making it:
```js
    } else if (publication.kind === Track.Kind.Audio) {
      // skip adding local audio track, to avoid your own sound
      // only process local video tracks
      audioTrack = <LocalAudioTrack>publication.track;
      appendLog('muting audio');
      audioTrack.mute();
    }
```

I ran into this when working on an app that saves the user's mute state between sessions, so a track is immediately muted before/during connection if that is the state they last used.

Two requests/suggestions:
1. Reverse the logic of `LocalParticipant.updateInfo` to push the local state to the server instead of pulling local state from the server. I don't think the server mute state should ever override the local mute state of a track.
2. Push the mute state of a track to the server at the time it is added.

I can try to make a PR for this soon, but I'm still trying to figure out the best way to pass the mute state during the track add step.

bekriebel:
Instead of changing the local track mute status when the server status
does not match, update the server's status to the local state. This
prevents a local track from getting unmuted without the user's action
when the states do not match.

Fixes #28

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -230,17 +230,15 @@ export default class LocalParticipant extends Participant {
   updateInfo(info: ParticipantInfo) {
     super.updateInfo(info);
 
-    // match local track mute status to server
+    // match server track mute status to local
     info.tracks.forEach((ti) => {
       const pub = <LocalTrackPublication> this.tracks.get(ti.sid);
       if (!pub) {
         return;
       }
 
-      if (ti.muted && !pub.isMuted) {
-        pub.mute();
-      } else if (!ti.muted && pub.isMuted) {
-        pub.unmute();
+      if (ti.muted !== pub.isMuted) {
+        this.engine.updateMuteStatus(ti.sid, pub.isMuted);
       }
     });
   }


bekriebel:
Though the original issue I was having is fixed with other server & client changes, I still think that this change or something like it is important. For mute status, the local client should be the source of truth to avoid a user's tracks from getting unmuted without their knowledge or consent.

davidzhao:
@bekriebel I'm closing this ticket since your needs are resolved by the server-side fix. lmk if you are having any other issues.

PR #40, opened 2021-09-02
handle explicit remote unmute, removed mute status match

### LINKED ISSUE #38: Quickly muting and unmuting audio causes a muting loop
If you quickly toggle the mute state of an audio track, the track goes into a state of muting and unmuting until disconnecting from the server. I'm not sure if this issue is in the client or server code.

Reproduction steps with the sample client:
1. Using the example/sample, add a set of hooks to log mute and unmute room events
```ts
    .on(RoomEvent.TrackMuted, () => appendLog("Track muted"))
    .on(RoomEvent.TrackUnmuted, () => appendLog("Track unmuted"))
```
2. Launch the sample and connect two clients.
3. Open the console on a client and quickly mute and unmute the audio track.
```ts
window.muteAudio(); window.muteAudio();
```
4. Observe from the second client that the audio track continues to get muted and unmuted.

The console command is the most reliable way to trigger it, but I have had it trigger in a real scenario with a push-to-talk setup where the user presses and immediately releases or accidentally double taps the push-to-talk key. This can be mitigated by adding a delay between mute calls in a client implementation, but it seems like something is getting into a bad state when a mute/unmute call are done back to back.

davidzhao:
fixes #38

### DIFF
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -1,6 +1,5 @@
 import {
-  connect, createLocalScreenTracks, CreateVideoTrackOptions,
-  LocalAudioTrack,
+  connect, createLocalScreenTracks, CreateVideoTrackOptions, DataPacket_Kind, LocalAudioTrack,
   LocalTrack,
   LocalVideoTrack,
   LogLevel,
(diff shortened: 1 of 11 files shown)


bekriebel:
Thanks! I should be able to test this tomorrow, but it'll be later in the day before I'll get the chance.

bekriebel (approved):
I'm only okay at the interface between this and golang, so I can't speak to the code quality. However, I tested this with my "mute loop" tests  for #38 using my client implementation and it appears to fix my issue. Thanks!

PR #41, opened 2021-09-08
Use subscriber connection as primary

davidzhao:
Implements client protocol version 3 - subscriber primary peer connections. Designed to reduce publisher connection load when the client is in read-only mode.

* rely on subscriber connection as the primary PeerConnection
* backwards compatible with previous server versions (and uses publisher as primary)
* receive data messages on subscriber PC


frostbyte73 (approved):
lgtm

PR #44, opened 2021-09-19
Still emit event even participant is null

hiroshihorie:
userPacket.participantSid could be null for userPackets set by Server-API.

Would this work as intended or should we check against `userPacket.participantSid` ?


davidzhao (approved):
lgtm!

PR #46, opened 2021-09-28
Stricter tsconfig

hn8:
Enable noUnusedLocals

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -30,10 +30,12 @@ export default class RTCEngine extends EventEmitter {
 
   private lossyDC?: RTCDataChannel;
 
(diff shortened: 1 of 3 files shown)


davidzhao:
hey Hann, what's the benefit of `importsNotUsedAsValues` ?

hn8:
@davidzhao It prevents accidentally non-type import (increase build size). But since LiveKit JS only has 1 build output, it is only for explicit "documentation" purpose. Will revert that commit if it is not worth extra check.

hn8:
@davidzhao I reverted noImplicitReturns and importsNotUsedAsValues to minimize changes. Keep noUnusedLocals check which is similar to Go.

davidzhao:
thanks! this looks good.

PR #51, opened 2021-10-14
Reuse transceivers in subscribe path

boks1971:
LK-22 (https://linear.app/livekit/issue/LK-22/feature-pooled-transceivers-to-avoid-re-negotiation)

Do not stop remote tracks. That would change `readyState` of track
to `ended` and the corresponding transceiver cannot be reused.
Instead use the `enabled` attribute of track to enable/disable
remote tracks.

Bumping up the protocol version that informs the server that
client is capable of reusing transceivers.

Testing:
--------
Connect from Chrome. For remote side, connect/disconnect
from Firefox several times and ensure the following
- SDP does not grow (it has 4 sections as reported by
chrome://webrtc-internals (version, data, audio, video) each
time the remote side connects.
- Media flows from remote side on every connection.

Oddities:
---------
- Chrome callback `ontrack` sends a track. When the remote side
reconnects and the same transceiver is used, the callback does
not contain the new track id. It retains what was returned in the
first callback. Don't think it is an issue as we do not use that
track id anywhere.

### DIFF
--- a/src/room/track/Track.ts
+++ b/src/room/track/Track.ts
@@ -110,11 +110,11 @@ export class Track extends EventEmitter {
     this.mediaStreamTrack.stop();
   }
 
-  enable() {
+  protected enable() {
     this.mediaStreamTrack.enabled = true;
   }
 
-  disable() {
+  protected disable() {
     this.mediaStreamTrack.enabled = false;
   }
 


boks1971:
@davidzhao Yes I verified that the stream id and the parsing gets the new information. I will verify it again just to be doubly sure.

PR #53, opened 2021-10-14
Device management API and simpler APIs to work with common tracks

davidzhao:


davidzhao:
Adding a few folks on this to ensure I get this right. My hope is to converge to something like across platforms, making it easier to work with common track types.

Most of the device management APIs only apply to the browser, but I wonder how much of it make sense on mobile platforms too (especially for audio management). Ideally we can converge there too.

PR #62, opened 2021-10-26
emit unpublish event also for local participant

lukasIO:
but only if it was caused by mediatrack ended event.

Reasoning: 
When screensharing is stopped via browser UI, it comes in handy to subscribe to `TrackUnpublished`

```ts
this.currentRoom.localParticipant.on(ParticipantEvent.TrackUnpublished, async (track) => {
			log.debug('local track unpublished (most likely by browser), stop screen share');
```

the naming and default behaviour is mirrored from `RemoteParticipant.ts`


lukasIO:
closing in favor of #72

PR #70, opened 2021-10-29
override track name if set via options

lukasIO:
Right now setting the `name` property in the track publish option like 
```ts
	await this.currentRoom.localParticipant.publishTrack(audioTrack, {
					name: `my-microphone`
				});
```

doesn't have any effect if the track is not an instance of `MediaStreamTrack`.
This PR overrides the track name if it is set explicitly via options also for a `LocalTrack` type.


davidzhao (approved):
great catch!

PR #71, opened 2021-10-29
add screenshareaudio track source

lukasIO:
Right now screen share audio has an "unknown" source type set. 
I found myself needing to know specifically if it is audio from a screen share, thus this PR.

### DIFF
--- a/src/room/participant/Participant.ts
+++ b/src/room/participant/Participant.ts
@@ -82,13 +82,16 @@ export default class Participant extends EventEmitter {
         return pub;
       }
       if (pub.source === Track.Source.Unknown) {
-        if (source === Track.Source.Microphone && pub.kind === Track.Kind.Audio) {
+        if (source === Track.Source.Microphone && pub.kind === Track.Kind.Audio && pub.trackName !== 'screen') {
           return pub;
         }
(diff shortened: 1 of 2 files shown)


davidzhao:
> Hm, thinking about it more, it might also suffice to just set the source as `ScreenShare` also for screen audio and differentiate between screen video and audio by looking at `Track.Kind`.

Yeah I think that's true.. though we'd need our helper functions to perform filtering correctly. I think that's better technically.. because the source of the stream is from the screen share.

lukasIO:
@davidzhao are there any other helper functions apart from `getTrack` that need to adjust their filtering for screenshare audio?

davidzhao:
hey @lukasIO I'm thinking about this a bit more. It does seem like this could be kind of confusing.. It was nice before to have a one to one relationship between TrackSource -> Track

now this introduces another dimension to consider... So my apologies.. but it's better to add a unique source for this. I can add this to the protocol layer so clients could get the generated protobufs.

lukasIO:
I had the same feeling when writing it out... 
Would be great if you could add it in the protocol, I'll adjust the naming here (or feel free to do it yourself) if you want to change the event name.

davidzhao:
@lukasIO protobufs and helper methods have been added.

davidzhao (approved):
lg!

PR #72, opened 2021-10-29
Emit LocalTrackUnpublished event when a local track is unpublished

davidzhao:
When a local track ends, sometimes it's due to browser sending an `ended` event on the MediaStreamTrack. For example, if user interacted with browser UI to stop publishing.

This is another attempt at #62. We've decided to use a different event from `TrackUnpublished` to be more explicit about the action. Since majority of the users don't have to pay attention to `TrackUnpublished` when using auto subs.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from 'events';
-import log from '../logger';
 import { SignalClient, SignalOptions } from '../api/SignalClient';
+import log from '../logger';
 import {
   DataPacket_Kind, ParticipantInfo,
   ParticipantInfo_State, Room as RoomModel, SpeakerInfo, UserPacket,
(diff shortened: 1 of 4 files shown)


src/room/Room.ts
```diff
@@ -174,9 +175,12 @@ class Room extends EventEmitter {
         .on(ParticipantEvent.TrackUnmuted, (pub: TrackPublication) => {
           this.emit(RoomEvent.TrackUnmuted, pub, this.localParticipant);
         })
-        .on(ParticipantEvent.LocalTrackPublished, (pub: TrackPublication) => {
+        .on(ParticipantEvent.LocalTrackPublished, (pub: LocalTrackPublication) => {
           this.emit(RoomEvent.LocalTrackPublished, pub, this.localParticipant);
         })
+        .on(ParticipantEvent.LocalTrackUnpublished, (pub: LocalTrackPublication) => {
+          this.emit(ParticipantEvent.LocalTrackUnpublished, pub, this.localParticipant);
```
lukasIO, line 182: For consistency it seems like the emitted event should be a `RoomEvent`

davidzhao: yes! great catch.

PR #73, opened 2021-11-01
feat: added connection status feature

### LINKED ISSUE #59: [Feature] callback event for Stats
It will be best if it was possible to get live Stats by callback event https://github.com/livekit/client-sdk-js/blob/1532080948c2722cbcf40a35fe31529fa361c815/src/room/track/LocalVideoTrack.ts#L82
This way we can detect user's connection status & give warning or suggestion for their slow connection. The current implementation only working if `simulcast` enabled. Will be better if we can get some way to get status for any type connection.

jibon57:
Connection status monitor based on webrtc stats (inbound-rtp and outbound-rtp). I've followed from [bigbluebutton-connection-status](https://github.com/bigbluebutton/bigbluebutton/blob/706d9a1c5a5e9b43199eb01b5dc43b179aaf49d8/bigbluebutton-html5/imports/ui/components/connection-status/service.js#L461). I've tested from firefox & chrome. Both worked as expected. User can get outbound (upload) and inbound (download) rates (kbps), jitter and packetsLost 

Close #59

### DIFF
--- a/src/connect.ts
+++ b/src/connect.ts
@@ -39,7 +39,6 @@ export async function connect(
   options.logLevel ||= LogLevel.info;
   if (options.audio === undefined) options.audio = false;
   if (options.video === undefined) options.video = false;
-  if (options.enableConnectionMonitor === undefined) options.enableConnectionMonitor = false;
 
   log.setLevel(options.logLevel);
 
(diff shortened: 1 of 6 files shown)


davidzhao:
hey @jibon57, appreciate your contribution here!

Connection quality reporting is something we had been planning for awhile. With an SFU, its actually quite a bit easier to perform this detection on the server side. In addition to getting the current participant's connection information, we'd also want to know the quality of connection to others we are interacting with.

We've implemented connection quality APIs in #75. At the moment, we will not expose the raw data to the clients as this is quite a bit simpler for clients to use.

jibon57:
@davidzhao no problem!! The other solution is better. I'm closing it. Not sure if it possible to show download & upload speed as well as. This will be helpful for user to check their internet speed.

PR #74, opened 2021-11-02
wait until DataChannel is ready to send message

jibon57:
We'll check if status is `connecting`. In that case wait 1 second before sending message. 

Fixed: https://github.com/livekit/livekit-server/issues/144

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -389,22 +389,16 @@ export default class RTCEngine extends EventEmitter {
     const msg = DataPacket.encode(packet).finish();
 
     // make sure we do have a data connection
-    await this.ensurePublisherConnected();
+    await this.ensurePublisherConnected(kind);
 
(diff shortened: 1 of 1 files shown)


src/room/RTCEngine.ts
```diff
@@ -392,8 +392,14 @@ export default class RTCEngine extends EventEmitter {
     await this.ensurePublisherConnected();
 
     if (kind === DataPacket_Kind.LOSSY && this.lossyDC) {
+      if (this.lossyDC.readyState === 'connecting') {
```
davidzhao, line 395: It'd be better to move this check to `ensurePublisherConnected` [here](https://github.com/livekit/client-sdk-js/blob/main/src/room/RTCEngine.ts#L416)

jibon57: Yes, that's better. Thank you!

davidzhao (approved):
looks good!

PR #75, opened 2021-11-03
Connection quality APIs

davidzhao:
See server PR: https://github.com/livekit/livekit-server/pull/167

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -5,6 +5,7 @@ import {
   DataPacket_Kind, ParticipantInfo,
   ParticipantInfo_State, Room as RoomModel, SpeakerInfo, UserPacket,
 } from '../proto/livekit_models';
+import { ConnectionQualityUpdate } from '../proto/livekit_rtc';
 import {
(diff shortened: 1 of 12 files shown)


davidliu (approved):
LGTM

hiroshihorie (approved):
🚀

jibon57:
This solution is better then mine. Not sure if it possible to show download & upload speed as well as. This will be helpful for user to check their internet speed.

src/room/Room.ts
```diff
@@ -116,6 +117,8 @@ class Room extends EventEmitter {
       this.state = RoomState.Connected;
       this.emit(RoomEvent.Reconnected);
     });
+
+    this.engine.on(EngineEvent.ConnectionQualityUpdate, this.handleConnectionQualityUpdate);
```
lukasIO, line 121: Wouldn't this effectively override any quality setting that has been set manually?
In our usecase we need to decide the quality not only on connection speed but also depending on other parameters.

PR #77, opened 2021-11-07
Automatically manage remote video track layer

davidzhao:
Opt-in feature for now. When enabled, the client will automatically request an appropriately sized layer that matches the attached video elements. It'll also pause the stream when the object becomes invisible, without requiring explicit management from client apps.

### DIFF
--- a/src/room/track/RemoteTrackPublication.ts
+++ b/src/room/track/RemoteTrackPublication.ts
@@ -1,3 +1,4 @@
+import log from '../../logger';
 import { TrackInfo } from '../../proto/livekit_models';
 import {
   UpdateSubscription,
@@ -5,17 +6,23 @@ import {
   VideoQuality,
 } from '../../proto/livekit_rtc';
 import { TrackEvent } from '../events';
+import { Track } from './Track';
 import TrackPublication from './TrackPublication';
 import { RemoteTrack } from './types';
 
 export default class RemoteTrackPublication extends TrackPublication {
   track?: RemoteTrack;
 
+  /** @internal */
+  autoManageVideo?: boolean;
+
   protected subscribed?: boolean;
 
   protected disabled: boolean = false;
 
-  protected currentVideoQuality: VideoQuality = VideoQuality.HIGH;
+  protected currentVideoQuality?: VideoQuality = VideoQuality.HIGH;
+
+  protected videoDimensions?: Track.Dimensions;
 
   /**
    * Subscribe or unsubscribe to this remote track
@@ -49,7 +56,7 @@ export default class RemoteTrackPublication extends TrackPublication {
    * @param enabled
    */
   setEnabled(enabled: boolean) {
-    if (this.disabled === !enabled) {
+    if (this.autoManageVideo || this.disabled === !enabled) {
       return;
     }
     this.disabled = !enabled;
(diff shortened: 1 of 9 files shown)


jibon57:
Excellent job @davidzhao

hiroshihorie (approved):
👏

src/room/track/RemoteVideoTrack.ts
```diff
@@ -18,6 +32,8 @@ export default class RemoteVideoTrack extends Track {
     // override id to parsed ID
     this.sid = sid;
     this.receiver = receiver;
+    this.intersectionObserver = new IntersectionObserver(this.handleVisibilityChanged);
```
lukasIO, line 35: Instead of creating an own IntersectionObserver for each RemoteVideoTrack, it would probably be more performant to use a shared IntersectionObserver (maybe on Room level or as a singleton). I'm not sure how big the impact would be for a couple of video elements and if it's worth it, but when implementing lazy-loading where potentially hundreds of image elements get observed, the difference becomes noticable. (https://www.bennadel.com/blog/3954-intersectionobserver-api-performance-many-vs-shared-in-angular-11-0-5.htm)
The same might be true for ResizeObserver.

src/room/track/RemoteTrackPublication.ts
```diff
@@ -1,21 +1,28 @@
...
 import { TrackEvent } from '../events';
+import { Track } from './Track';
 import TrackPublication from './TrackPublication';
 import { RemoteTrack } from './types';
 
 export default class RemoteTrackPublication extends TrackPublication {
   track?: RemoteTrack;
 
+  /** @internal */
+  autoManageVideo?: boolean;
```
lukasIO, line 17: is it intended that users can change the flag also after the TrackPublication and it's corresponding MediaTrack have been created?
I'm only asking because setting the `autoManageVideo` option (additionally) on `RemoteVideoTrack` would allow to actually opt out of instantiating the observers all together (or not calling observe in `attach`)
Which in turn would make all the `if (this.autoManageVideo) return` checks in `RemoteVideoPublication` redundant.

src/room/track/RemoteVideoTrack.ts
davidzhao: @lukasIO this is a great insight! thanks for sharing it.

I'll modify this to use a shared instance instead.

src/room/track/RemoteTrackPublication.ts
davidzhao: I think this would be a room-level setting, where you couldn't change it on a track by track basis. Instead of this flag though, I'll pass down the shared observer instances and use those to determine eligibility

src/room/track/RemoteVideoTrack.ts
davidzhao: actually reading this a bit more. this is quite complicated to share an instance.. because currently it's nicely scoped to a single `RemoteVideoTrack`. from the post, the author wasn't sure the source of the speedup with sharing a single instance, and suspected it could be angular.

```
it could be that the performance bottleneck is actually the Angular template reconciliation and the number of change-detection digests.
```

Because of the added complexity without clear payoff, I'm going to leave it the way it is for now. We can always optimize in a future PR.

davidzhao: Though I did remove the observers when this is not in use. Just so that we are not adding unnecessary processing.

boks1971 (commented):
Excellent stuff @davidzhao . This will be magical!

PR #78, opened 2021-11-08
use shared observers for auto managed video

lukasIO:
an attempt at using shared observers. 
One downside is that one InteractionObserver and one ResizeObserver will be created even if autoManageVideo is false.
but at least they will never observe any elements in that case, so the impact should be negligible.

not sure about where to put the `ObservableMediaElement` interface. Putting it into `types` resulted in a dependency loop.

reference for implementation: https://groups.google.com/a/chromium.org/g/blink-dev/c/z6ienONUb5A/m/F5-VcUZtBAAJ


davidzhao (approved):
really solid change @lukasIO!

PR #79, opened 2021-11-14
add option to keep local tracks running after disconnect

lukasIO:
currently whenever you disconnect from a room, all local tracks get stopped and detached from their HTMLElements.

This PR adds an option to keep the local tracks running after a disconnect. The implementation introduces `stopTracks` as a way to determine whether the localTracks should be stopped (and detached) or not. Defaults to `stopTracks = true` in order to not change the current behaviour.

Reasoning:
When managing local tracks independently from rooms it is desirable to persist them across room changes.
In a scenario where room switching should happen as smoothly as possible one would not want to attach and reattach local videotracks to html elements.

It's unfortunate that the event handler of `unbeforeunload` complains with the new boolean function property. 
To minimize code changes I added ts-ignore comments before the event handler assignments, because I couldn't think of a cleaner way to handle it.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -236,12 +236,12 @@ class Room extends EventEmitter {
   /**
    * disconnects the room, emits [[RoomEvent.Disconnected]]
    */
-  disconnect = () => {
-    // send leave
+  disconnect(): void;
+  disconnect(stopTracks = true): void {
     this.engine.client.sendLeave();
     this.engine.close();
-    this.handleDisconnect();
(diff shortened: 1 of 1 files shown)


lukasIO:
9f06ed0  adds a `onBeforeUnload` function in order to avoid the ts-ignore comments. 
Makes it a bit cleaner compared to the ts-ignore comments.

davidzhao (approved):
this is an elegant way to handle it. well done!

PR #80, opened 2021-11-14
fix shared resize observer for SSR frameworks

lukasIO:
When implementing #78 I unfortunately forgot about SSR frameworks (e.g. nextjs, nuxt, sveltekit).

I only tested now with sveltekit but the client library cannot be imported anymore (without workarounds) as sveltekit analyzes the library within nodejs and the implementation of #78 uses the browser-specific `ResizeObserver` and `IntersectionObserver` and executes them on import -> errors thrown within node.

This PR tries to address this. RO and IO are only instantiated when they are first called, thus circumventing the automatic instantiation on import.


davidzhao (approved):
Thanks for fixing this. I'm glad you caught it before we cut a release with it!

PR #83, opened 2021-12-04
Enable simulcast by default, add jest tests

davidzhao:


src/room/RTCEngine.ts
```diff
@@ -87,7 +87,7 @@ export default class RTCEngine extends EventEmitter {
   close() {
     this.isClosed = true;
 
-    if (this.publisher) {
+    if (this.publisher && this.publisher.pc.signalingState !== 'closed') {
```
noahlt, line 90: Style: could use optional chaining (`?.`) here

src/room/track/options.ts
```diff
@@ -27,7 +27,7 @@ export interface TrackPublishDefaults {
   dtx?: boolean;
 
   /**
-   * use simulcast, defaults to false.
+   * use simulcast, defaults to true.
    * When using simulcast, LiveKit will publish up to three version of the stream
```
noahlt, line 31: typo: "up to three version**s**"

src/room/defaults.ts
```diff
@@ -4,6 +4,8 @@ import {
 
 let publishDefaults: TrackPublishDefaults = {
   audioBitrate: AudioPresets.speech.maxBitrate,
+  dtx: true,
```
noahlt, line 7: defaulting `dtx` to `true` here appears to be a drive-by fix (because the [docs already say](https://github.com/livekit/client-sdk-js/blob/ac95ada/src/room/track/options.ts#L25) it defaults to true)

src/room/participant/LocalParticipant.ts
```diff
@@ -470,147 +470,151 @@ export default class LocalParticipant extends Participant {
...
+
+export const presetsScreenShare = [
+  ScreenSharePresets.vga,
+  ScreenSharePresets.hd_8,
+  ScreenSharePresets.hd_15,
+  ScreenSharePresets.fhd_15,
+  ScreenSharePresets.fhd_30,
+];
+
+export function computeVideoEncodings(
```
noahlt, line 499: ❓ this is no longer a private member method; I assume this change was necessary to make it testable?

(I'm a little surprised Jest doesn't have a faculty for testing private methods, but then again it's a newfangled feature)

src/room/track/options.ts
noahlt: It may be worth noting that this flag only has an effect if both `width` and `height` are set.

src/room/participant/LocalParticipant.ts
```diff
@@ -470,147 +470,151 @@ export default class LocalParticipant extends Participant {
...
-            maxBitrate: videoEncoding.maxBitrate,
-            /* @ts-ignore */
-            maxFramerate: videoEncoding.maxFramerate,
-          },
-        ];
-      }
+    // if resolution is high enough, we would send [q, h, f] res..
+    // otherwise only send [q, h]
+    // NOTE:
+    //   1. Ordering of these encodings is important. Chrome seems
```
noahlt, line 536: 😂

src/room/participant/LocalParticipant.ts
```diff
@@ -470,147 +470,151 @@ export default class LocalParticipant extends Participant {
...
-      const preset = presets[i];
-      if (width >= preset.width && height >= preset.height) {
-        encoding = preset.encoding;
-      }
-    }
+  return encodings;
+}
 
-    return encoding;
+export function determineAppropriateEncoding(
```
noahlt, line 591: ❓ Did you export this (and `presetsForResolution`, below) with the intention of testing them, as well?

Also, I'm not sure how our doc generation works, but if these are exported will they automatically show up in our public docs?

src/room/participant/LocalParticipant.ts
```diff
@@ -470,147 +470,151 @@ export default class LocalParticipant extends Participant {
...
 
-    if ((!videoEncoding && !useSimulcast) || !width || !height) {
-      // don't set encoding when we are not simulcasting and user isn't restricting
-      // encoding parameters
-      return undefined;
-    }
+  if (!videoEncoding) {
+    // find the right encoding based on width/height
+    videoEncoding = determineAppropriateEncoding(isScreenShare, width, height);
+    log.debug('using video encoding', videoEncoding);
```
noahlt, line 522: ❗ stray debug log

src/room/participant/LocalParticipant.test.ts
```diff
@@ -0,0 +1,48 @@
+import { VideoPresets } from '../..';
+import { computeVideoEncodings } from './LocalParticipant';
+
+describe('computeVideoEncodings', () => {
+  it('returns nil non-simulcast', () => {
+    const encodings = computeVideoEncodings(false, 640, 480, {
+      simulcast: false,
+    });
+    expect(encodings).toBeUndefined();
```
noahlt, line 9: ❓ What are the implications when `computeVideoEncodings` returns `undefined` here?  More specifically, I imagine if we return `undefined` here that some other part of the system (perhaps the browser) picks an encoding for us?

noahlt (commented):
Some questions and minor comments

src/room/RTCEngine.ts
davidzhao: are you thinking `this.publisher?.pc.signalingState !== 'closed'`? 

I thought about that but it wouldn't do what I want here: publisher.pc is defined *and* not closed.

src/room/track/options.ts
davidzhao: width and height are internally provided, where as this is options exposed to end users.

src/room/defaults.ts
davidzhao: yup.. it's already defaulting to true, just making it explicit for those that are reading this code.

src/room/participant/LocalParticipant.ts
davidzhao: right, doing this for testing.. I've read something about using rewire to accomplish this.. but it seemed too magical and I wanted to avoid introducing another dev dependency.

src/room/participant/LocalParticipant.ts
davidzhao: yup, coming up in a different PR, but great call on docs! I forgot about that. will mark them as internal.

src/room/participant/LocalParticipant.test.ts
davidzhao: the browser picks its defaults when encodings aren't defined.

src/room/RTCEngine.ts
noahlt: Ah, right, because with optional chaining when `this.publisher` is undefined it ends up comparing `undefined !== 'closed'`.

PR #85, opened 2021-12-06
Explicitly send video layers in use to SFU

davidzhao:
Implementing protocol defined in https://github.com/livekit/protocol/pull/52


hiroshihorie (approved):
✅

PR #97, opened 2021-12-26
unsubscribe from track event listeners when unpublishing

lukasIO:
I was reusing `LocalTrack`s across different rooms and ran into a memory leak:

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unmuted listeners added. Use emitter.setMaxListeners() to increase limit
    _addListener events.js:211
    addListener events.js:227
    client livekit-client.js:4364
    client livekit-client.js:4185
    client livekit-client.js:4167
    publishTrack livekit-client.js:4328
```

I realized that the event listeners are not unsubscribed when you unpublish a local track. 
I didn't check more closely if there are any other things that should get disposed when unpublishing, that could cause problems when reusing Tracks across rooms, but maybe worth investigating.


davidzhao (approved):

davidzhao:
looks good!

PR #98, opened 2021-12-27
set metadata on room connect

lukasIO:
seems like this was missing and lead to metadata always being undefined when connecting.


davidzhao:
great catch!

PR #101, opened 2021-12-30
Fixed data channel readiness check

davidzhao:


davidzhao:
@hiroshihorie @davidliu FYI, the correct way to check for data channel readiness.

hiroshihorie:
Got it!

Issue #106, opened 2022-01-02
LocalTrack doesn't remove event listener when unpublished


lukasIO:
https://github.com/livekit/client-sdk-js/blob/ae849e477a8a20f07f362313af057262badc0455/src/room/track/LocalTrack.ts#L19

Adds an 'ended' event to the underlying mediatstreamtrack that doesn't get removed.

Maybe a dispose function for the track that (for now) only removes this event listener and is called on track unpublish?

davidzhao:
The 'ended' listener is supposed to be there as long as the LocalTrack claims ownership over the MediaStreamTrack.

I think the only thing missing is this line in your PR

```
track.off(TrackEvent.Ended, this.onTrackUnpublish);
```

When unpublished, the Publication object should no longer care about events that LocalTrack is firing.

lukasIO:
There is currently one situation in which that line gets called and that's the `unpublishTrack `function of the `LocalParticipant`:

https://github.com/livekit/client-sdk-js/blob/f5e69ed6255aff01af1b900f6fb5f1d78ecc5e7f/src/room/participant/LocalParticipant.ts#L453

I used an override on the `LocalTrack `to subscribe and unsubscribe to the underlying mediatrack 'ended' event when the livekit `TrackEvent.Ended` is being subscribed/unsubscribed
https://github.com/livekit/client-sdk-js/blob/f5e69ed6255aff01af1b900f6fb5f1d78ecc5e7f/src/room/track/LocalTrack.ts#L127-L141

davidzhao:
I see that line.. but not convinced why this is necessary. A LocalTrack, even when unpublished, should listen to the underlying MediaStreamTrack events.

lukasIO:
Hm, I guess you are right. On top it now seems like a rather stupid idea of me to add the same event handler to the mediaStreamTrack each time TrackEvent.Ended is being subscribed to.

Then the only question remains when to actually remove the event listener from the mediaStreamTrack. 
Should I reconsider adding a `dispose` function to LocalTrack?

davidzhao:
I still think we'd only need to update listener to `TrackEvent.Ended` inside `LocalTrackPublication.setTrack` and `LocalParticipant.unpublishTrack`:

i.e.

```
track.off(TrackEvent.Ended, this.onTrackUnpublish);
```

MediaStreamTrack isn't supposed to be reused across other LocalTracks, as long as the LocalTrack itself is re-usable.

[...]

PR #107, opened 2022-01-02
Add local track publication dispose fn

lukasIO:
Follow up for #97 and addresses #106.

Wanted to track down all places where there were still event listeners that didnt get removed at one point. 
I hope I found all of them now. Tested with excessive connecting/disconnecting to different livekitrooms and reusing the local tracks for that. 

One actual change would be that on disconnect the local tracks get unpublished now. I'm not sure if this has further implications, but it seemed to make sense to me and made it cleaner to unsubscribe from some eventlisteners.


src/room/Room.ts
```diff
@@ -394,12 +394,17 @@ class Room extends EventEmitter {
         p.unpublishTrack(pub.trackSid);
       });
     });
-    if (shouldStopTracks) {
-      this.localParticipant.tracks.forEach((pub) => {
+
+    this.localParticipant.tracks.forEach((pub) => {
+      if (pub.track) {
+        this.localParticipant.unpublishTrack(pub.track);
```
lukasIO, line 400: explicitly unpublish tracks in order to remove remaining eventhandlers from the underlying LocalTrack of the LocalTrackPublication

src/room/participant/LocalParticipant.ts
```diff
@@ -448,8 +448,9 @@ export default class LocalParticipant extends Participant {
     }
 
     if (track instanceof LocalAudioTrack || track instanceof LocalVideoTrack) {
-      track.removeListener(TrackEvent.Muted, this.onTrackMuted);
-      track.removeListener(TrackEvent.Unmuted, this.onTrackUnmuted);
+      track.off(TrackEvent.Muted, this.onTrackMuted);
+      track.off(TrackEvent.Unmuted, this.onTrackUnmuted);
+      track.off(TrackEvent.Ended, this.onTrackUnpublish);
```
lukasIO, line 453: unsubscribing form the TrackEvent.Ended event, which was missing before

src/room/participant/LocalParticipant.ts
```diff
@@ -493,6 +490,8 @@ export default class LocalParticipant extends Participant {
         break;
     }
 
+    publication.dispose();
```
lukasIO, line 493: explicitly calling dispose on the publication, because we have no other means to determine that the publication is no longer needed. 
an alternative might be to call `publication.setTrack(null)` here in order to remove the eventhandlers that the publication attached to the LocalTrack.

src/room/track/TrackPublication.ts
```diff
@@ -36,17 +36,17 @@ export default class TrackPublication extends EventEmitter {
 
   /** @internal */
   setTrack(track?: Track) {
+    if (this.track) {
```
lukasIO, line 39: make sure that we remove event listeners from previous tracks before adding new ones

src/room/participant/LocalParticipant.ts
davidzhao: Unrelated to your PR, but we are missing a case if a user calls unpublishTrack(mediaStreamTrack).

A better check should probably be to set `track = publication.track` above this.

src/room/track/TrackPublication.ts
davidzhao: we should also deregister/register `TrackEvent.Ended` in subclass `LocalTrackPublication`

src/room/participant/LocalParticipant.ts
davidzhao: I like `setTrack(undefined)` better, since the concept of dispose doesn't exist anywhere within the SDK.

It also helps to catch the cases where we accidentally change the underlying LocalTrack without updating the handlers.

it seems that if we did it here, we would not have to worry doing so [above](https://github.com/livekit/client-sdk-js/pull/107/files#diff-cf84874acd5677364a55762127f033eb5a9e7ca344a666b6305c3d76452bff02R450)

davidzhao (commented):
solid change, just a couple of Qs and comments

lukasIO:
thanks for the review, I updated the code to implement your suggestions, but haven't had a chance yet to properly test these changes. I'll confirm once I've done that.

davidzhao (approved):
this is looking great! thanks for cleaning it up. I wish we had better tests for these. Dealing with MediaStreamTrack in jest is quite annoying

davidzhao:
Merging this in as I'm about to do some work that will touch these files.

PR #108, opened 2022-01-03
Support participant name and track mime type

davidzhao:


hiroshihorie (approved):
✅

src/room/track/TrackPublication.ts
```diff
@@ -18,6 +18,9 @@ export default class TrackPublication extends EventEmitter {
 
   source: Track.Source;
 
+  /** MimeType of the published track */
+  mimeType?: string;
```
davidliu, line 22: Are these mimeTypes auto detected by the server? Doesn't look like there's anything for the publisher to send up these mime types.

davidzhao: yup, we detect them when the media tracks are received. thanks for checking in on it!

PR #110, opened 2022-01-10
Fix handling of TrackEvent.Ended

davidzhao:
Track wasn't unpublished because it was not sent in the event.

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -119,6 +119,6 @@ export default class LocalTrack extends Track {
   }
 
   private handleEnded = () => {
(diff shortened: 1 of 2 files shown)


dsa (approved):
LG!

lukasIO (commented):
ah, sorry I missed this! 
LG!

ismaserrano:
This is occurss again! TrackEvent.Ended not fires.
In this commit https://github.com/livekit/client-sdk-js/commit/8a6fbba98c072836d54f0aee6b002f25f4f38b7a the track is not passed anymore. Some reason for that?

PR #113, opened 2022-01-11
Sample app's ability to simulate server-side conditions

davidzhao:


src/room/Room.ts
```diff
@@ -294,6 +296,34 @@ class Room extends EventEmitter {
...
+        });
+        break;
+      case 'node-failure':
+        req = SimulateScenario.fromPartial({
+          nodeFailure: true,
+        });
+        break;
+      case 'migration':
+        req = SimulateScenario.fromPartial({
+          migration: true,
```
cnderrauber, line 317: Are there something need server to do for simulate migration? Client can reconnect to trigger migration

PR #115, opened 2022-01-12
added option to update token

jibon57:
Added option to update token. In long run meeting sometime token got expire & if client need to run reconnect for any reason that time it failed. Now it will be possible to update token periodically.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -120,6 +120,10 @@ export default class RTCEngine extends EventEmitter {
     this.client.sendMuteTrack(trackSid, muted);
   }
 
+  updateToken(token: string) {
+    this.token = token;
+  }
+
   get dataSubscriberReadyState(): string | undefined {
     return this.reliableDCSub?.readyState;
   }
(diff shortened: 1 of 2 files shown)


davidzhao:
hey @jibon57 this is an excellent fix/idea.

I think we can handle this on the server side automatically without you having to regenerate/set token. What do you think about the following policy? If a participant is currently/recently connected, then we shall allow the expired token to be used.

This way all clients could continue to work seamlessly.

jibon57:
@davidzhao thank you. How you are planning to differentiate if that user joined recently? If user is smart enough then can reuse same token again & again. In our new system (which will be open source :-D) we have set validity of token for 10 minutes & renew using backend server periodically every 8 minutes which will require to send valid token before expire to renew. So, if meeting ended there have very less chances to reuse same token.

davidzhao:
Since it's a real time system, we know exactly when users are connected and have left. The idea is to have only a <1m window where users could re-connect after leaving.

jibon57:
This will be perfect!

davidzhao:
Should be handled by https://github.com/livekit/client-sdk-js/pull/131. The server will issue new, valid tokens to the client

PR #118, opened 2022-01-13
add getConnectedAddress

cnderrauber:
test on chrome and ff, can't get host addr in safari(can get port), MDN doc said safari has 'address' field in icecandidatestats, but log show it not. need to verify whether it not shows. Or we can get it from serverside's icecandidate msg, that is tricky, we can keep all candidates server sent and compair port to verify which one is used.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -126,6 +126,10 @@ export default class RTCEngine extends EventEmitter {
     return this.reliableDCSub?.readyState;
   }
 
+  get connectedServerAddress(): string | undefined {
+    return this.connectedServerAddr;
+  }
+
   private configure(joinResponse: JoinResponse) {
     // already configured
     if (this.publisher || this.subscriber) {
@@ -178,9 +182,8 @@ export default class RTCEngine extends EventEmitter {
           this.emit(EngineEvent.Connected);
         }
         getConnectedAddress(primaryPC).then((v) => {
-          this.connectedServerAddr = v
-          console.log("conncted to address", this.connectedServerAddr)
-        })
+          this.connectedServerAddr = v;
+        });
       } else if (primaryPC.iceConnectionState === 'disconnected' || primaryPC.iceConnectionState === 'failed') {
         log.trace('ICE disconnected');
         if (this.iceConnected) {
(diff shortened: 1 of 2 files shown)


src/room/RTCEngine.ts
```diff
@@ -437,3 +443,39 @@ export default class RTCEngine extends EventEmitter {
...
+  let candidates = new Map<string, string>();
+  const stats: RTCStatsReport = await pc.getStats()
+  stats.forEach((v) => {
+    switch (v.type) {
+      case 'transport':
+        selectedCandidatePairId = v.selectedCandidatePairId;
+        console.log(v.type, v)
+        break;
+      case 'candidate-pair':
+        if (selectedCandidatePairId === '' && v.selected){
```
cnderrauber, line 460: ff don't have transport stats, so use selected to determine selectedCandidatePairID

src/room/participant/LocalParticipant.ts
```diff
@@ -77,6 +77,10 @@ export default class LocalParticipant extends Participant {
     return this.microphoneError;
   }
 
+  get currentServerAddress(): String | undefined {
```
davidzhao, line 80: nit: since we are exposing engine, it's probably better to define an additional method here.

src/room/RTCEngine.ts
```diff
@@ -437,3 +446,37 @@ export default class RTCEngine extends EventEmitter {
     }
   }
 }
+
+async function getConnectedAddress(pc: RTCPeerConnection): Promise<string | undefined> {
```
davidzhao, line 450: nit: let's make this private. since we are setting it to a variable each time, there is no need for other classes to call it directly.

cnderrauber: not familiar with typescript, it is a helper function not a member of RTCEngine, if I add private on it , ide shows error (ts 1044)

src/room/participant/LocalParticipant.ts
cnderrauber: does additional method means pick another name?

davidzhao: sorry I meant to type: probably don't need this wrapper. we could remove this definition in local participant.

src/room/RTCEngine.ts
davidzhao: ah yes.. you are right.

cnderrauber: or should add other modifier on that?

davidzhao (approved):
lg! let's just remove the LocalParticipant method definition and leave it all in `engine`

PR #120, opened 2022-01-13
Default case for createLocalAudio/VideoTrack

jmswaney:
Hey LiveKit team! 

First of all, upgrading from v0.14 to v0.15 was super easy, so nice work on that release.

I noticed the previous `createLocalVideoTrack` behavior didn't require any arguments to actually get a video track with the default camera. It seemed redundant to have to add `createLocalVideoTrack(true)`, and Chrome threw a getUserMedia error for me if I don't include it. I just made this PR to automatically fallback to that for these helper functions.

### DIFF
--- a/src/room/track/create.ts
+++ b/src/room/track/create.ts
@@ -55,7 +55,7 @@ export async function createLocalVideoTrack(
 ): Promise<LocalVideoTrack> {
   const tracks = await createLocalTracks({
     audio: false,
-    video: options,
+    video: options ?? true,
   });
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
Thanks for the fix! We'll cut a new release shortly.

davidzhao:
actually, I think the default behavior for `createLocalTracks` is also wrong. I have another PR that will address both issues: https://github.com/livekit/client-sdk-js/pull/121

PR #122, opened 2022-01-14
Fix scaleResolutionDownBy going below 1 for tiny portrait source

davidzhao:


davidzhao:
@hiroshihorie @davidliu this is reported by a user seen on a mobile device. With portrait inputs this becomes tricky to compute the right scale var.

hiroshihorie (approved):
✅

PR #126, opened 2022-01-16
Fix reconnect loop on Safari

davidzhao:
Safari flips ICE connection to disconnected temporarily during re-negotiations. We cannot use it as a trigger to reconnect to the server


dsa (approved):
lg

PR #127, opened 2022-01-17
use ConnectionState insteandof ICEConnectionState

cnderrauber:
use ConnectionState for consistent with server side;
refine RTCEngine's Reconnected Event

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -39,7 +39,9 @@ export default class RTCEngine extends EventEmitter {
 
   private subscriberPrimary: boolean = false;
 
-  private iceConnected: boolean = false;
+  private pcConnected: boolean = false;
+
+  private reconnected: boolean = false;
 
   private isClosed: boolean = true;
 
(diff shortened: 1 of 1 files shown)


src/room/RTCEngine.ts
```diff
@@ -371,8 +375,9 @@ export default class RTCEngine extends EventEmitter {
 
     const startTime = (new Date()).getTime();
 
+    this.reconnected = false;
     while ((new Date()).getTime() - startTime < maxICEConnectTimeout * 2) {
-      if (this.iceConnected) {
+      if (this.reconnected) {
```
cnderrauber, line 380: if we check connected here, will return immediately. the peer connection state transition during reconnect:
1. connected
2. start reconnect
3. swap sdp & ice candidate, connected->connecting
4. ice restart complete, connecting->connected

davidzhao: I believe it stays connected only if signal connection is the one that becomes severed? this makes sense.. but I wonder if we can accomplish the same by setting `this.pcConnected = false` prior to entering the while check loop

davidzhao (approved):

src/room/RTCEngine.ts
cnderrauber: I think it can be done. Let me check is there other place use pcConnected

PR #130, opened 2022-01-20
Process reconnect when connectionstatechange callback not called

cnderrauber:
in case of ice restart, there is a chance that neither iceconnectionstatechange or connectionstatechange will be called, there is no ice conn or pc conn state transit (sticky on connected). in this case, detect peerconnection state is still connected after 2 second.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -39,7 +39,9 @@ export default class RTCEngine extends EventEmitter {
 
   private subscriberPrimary: boolean = false;
 
-  private iceConnected: boolean = false;
+  private pcConnected: boolean = false;
+
+  private reconnected: boolean = false;
 
   private isClosed: boolean = true;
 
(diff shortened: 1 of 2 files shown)


src/room/RTCEngine.ts
```diff
@@ -15,6 +15,7 @@ import { sleep } from './utils';
 const lossyDataChannel = '_lossy';
 const reliableDataChannel = '_reliable';
 const maxReconnectRetries = 5;
+const maxReconnectingTimeout = 2 * 1000;
```
davidzhao, line 18: do you think a minimum wait for 1s is sufficient to allow connected to flip to connecting?

```suggestion
const minReconnectWait = 1 * 1000;
```

src/room/RTCEngine.ts
```diff
@@ -39,6 +40,8 @@ export default class RTCEngine extends EventEmitter {
 
   private subscriberPrimary: boolean = false;
 
+  private pcConnectionState: RTCPeerConnectionState = 'new';
```
davidzhao, line 43: can we not store this var, and check primaryPC.connectionState?

src/room/RTCEngine.ts
cnderrauber: I'm not sure, 2s just a initial value. need more observe

src/room/RTCEngine.ts
```diff
@@ -372,13 +376,20 @@ export default class RTCEngine extends EventEmitter {
 
     const startTime = (new Date()).getTime();
 
-    while ((new Date()).getTime() - startTime < maxICEConnectTimeout) {
+    let now = startTime;
+    while (now - startTime < maxICEConnectTimeout) {
+      // if there is no connectionstatechange callback fired
+      // check connectionstate after maxReconnectingTimeout
```
davidzhao, line 382: nit: update comment too

```suggestion
      // check connectionstate after minReconnectWait
```

davidzhao (approved):

PR #131, opened 2022-01-24
Handle server refreshed tokens

davidzhao:


jibon57:
@davidzhao thank you! Won't it expose with any event back to App? For my case I'm using this token to verify other part of the application.

hiroshihorie (approved):

davidzhao:
> @davidzhao thank you! Won't it expose with any event back to App? For my case I'm using this token to verify other part of the application.

This is not supposed to be exposed to the app itself. The idea is the ability for LiveKit clients to remain connected regardless of when the original token expires.

davidzhao:
@jibon57 I'm curious about your use case that would require getting the new token.

jibon57:
@davidzhao in my application I'm using token to verify other request from frontend to backend. In this case I ain't using any other authentication system. If user's token is valid then other operation like update metadata (my case different permission), upload files, different moderator task etc will allow by backend server. I'm developing a complete web conference system based on`livekit` which will be published as open source very soon. Hopefully within next month.

davidliu (approved):

lukasIO:
@davidzhao  Just to understand the main idea: 
The updated token would only (?) be used in a reconnection attempt, right?

davidzhao:
> @davidzhao Just to understand the main idea: The updated token would only (?) be used in a reconnection attempt, right?

yes! that's the main goal. we do not use the token for anything else. However, the refreshed token will retain all of the permissions/attributes as the original token. It simply updates the expiration field.

lukasIO (commented):
then everything is clear to me, LG!

Maybe we should consider firing an event with the updated token?
It would allow apps to update the token within their own state handling. And next time they connect manually they already have the updated token.

PR #137, opened 2022-02-02
Workaround Safari 15 video rendering bug

davidzhao:


dsa (approved):
i wish this wasn't necessary, but LG and thanks for digging into it!

lukasIO (approved):
I don't have safari to test, but looks good and still works in firefox and chrome.

As a general side note: one thing I had noticed previously, that some browsers (iirc also safari) were a lot more likely to autoplay videos if the HTMLVideoElement was explicitly set to `muted = true`.

PR #139, opened 2022-02-07
typesafe roomevents

lukasIO:
this would be a way to type the events. Works for both emitter and listener.
It's not finished and some cyclic dependencies would need to be fixed.

I also tried to get rid of the need to declare the events twice (once with the callback types and once as the enum) by generating the enum automatically from the callback types, but it seems like we would lose the JSdocs description of the enum, which I guess is a rather big loss.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -1,4 +1,5 @@
 import { EventEmitter } from 'events';
+import type TypedEmitter from 'typed-emitter';
 import { toProtoSessionDescription } from '../api/SignalClient';
 import log from '../logger';
 import { RoomConnectOptions, RoomOptions } from '../options';
@@ -16,8 +17,9 @@ import {
 import DeviceManager from './DeviceManager';
 import { ConnectionError, UnsupportedServer } from './errors';
 import {
-  EngineEvent, ParticipantEvent, RoomEvent, TrackEvent,
+  EngineEvent, ParticipantEvent, RoomEvent, RoomEventCallbacks, TrackEvent,
 } from './events';
+import type { RoomEventType } from './events';
 import LocalParticipant from './participant/LocalParticipant';
 import Participant, { ConnectionQuality } from './participant/Participant';
 import RemoteParticipant from './participant/RemoteParticipant';
@@ -44,7 +46,7 @@ export enum RoomState {
  *
  * @noInheritDoc
  */
-class Room extends EventEmitter {
+class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) {
   state: RoomState = RoomState.Disconnected;
 
   /** map of sid: [[RemoteParticipant]] */
@@ -144,7 +146,7 @@ class Room extends EventEmitter {
       .on(EngineEvent.DataPacketReceived, this.handleDataPacket)
       .on(EngineEvent.Resuming, () => {
         this.state = RoomState.Reconnecting;
-        this.emit(RoomEvent.Reconnecting);
+        this.emit(RoomEvent.Reconnected);
       })
       .on(EngineEvent.Resumed, () => {
         this.state = RoomState.Connected;
@@ -830,8 +832,12 @@ class Room extends EventEmitter {
     }
   }
 
-  /** @internal */
-  emit(event: string | symbol, ...args: any[]): boolean {
+  // /** @internal */
+  // emit(event: string | symbol, ...args: any[]): boolean {
+  //   log.debug('room event', event, ...args);
+  //   return super.emit(event, ...args);
+  // }
+  emit<E extends keyof RoomEventCallbacks>(event: E, ...args: Parameters<RoomEventCallbacks[E]>): boolean {
     log.debug('room event', event, ...args);
     return super.emit(event, ...args);
   }
--- a/src/room/events.ts
+++ b/src/room/events.ts
@@ -1,3 +1,6 @@
+import { DataPacket_Kind, LocalTrackPublication, Participant, RemoteTrack, RemoteTrackPublication } from '..';
+import RemoteParticipant from './participant/RemoteParticipant';
+
 /**
  * Events are the primary way LiveKit notifies your application of changes.
  *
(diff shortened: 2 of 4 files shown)


davidzhao (commented):
this is awesome! such a good way to introduce type safety w/o breaking compatibility!

the only tricky thing here is to avoid import cycles. maybe moving `RoomEventCallbacks` to another file?

lukasIO:
@davidzhao Yeah, another file will be needed for those events, do you think a `types.ts`  next to `Room.ts` one next to `Participant.ts` would be the way to go? (Looks like we have to split up RoomEventCallbacks and ParticipantEventCallbacks). 
Also `ConnectionQuality` cannot live in `Participant.ts` anymore.

lukasIO:
To fix another dependency loop around the RTCEngine I had to include the EventCallbacks for the engine within the file (or create an extra types file just for the engine), so I now went with the consistent approach to put them into the equivalent class files.

If you think it's better to separate them, I'll gladly do, I just didn't have a nice idea of how to keep them all seperate without cluttering the file tree. Maybe it could work if we add another folder for RTCEngine and also give it a `types.ts` file.

src/room/Room.ts
```diff
@@ -830,11 +831,88 @@ class Room extends EventEmitter {
     }
   }
 
-  /** @internal */
-  emit(event: string | symbol, ...args: any[]): boolean {
+  // /** @internal */
+  // emit(event: string | symbol, ...args: any[]): boolean {
```
davidzhao, line 835: nice! it's great you are able to type this too.

src/room/Room.ts
```diff
@@ -830,11 +831,88 @@ class Room extends EventEmitter {
...
+  ) => void,
+  trackSubscriptionPermissionChanged: (
+    publication: RemoteTrackPublication,
+    status: TrackPublication.SubscriptionStatus,
+    participant: RemoteParticipant,
+  ) => void,
+  audioPlaybackChanged: (playing: boolean) => void,
+};
+
+// type KeysMissingFromRoomEventCallbacks = Exclude<RoomEvent, keyof RoomEventCallbacks>;
```
davidzhao, line 909: are these test code?

src/room/events.ts
```diff
@@ -146,7 +147,7 @@ export enum RoomEvent {
    * args: (prevMetadata: string, [[Participant]])
    *
    */
-  ParticipantMetadataChanged = 'participantMetaDataChanged',
```
davidzhao, line 149: thank you for catching this 🙏

davidzhao (commented):
this looks good! I'm excited to have type safety in our callbacks!

There appears to be a few other compilation errors that remain (when I run this branch with `yarn sample`). They don't appear in GH CI though. Do you know why?

```
ERROR in /Users/davidzhao/dev/livekit/client-sdk-js/src/room/participant/RemoteParticipant.ts
../src/room/participant/RemoteParticipant.ts 239:22-27
[tsl] ERROR in /Users/davidzhao/dev/livekit/client-sdk-js/src/room/participant/RemoteParticipant.ts(239,23)
      TS2345: Argument of type 'string | symbol' is not assignable to parameter of type 'keyof ParticipantEventCallbacks'.
  Type 'string' is not assignable to type 'keyof ParticipantEventCallbacks'.
 @ ../src/index.ts 39:44-91
 @ ./sample.ts 12:16-39
```

src/room/Room.ts
lukasIO: It would be a way to raise a TS Error for the case that we haven't specified all RoomEvents within the RoomEventsCallback. 
I'm not sure if we'd want that. Will remove for now, as a comment it's useless anyways

src/room/participant/Participant.ts
```diff
@@ -158,8 +157,8 @@ export default class Participant extends EventEmitter {
     this.metadata = md;
 
     if (changed) {
-      this.emit(ParticipantEvent.MetadataChanged, prevMetadata, this);
-      this.emit(ParticipantEvent.ParticipantMetadataChanged, prevMetadata, this);
```
lukasIO, line 162: @davidzhao do we need to pass the participant instance here? 
there are a couple of occasions within this PR where I removed the `this` from event emitters.

The roomevents, which actually have the participant as one of the arguments, do not (afaict) use the reference from the participantevent emitter, but rather the participant object on which they listen to events, as in 
```ts
participant.on(ParticipantEvent.ParticipantMetadataChanged, (metadata: string | undefined) => {
          this.emit(RoomEvent.ParticipantMetadataChanged, metadata, participant);
        })
```

lukasIO:
Sorry about that, it got late and I was a bit too excited, so for some late changes I naively trusted the fact that the type checker was not complaining.

src/room/participant/Participant.ts
davidzhao: I think we would want to pass in *this* since it was doing it before. agreed the params aren't always consistent. perhaps we could update that in the next major version.

example/sample.ts
```diff
@@ -101,8 +101,8 @@ const appActions = {
         appendLog('media device failure', failure);
       })
       .on(RoomEvent.ConnectionQualityChanged,
-        (quality: ConnectionQuality, participant: Participant) => {
-          appendLog('connection quality changed', participant.identity, quality);
+        (quality: ConnectionQuality, participant?: Participant) => {
```
davidzhao, line 104: seems that `participant` should always be set here?

src/room/Room.ts
```diff
@@ -830,11 +831,73 @@ class Room extends EventEmitter {
...
 export default Room;
+
+export type RoomEventCallbacks = {
+  reconnecting: () => void,
+  reconnected: () => void,
+  disconnected: () => void,
+  mediaDevicesChanged: () => void,
+  participantConnected: (participant: RemoteParticipant) => void,
+  participantDisconnected: (participant: RemoteParticipant) => void,
+  trackPublished: (publication: RemoteTrackPublication, participant?: RemoteParticipant) => void,
```
davidzhao, line 852: for the room events, I believe participants should always be present. in `Room.getOrCreateParticipant`, the listeners are only attached if participant is defined. There's probably some typescript coercion that needs to happen.

davidzhao (approved):
looks good, only change is re: `participant?` in Room callbacks

hiroshihorie:
This is great 😁

PR #140, opened 2022-02-07
queue up signal events while reconnecting

lukasIO:


src/api/SignalClient.ts
```diff
@@ -287,6 +297,13 @@ export class SignalClient {
   }
 
   async sendRequest(req: SignalRequest) {
+    // capture all requests while reconnecting and put them in a queue.
+    // TODO keep the order of requests by enqueuing new events until all
+    // enqueued events have been fired
+    if (this.isReconnecting && !req.simulate) {
```
boks1971, line 303: Is this covering the migration case? I guess `isReconnecting` is not set for migration. Is that correct? We need to be selective for migration. We need to send `SyncState` during migration, but hold other messages I believe. @davidzhao @cnderrauber would know for sure.

src/api/RequestQueue.ts
```diff
@@ -0,0 +1,56 @@
...
+    logger.debug('firing request from queue');
+  }
+
+  async run() {
+    if (this.running) return;
+    logger.debug('start queue');
+    this.running = true;
+    while (this.running && this.queue.length > 0) {
+      this.dequeue();
+      await new Promise((resolve) => setTimeout(() => resolve(true), this.msInbetweenEvents));
```
boks1971, line 32: There is a util called `sleep` in utils.ts. Would be good to use that for consistency.

boks1971 (approved):

src/api/RequestQueue.ts
```diff
@@ -0,0 +1,57 @@
...
+    logger.debug('firing request from queue');
+  }
+
+  async run() {
+    if (this.running) return;
+    logger.debug('start queue');
+    this.running = true;
+    while (this.running && this.queue.length > 0) {
+      this.dequeue();
+      await sleep(this.msInbetweenEvents);
```
davidzhao, line 33: Is there a reason why we wanted to delay between signal messages?

src/api/SignalClient.ts
davidzhao: this is only needed if we had resumed a session (either on the same server, or migrated to a different server). `SignalClient.reconnect` is only used in the case we are resuming a session. so this is right.

davidzhao: I agree we'd need to do the note in TODO before merging. the order of signal messages should be kept. Perhaps as long as the queue isn't empty, it should use the queue prior to sending directly.

davidzhao (commented):
looks good! simple way to accomplish this.

src/api/RequestQueue.ts
lukasIO: the idea behind it was to ensure the order even in tougher network conditions. But I guess this is already taken care of by the websocket connection anyways... Will remove!

src/api/SignalClient.ts
```diff
@@ -286,7 +296,15 @@ export class SignalClient {
     this.sendRequest(SignalRequest.fromPartial({ leave: {} }));
   }
 
-  async sendRequest(req: SignalRequest) {
+  async sendRequest(req: SignalRequest, fromQueue: boolean = false) {
```
lukasIO, line 299: I added the optional `fromQueue` param in order to differentiate between requests that are originating from the queue. Without that we'd need refactoring of the sendRequest function to ensure that the queue is emptied first.

davidzhao (approved):
looks good!

PR #141, opened 2022-02-07
account for devicepixelratio in adaptive streaming

lukasIO:


boks1971 (approved):

hiroshihorie (approved):

davidzhao:
@lukasIO I tested this a bit and I think we should revert this and send pixelDensity to the server instead. Let's discuss.. but I think
* this change alters the default behavior and could be a surprise to users. On modern displays (typically with pixelDensity of 2), it'll cause it to double the resolution of layers sent.
* server could have more flexibility of handling resolution matching differently for pixelDensity. i.e. should we really send 3x the size for pixelDensity = 3? or is 2x sufficient.
* devs may be able to customize pixelDensity multiplier from config/dashboard.

lukasIO:
I agree that it might be unexpected. 
I would also vote for it to be a config setting that defaults to false, only difference being that my initial incentive was to have it as a client side config option. I thought that might be easier to handle than updating the serverconfig, but definitely not a hill to die on.

PR #146, opened 2022-02-14
Add new publishing presets and options to configure custom simulcast layers

lukasIO:
I went through a couple of iterations on this and ended up putting custom simulcast layers into the publishing config. 

In terms of flexibility, I think, this works great and it turned out to be the most minimal change of the ones I had played with.

One drawback: Because the simulcast layers are not set via a function, we cannot provide log warnings when the simulcast layers are, for example, in the wrong order.

### DIFF
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -4,7 +4,7 @@ import {
   Participant, ParticipantEvent, RemoteParticipant, Room,
   RoomConnectOptions, RoomEvent,
   RoomOptions, RoomState, setLogLevel, Track, TrackPublication,
-  VideoCaptureOptions, VideoPresets,
+  VideoCaptureOptions, VideoPreset, VideoPresets,
 } from '../src/index';
 
 const $ = (id: string) => document.getElementById(id);
@@ -38,7 +38,7 @@ const appActions = {
         simulcast,
         videoSimulcastLayers: [
           VideoPresets.h90,
-          VideoPresets.h180,
+          VideoPresets.h216,
         ],
       },
       videoCaptureDefaults: {
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -201,8 +201,8 @@ const appActions = {
       const msg = state.encoder.encode(textField.value);
       currentRoom.localParticipant.publishData(msg, DataPacket_Kind.RELIABLE);
       (<HTMLTextAreaElement>(
-        $('chat')
-      )).value += `${currentRoom.localParticipant.identity} (me): ${textField.value}\n`;
+      $('chat')
+    )).value += `${currentRoom.localParticipant.identity} (me): ${textField.value}\n`;
       textField.value = '';
     }
   },
@@ -465,7 +465,7 @@ function renderParticipant(participant: Participant, remove: boolean = false) {
       break;
     default:
       signalElm.innerHTML = '';
-    // do nothing
+      // do nothing
   }
 }
 
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -36,9 +36,13 @@ const appActions = {
       dynacast,
       publishDefaults: {
         simulcast,
+        videoSimulcastLayers: [
+          VideoPresets.h90,
+          VideoPresets.h180,
+        ],
       },
       videoCaptureDefaults: {
-        resolution: VideoPresets.p720.resolution,
+        resolution: VideoPresets.h720.resolution,
       },
     };
 
@@ -197,8 +201,8 @@ const appActions = {
       const msg = state.encoder.encode(textField.value);
       currentRoom.localParticipant.publishData(msg, DataPacket_Kind.RELIABLE);
       (<HTMLTextAreaElement>(
-      $('chat')
-    )).value += `${currentRoom.localParticipant.identity} (me): ${textField.value}\n`;
+        $('chat')
+      )).value += `${currentRoom.localParticipant.identity} (me): ${textField.value}\n`;
       textField.value = '';
     }
   },
@@ -461,7 +465,7 @@ function renderParticipant(participant: Participant, remove: boolean = false) {
       break;
     default:
       signalElm.innerHTML = '';
-      // do nothing
+    // do nothing
   }
 }
 
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -38,7 +38,7 @@ const appActions = {
         simulcast,
       },
       videoCaptureDefaults: {
-        resolution: VideoPresets.hd.resolution,
+        resolution: VideoPresets.p720.resolution,
       },
     };
 
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -4,7 +4,7 @@ import {
   Participant, ParticipantEvent, RemoteParticipant, Room,
   RoomConnectOptions, RoomEvent,
   RoomOptions, RoomState, setLogLevel, Track, TrackPublication,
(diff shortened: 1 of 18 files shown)


example/sample.ts
```diff
@@ -36,9 +36,13 @@ const appActions = {
       dynacast,
       publishDefaults: {
         simulcast,
+        videoSimulcastLayers: [
```
lukasIO, line 39: not sure if we should have a "power feature" in the sample app?

src/room/participant/publishUtils.ts
```diff
@@ -65,6 +55,8 @@ export function computeVideoEncodings(
   if (isScreenShare) {
     videoEncoding = options?.screenShareEncoding;
   }
+  // TODO this currently prevents any screenshare tracks from using simulcast
+  // does it even make sense to have simulcast layers for screenshare then?
   const useSimulcast = !isScreenShare && options?.simulcast;
```
lukasIO, line 60: Should we make a change here to allow screensharing to use simulcast?

lukasIO:
should https://github.com/livekit/client-sdk-js/blob/main/src/room/participant/publishUtils.ts#L111
be `480` instead of `500` to match https://chromium.googlesource.com/external/webrtc/+/master/media/engine/simulcast.cc#103 ?

example/sample.ts
davidzhao: I think this is great! helps to show users how this could be used.

slight request, could we make a middle layer more like ~h200 ? since the sample app renders video at roughly 200-250 pixels high

src/room/participant/publishUtils.ts
davidzhao: I think we shouldn't change it yet.. when screenshare are simulcasted, lower layers add very little value due to how blurry they tend to get.. let's talk it through and experiment with it later.

src/room/participant/publishUtils.ts
```diff
@@ -27,30 +27,20 @@ export function mediaTrackToLocalTrack(
...
 
 /* @internal */
 export const presets43 = [
-  VideoPresets43.qvga,
-  VideoPresets43.vga,
-  VideoPresets43.qhd,
-  VideoPresets43.hd,
-  VideoPresets43.fhd,
+  VideoPresets43.h120,
+  VideoPresets.h360,
```
davidzhao, line 38:
```suggestion
  VideoPresets43.h360,
```

src/room/participant/publishUtils.ts
```diff
@@ -82,17 +74,27 @@ export function computeVideoEncodings(
...
-
-  const presets = presetsForResolution(isScreenShare, width, height);
+  let presets: Array<VideoPreset> = [];
+  if (isScreenShare) {
+    // TODO should we pre-sort the incoming presets array by dimensions and/or bitrate?
+    presets = options?.screenShareSimulcastLayers
+      ?? presetsForResolution(isScreenShare, width, height);
+  } else {
+    // TODO should we pre-sort the incoming presets array by dimensions and/or bitrate?
+    presets = options?.videoSimulcastLayers
```
davidzhao, line 84: if user specified simulcast layers are passed in, we should always sort the layers so that it's low -> high. It's probably simpler for users to pass in an array of layers instead of low/mid/high.

there's another reason.. is that we'll always need a low layer when simulcast is used.. and the layers have to be continuous (i.e. low, _, high would break our adaptive logic on the server side).

src/room/participant/publishUtils.ts
```diff
@@ -82,17 +74,27 @@ export function computeVideoEncodings(
...
   }
   const original = new VideoPreset(
     width, height, videoEncoding.maxBitrate, videoEncoding.maxFramerate,
   );
 
+  log.debug('simulcast layers for publishing: ', [lowPreset, midPreset, original]);
+
   // NOTE:
   //   1. Ordering of these encodings is important. Chrome seems
   //      to use the index into encodings to decide which layer
```
davidzhao, line 100: I wonder if this logic should change. since now a user could specify presets. i.e. should we still enforce there can only be one under 500w?

```
  if (size >= 500) {
    return encodingsFromPresets(width, height, [
      lowPreset, original,
    ]);
  }
```

src/room/track/options.ts
```diff
@@ -202,28 +216,65 @@ export namespace AudioPresets {
...
   qhd: new VideoPreset(720, 540, 450_000, 25),
+  /** @deprecated */
   hd: new VideoPreset(960, 720, 1_500_000, 30),
+  /** @deprecated */
   fhd: new VideoPreset(1440, 1080, 2_800_000, 30),
-};
+} as const;
 
 export const ScreenSharePresets = {
+  h360fps3: new VideoPreset(640, 360, 200_000, 3),
```
davidzhao, line 265: nice!

src/room/track/options.ts
```diff
@@ -202,28 +216,65 @@ export namespace AudioPresets {
  * Sane presets for video resolution/encoding
  */
 export const VideoPresets = {
+  h90: new VideoPreset(160, 90, 60_000, 10),
+  h180: new VideoPreset(320, 180, 120_000, 10),
```
davidzhao, line 220: was wondering if we can increase the framerate on h180 to ~15, Zoom seems to be doing 15 and feels smoother than ours currently. I think we can keep bitrate roughly the same.

davidzhao (commented):
looks great!

src/room/participant/publishUtils.ts
lukasIO: Good question, but generally this logic pretty much matches what chrome sees as layer restrictions anyways, right?
https://chromium.googlesource.com/external/webrtc/+/master/media/engine/simulcast.cc#103

we could change 500 to 480 in order to align exactly with what they are doing

src/room/participant/publishUtils.ts
lukasIO: if we leave it as is, it might be confusing to expose `screenShareSimulcast` as a publishing option.
I'll mark that option as exerimental then.

src/room/participant/publishUtils.ts
```diff
@@ -27,30 +27,20 @@ export function mediaTrackToLocalTrack(
...
 /* @internal */
 export const presets43 = [
-  VideoPresets43.qvga,
-  VideoPresets43.vga,
-  VideoPresets43.qhd,
-  VideoPresets43.hd,
-  VideoPresets43.fhd,
+  VideoPresets43.h120,
+  VideoPresets43.h360,
+  VideoPresets.h540,
```
davidzhao, line 39:
```suggestion
  VideoPresets43.h540,
```

davidzhao (approved):
awesome work! just one nit :)

src/room/participant/publishUtils.ts
```diff
@@ -27,30 +27,20 @@ export function mediaTrackToLocalTrack(
...
 ];
 
 /* @internal */
 export const presets43 = [
-  VideoPresets43.qvga,
-  VideoPresets43.vga,
-  VideoPresets43.qhd,
-  VideoPresets43.hd,
-  VideoPresets43.fhd,
+  VideoPresets43.h120,
```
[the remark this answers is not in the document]
lukasIO: it was intentional, but arbitrary, if that makes any sense. happy to use 180 as well.

PR #147, opened 2022-02-14
device pixel ratio as setting

lukasIO:
With #141 reverted, this would be an approach of letting users define the `pixelDensity`.

In this PR this is achieved with a client side setting, which gives users the flexibility to decide whether or not to use it depending on custom constraints. For example *"use a higher pixelDensity on desktop devices, but default to 1 on mobile devices"*.

Biggest change here is that the `adaptiveStream` setting can now take the type `AdaptiveStreamSettings`. 
Conceptually `pixelDensity` belongs to the `adaptiveStream` setting and to foreshadow for when it might be possible to customize the adaptiveStream even further, I thought it would be worth it.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -522,7 +522,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     // handle changes to participant state, and send events
     participantInfos.forEach((info) => {
       if (info.sid === this.localParticipant.sid
-        || info.identity === this.localParticipant.identity) {
+          || info.identity === this.localParticipant.identity) {
         this.localParticipant.updateInfo(info);
         return;
       }
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -28,7 +28,7 @@ import LocalTrackPublication from './track/LocalTrackPublication';
 import RemoteTrackPublication from './track/RemoteTrackPublication';
 import { Track } from './track/Track';
 import { TrackPublication } from './track/TrackPublication';
-import { RemoteTrack } from './track/types';
+import { AdaptiveStreamSettings, RemoteTrack } from './track/types';
 import { unpackStreamId } from './utils';
 
 export enum RoomState {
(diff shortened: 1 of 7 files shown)


src/room/Room.ts
```diff
@@ -522,7 +522,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     // handle changes to participant state, and send events
     participantInfos.forEach((info) => {
       if (info.sid === this.localParticipant.sid
-          || info.identity === this.localParticipant.identity) {
+        || info.identity === this.localParticipant.identity) {
```
davidzhao, line 525: nit: is eslint making these formatting changes? it'd be good to align the config in .eslintrc.js so developers use the same rules.

src/room/track/types.ts
```diff
@@ -6,3 +6,19 @@ import type RemoteVideoTrack from './RemoteVideoTrack';
 export type RemoteTrack = RemoteAudioTrack | RemoteVideoTrack;
 export type AudioTrack = RemoteAudioTrack | LocalAudioTrack;
 export type VideoTrack = RemoteVideoTrack | LocalVideoTrack;
+
+export type AdaptiveStreamSettings = {
+  /**
+   * Enable adaptive streaming
+   */
+  enabled: boolean,
```
davidzhao, line 14: do you think this becomes redundant? i.e. it seems simpler to set `adaptiveStream: false` vs `adaptiveStream: { enabled: false }`

davidzhao (commented):
it makes sense to handle it on the client side here. nice work!

src/room/Room.ts
lukasIO: turns out eslint was only configured on my machine to check, but not to actually fix these kind of errors. I think it was using a default prettier config for that. sorry about that, it's fixed now and I will revert these changes!

src/room/track/types.ts
lukasIO: I considered leaving `enabled` out entirely. The reason why I didn't go down that way is, that it becomes easier to switch between enabled and disabled without touching the other option(s) if you set `enabled` explicitly. 

It was nothing more than a slight tendency towards being more explicit, but I share your sentiment that it feels a bit redundant.

hiroshihorie:
for Swift SDK I was thinking may be something like :
```swift
enum AdaptiveStreamMode {
  case off
  // up to sdk what to do
  case auto
  // depends on current screen's pixel density
  // In macOS window can be dragged in to another display with different pixel density, so better re-compute
  case screen
  case fixed(pixelDensity: Int)
}
```

lukasIO:
@hiroshihorie dragging to another screen with a different pixel density is a very good point. In that case 'screen' makes a lot of sense. 
Do you think that `pixelDensity` will stay the only factor potentially influencing `AdaptiveStreamMode` ? 
Your suggestion looks a lot nicer than mine, with the caveat that it is tightly bound to pixel density only.
But I also cannot think of another factor influencing the adaptive stream mode, you?

hiroshihorie:
> But I also cannot think of another factor influencing the adaptive stream mode, you?

You have a good point, if i think about it there could be a few more options.
for example, how often to compute / send to the server (debounce time)

lukasIO:
Nothing concrete, but looking at your enum maybe looking at `pixelDensity` with the options for `auto, screen, fixed` as one parameter of `AdaptiveStreamMode` ?

hiroshihorie (approved):
I think It makes sense the way we specify options for JS and Swift will be different since Swift can set associated values for the enum.

PR #149, opened 2022-02-17
fix simulcast preset computation

lukasIO:


src/room/participant/publishUtils.ts
```diff
@@ -26,22 +26,28 @@ export function mediaTrackToLocalTrack(
...
+export const presets43 = Object.values(VideoPresets43);
+
+/* @internal */
+export const presetsScreenShare = Object.values(ScreenSharePresets);
+
+/* @internal */
+export const defaultSimulcastPresets169 = [
   VideoPresets.h180,
   VideoPresets.h360,
   VideoPresets.h540,
```
davidzhao, line 41: I think the third entry won't be used since it'll be the original?

boks1971 (approved):
lgtm!

src/room/participant/publishUtils.ts
```diff
@@ -26,22 +26,28 @@ export function mediaTrackToLocalTrack(
 }
 
 /* @internal */
-export const presets169 = [
+export const presets169 = Object.values(VideoPresets);
```
davidzhao, line 29: we'll need a way to filter out the @deprecated values in there.

src/room/participant/publishUtils.ts
```diff
@@ -26,22 +26,28 @@ export function mediaTrackToLocalTrack(
...
+/* @internal */
+export const defaultSimulcastPresets169 = [
   VideoPresets.h180,
   VideoPresets.h360,
   VideoPresets.h540,
 ];
 
 /* @internal */
-export const presets43 = [
+export const defaultSimulcastPresets43 = [
```
davidzhao, line 45: thanks for cleaning up the var names! this is more clear

src/room/participant/publishUtils.ts
lukasIO: very good point. but the values of the presets themselves didn't change, right? 
The only thing happening now is that the array might contain 720p twice, which shouldn't be a problem when trying to find the closest match...? We are iterating over it in a loop and I put the deprecated presets to be the last ones in the object, so theoretically it should never actually reach those

src/room/participant/publishUtils.ts
lukasIO: yes, unintentionally preparing for the glorious future, where browsers support 4 layers... will fix

src/room/participant/publishUtils.ts
davidzhao: it would have only been a problem if they were publishing >h2160, so that should be ok for 99.99999% of the cases :)

davidzhao (approved):

PR #151, opened 2022-02-17
Queue signalling until ICE reconnected

lukasIO:
I'm not entirely sure if
1. just after `sendAnswer` is the right place to set `isReconnecting` to false
2. we'd need more (or rather less) events to pass in `passThroughQueueSignals`. 

tried to keep it flexible by just putting the requests that need to pass in an array.

### DIFF
--- a/src/api/RequestQueue.ts
+++ b/src/api/RequestQueue.ts
@@ -1,4 +1,4 @@
-import logger from '../logger';
+import log from '../logger';
 
 export default class Queue {
   private queue: Array<() => void>;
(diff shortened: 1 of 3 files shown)


davidzhao:
> 1. just after `sendAnswer` is the right place to set `isReconnecting` to false

we would not be quite ready to send the events then. after [this line](https://github.com/livekit/client-sdk-js/blob/d029f36a670384fadec5fb58ebbcf2336f238434/src/room/RTCEngine.ts#L480) is when we are certain the peer connection is re-established.

> 2. we'd need more (or rather less) events to pass in passThroughQueueSignals.

you are right.. thinking about this, we are not able to queue up the following
* ICE candidate exchange
* offer/answer exchange
* sync state
* leave

they are required in order to establish the peer connection, or to signal to the server that the client is exiting.

davidzhao (approved):

PR #153, opened 2022-02-18
initial draft for client configuration system

lukasIO:
current state only handles the `createTracks` function and an update if tracks had already been published. 
some things are still not entirely clear to me, will leave some comments in those places


src/room/participant/LocalParticipant.ts
```diff
@@ -212,8 +216,27 @@ export default class LocalParticipant extends Participant {
       this.roomOptions?.audioCaptureDefaults,
       this.roomOptions?.videoCaptureDefaults,
     );
-
     const constraints = constraintsForOptions(opts);
+    if (this.clientConfiguration) {
```
lukasIO, line 220: not happy with this part, would probably be a good place to extract into some sort of util function. Will do, once I know, that the general approach is correct

src/room/participant/LocalParticipant.ts
```diff
@@ -668,6 +698,42 @@ export default class LocalParticipant extends Participant {
     }
   }
 
+  /** @internal */
+  async updateConfiguration(config: RTCClientConfiguration | undefined) {
+    config = {
```
lukasIO, line 703: This config object is only filled with dummy data in order to test if things are working ok on the js side. need to remove it

src/room/track/LocalTrack.ts
```diff
@@ -61,6 +66,16 @@ export default class LocalTrack extends Track {
     return this;
   }
 
+  /** @internal */
+  async applyClientConfiguration(config: RTCClientConfiguration) {
+    const clientConstraints = configToConstraints(this.kind, config);
+    const mergedConstraints = {
```
lukasIO, line 72: should probably use the `mergeWithoutOverwriting` util here

src/room/participant/LocalParticipant.ts
```diff
@@ -415,14 +438,21 @@ export default class LocalParticipant extends Participant {
 
     // store RTPSender
     track.sender = transceiver.sender;
+    // store Transceiver
+    track.transceiver = transceiver;
```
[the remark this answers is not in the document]
boks1971: Would be good to dig into configurability of an already active transceiver on different browsers. There are two things to consider here I think
1. Pre-publish: we can change codec. Won't be able to change codec after publishing. Would have to be a new track if a codec change is needed after publish.
2. Being able to change `encodings` (`layers`) on the fly. Can imagine a case of a wall of postage stamp sized videos. In this case, the dynacast (maybe v2 of dynacast :-) ), change the encoding so that lowest layer is of postage stamp size so that a lot of video streams can fit in the subscriber peer connection.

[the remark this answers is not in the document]
cnderrauber: Agree with  test for browsers. Spec not indicate about timing of set preferred codec, so we should prepare for both two use cases.

cnderrauber (approved):

davidzhao (approved):
looks good!

PR #155, opened 2022-02-20
compare video presets by bitrate and fps

lukasIO:
as mentioned in https://github.com/livekit/client-sdk-swift/pull/53 this is to adapt the same preset comparing logic for the js sdk.


davidzhao (approved):
lg!

PR #156, opened 2022-02-21
enable screen share simulcast by default

lukasIO:
Considerations:
- hardcode the default simulcast layer to something that makes sense with the default `screenShareEncoding`
  * would be nice to have all defaults in one place
- dynamically compute the default simulcast layer from the supplied default `screenShareEncoding`
  * a lot more flexibel, not sure if a linear computation of the bitrate makes sense

First commit is using the first option only, second commit computes the layer from the provided original preset.
I'd favor the automatic computation

### DIFF
--- a/src/room/participant/publishUtils.ts
+++ b/src/room/participant/publishUtils.ts
@@ -52,9 +52,9 @@ export const computeDefaultScreenShareSimulcastPresets = (fromPreset: VideoPrese
   return layers.map((t) => new VideoPreset(
     Math.floor(fromPreset.width / t.scaleResolutionDownBy),
     Math.floor(fromPreset.height / t.scaleResolutionDownBy),
-    Math.floor(fromPreset.encoding.maxBitrate
-      / (t.scaleResolutionDownBy * ((fromPreset.encoding.maxFramerate ?? 30) / t.fps))),
-    t.fps,
+    Math.max(150_000, Math.floor(fromPreset.encoding.maxBitrate
+      / (t.scaleResolutionDownBy ** 2 * ((fromPreset.encoding.maxFramerate ?? 30) / t.fps))),
+    t.fps),
   ));
 };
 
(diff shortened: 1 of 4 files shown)


src/room/participant/publishUtils.ts
```diff
@@ -46,6 +46,18 @@ export const defaultSimulcastPresets43 = [
...
 ];
 
+/* @internal */
+export const computeDefaultScreenShareSimulcastPresets = (fromPreset: VideoPreset) => {
+  const layers = [{ scaleResolutionDownBy: 2, fps: 3 }];
+  return layers.map((t) => new VideoPreset(
+    Math.floor(fromPreset.width / t.scaleResolutionDownBy),
+    Math.floor(fromPreset.height / t.scaleResolutionDownBy),
+    Math.floor(fromPreset.encoding.maxBitrate
+      / (t.scaleResolutionDownBy * ((fromPreset.encoding.maxFramerate ?? 30) / t.fps))),
```
boks1971, line 56: For FPS, linear should be fine I think. For resolution it should be quadratic as it scales in both dimensions by that factor.

boks1971 (approved):

src/room/participant/publishUtils.ts
lukasIO: good point!

lukasIO: quardratic would mean that a the simulcast layer for an (origninally) 720p screenshare, would have a maximum bitrate of `50_000` at 3fps and a half the resolution.
In our ScreenSharePresets we have a Video Preset defined with `new VideoPreset(640, 360, 200_000, 3)` which is taking 4times the maxBandwidth than the computed one.
Just want to double check that `50_000` would still make sense

boks1971: For presentation like screen share, that might be enough at that frame rate (amortised over time), but since this is the max, I think we should set it higher. Maybe, the equation is something like `max(K, calculated)` where `K` is some constant. I think 200 Kbps (or even 150 Kbps) is a good number for that.

davidzhao (approved):
looks good!

PR #157, opened 2022-02-22
fix: prefill url with local setup url

lukasIO:


example/sample.ts
```diff
@@ -62,7 +62,7 @@ const appActions = {
     };
 
     const connectOpts: RoomConnectOptions = {
-      autoSubscribe: publishOnly ? false : true,
+      autoSubscribe: !publishOnly,
```
lukasIO, line 65: seems like eslint is making that change automatically...

davidzhao (approved):
👍

PR #159, opened 2022-02-28
handle app in background for mobile devices

lukasIO:
- [x] pause video (camera only) when in background
- [x] reacquire tracks if neccessary (always reaquire video)
- [x] restore muted state after reacquiring
- [x] detect silence when acquiring microphone track and emit event 

for the last point I'm undecided on how to handle it.
I added a helper function `detectSilence` that allows users to pipe in AudioTracks and check if they are silent. It seems wrong to emit an error whenever there is silence detected on reaquiring a track (it might just be, that a user who put the app into background has an external mic switch and that mic is turned off).

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -9,6 +9,8 @@ export default class LocalTrack extends Track {
   sender?: RTCRtpSender;
 
   protected constraints: MediaTrackConstraints;
+  protected isInBackground: boolean;
+  protected reaquireTrack: boolean;
 
   protected constructor(
     mediaTrack: MediaStreamTrack, kind: Track.Kind, constraints?: MediaTrackConstraints,
@@ -17,6 +19,8 @@ export default class LocalTrack extends Track {
     this.mediaStreamTrack.addEventListener('ended', this.handleEnded);
     document.addEventListener('visibilitychange', this.handleAppVisibilityChanged);
     this.constraints = constraints ?? mediaTrack.getConstraints();
+    this.isInBackground = document.visibilityState === 'hidden';
+    this.reaquireTrack = false;
   }
 
   get id(): string {
@@ -120,7 +124,11 @@ export default class LocalTrack extends Track {
   }
 
   protected handleAppVisibilityChanged = () => {
+    this.isInBackground = document.visibilityState === 'hidden';
 
+    if(!this.isInBackground && this.reaquireTrack) {
+      this.restart(this.constraints);
+    }
   };
 
   private handleEnded = () => {
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -22,7 +22,7 @@ export default class LocalTrack extends Track {
   ) {
     super(mediaTrack, kind);
     this.mediaStreamTrack.addEventListener('ended', this.handleEnded);
-    document.addEventListener('visibilitychange', this.handleAppVisibilityChanged); // TODO when to remove this event listener?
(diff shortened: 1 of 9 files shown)


src/room/track/LocalTrack.ts
```diff
@@ -10,12 +11,22 @@ export default class LocalTrack extends Track {
...
+  protected wasMuted: boolean;
+
+  protected reacquireTrack: boolean;
+
   protected constructor(
     mediaTrack: MediaStreamTrack, kind: Track.Kind, constraints?: MediaTrackConstraints,
   ) {
     super(mediaTrack, kind);
     this.mediaStreamTrack.addEventListener('ended', this.handleEnded);
+    document.addEventListener('visibilitychange', this.handleAppVisibilityChanged); // TODO when to remove this event listener?
```
davidzhao, line 25: what about in `stop` ? that indicates that the track will no longer be used.

src/room/utils.ts
```diff
@@ -23,6 +23,10 @@ export function isSafari(): boolean {
   return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
 }
 
+export function isMobile(): boolean {
+  return /Mobile|Android|BlackBerry/.test(navigator.userAgent);
```
davidzhao, line 27: hah blackberry!

davidzhao (approved):
lg Lukas! this should improve our capabilities on mobile significantly.

I agree detectSilence is kind of tricky. why don't we punt it and see how often it happens in practice first?

src/room/utils.ts
lukasIO: haha, yeah.
btw. the regex is string is really concise here, but excludes (for example) iPads and assumes every Android device is a mobile device. 
Thinking about it, for what this PR does it might make sense to include iPads (and tablets more generally), what do you think?

lukasIO:
@davidzhao I realised that pausing (remote) video when a tab is in the background was also on the roadmap, so I added the functionality for it within this PR. Sorry for the PR changing that much after your review, but the two functionalities are closely related and it made sense to refactor the original PR to make it also work for remote video tracks.

src/room/track/RemoteVideoTrack.ts
```diff
@@ -174,7 +181,8 @@ export default class RemoteVideoTrack extends RemoteTrack {
       (prev, info) => Math.max(prev, info.visibilityChangedAt || 0),
       0,
     );
-    const isVisible = this.elementInfos.some((info) => info.visible);
+    const isVisible = this.elementInfos.some((info) => info.visible) && !this.isInBackground;
+    log.info('update visibility', isVisible);
```
davidzhao, line 185: remove?

src/room/utils.ts
davidzhao: yeah I think we should consider platforms that don't have true multi-tasking (so the browser could be "paused") as "mobile" platforms.

src/room/track/Track.ts
```diff
@@ -34,6 +36,8 @@ export class Track extends (EventEmitter as new () => TypedEventEmitter<TrackEve
     this.kind = kind;
     this.mediaStreamTrack = mediaTrack;
     this.source = Track.Source.Unknown;
+    this.isInBackground = document.visibilityState === 'hidden';
+    document.addEventListener('visibilitychange', this.appVisibilityChangedListener);
```
davidzhao, line 40: this is great! CC @davidliu, this will be another area where we'd want to switch for in RN-land.

davidzhao (approved):

PR #162, opened 2022-03-08
fix reconnect not set in resumeconnection

cnderrauber:


davidzhao (approved):

lukasIO (commented):
thanks for catching this!

PR #164, opened 2022-03-13
Ensure certain requests are never queued.

davidzhao:
Also disabled remote video pause for background tabs on desktop

### DIFF
--- a/src/room/track/RemoteVideoTrack.ts
+++ b/src/room/track/RemoteVideoTrack.ts
@@ -1,7 +1,9 @@
 import { debounce } from 'ts-debounce';
 import { TrackEvent } from '../events';
 import { computeBitrate, monitorFrequency, VideoReceiverStats } from '../stats';
-import { getIntersectionObserver, getResizeObserver, ObservableMediaElement } from '../utils';
+import {
+  getIntersectionObserver, getResizeObserver, isMobile, ObservableMediaElement,
+} from '../utils';
 import RemoteTrack from './RemoteTrack';
 import { attachToElement, detachTrack, Track } from './Track';
 import { AdaptiveStreamSettings } from './types';
@@ -166,8 +168,10 @@ export default class RemoteVideoTrack extends RemoteTrack {
   };
 
   protected async handleAppVisibilityChanged() {
-    if (!this.isAdaptiveStream) return;
     await super.handleAppVisibilityChanged();
+    if (!this.isAdaptiveStream) return;
+    // on desktop don't pause when tab is backgrounded
+    if (!isMobile()) return;
     this.updateVisibility();
   }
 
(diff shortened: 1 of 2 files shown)


src/room/track/RemoteVideoTrack.ts
```diff
@@ -166,8 +168,10 @@ export default class RemoteVideoTrack extends RemoteTrack {
   };
 
   protected async handleAppVisibilityChanged() {
-    if (!this.isAdaptiveStream) return;
     await super.handleAppVisibilityChanged();
+    if (!this.isAdaptiveStream) return;
+    // on desktop don't pause when tab is backgrounded
```
lukasIO, line 173: was this causing any issues on desktop?

lukasIO (approved):

src/room/track/RemoteVideoTrack.ts
davidzhao: it caused a visible delay when a user multitasks while in a meeting. It's near-instant against localhost, but a bit jarring when running against a remote server with noticeable latency

lukasIO: makes sense, just as a thought: an alternative to dropping it entirely for desktop would be a setting like `adaptiveStream.pauseWhenBackgrounded` that defaults to false.

davidzhao: yeah, definitely open to it.. let's see if anyone desires this behavior on desktop first?

Just wary about adding features that no one has asked for. The entire tab backgrounding behavior isn't very consistent in desktop systems. (i.e. window fully covered behaves differently than backgrounded tab).

hiroshihorie (approved):

src/room/track/RemoteVideoTrack.ts
ackermann: By the way: We noticed a very strange behavior in Safari where sometimes (some) video streams are only enabled after the window has been either:
- completely covered by another window and then moved back into visibility or 
- minimized and then reopened
i.e. only fully hiding the viewport followed by making it visible again triggered adapativeStream to initially start the stream.
Just wanted to bring this to your attention.

lukasIO: @ackermann Thank you, we've seen some issues in Safari recently that might be connected to what you are seeing. 
Did you by any chance find a way to repro the behaviour you are seeing?

ackermann: @lukasIO it sounds a little odd, but the only thing I think might help increase the chances of this happening is quickly cmd-tabbing between Safari and Chrome (which should fully cover the Safari window) right after joining the call in Safari and then toggling back to Safari before the application finished connecting (i.e. before it got the change to start the stream). It still requires some luck and I have not made the effort to somehow measure it, but it feels like it increases the chance quite a bit. Maybe it's connected to the actual window switching, maybe a CPU/GPU throttling in general would suffice, maybe it's just my imagination – but maybe it helps you too ;-)

PR #167, opened 2022-03-16
Support react native

davidliu:


davidzhao (commented):
looks good! could you install eslint in VS Code and take a look at the lint failures?

src/api/SignalClient.ts
```diff
@@ -20,9 +20,12 @@ import {
   UpdateSubscription, UpdateTrackSettings,
 } from '../proto/livekit_rtc';
 import { ConnectionError } from '../room/errors';
-import { getClientInfo, sleep } from '../room/utils';
 import Queue from './RequestQueue';
 
+if (isWeb()) {
+  import('webrtc-adapter');
```
davidliu, line 26: This dynamic import is the main thing I'm unsure about. The webrtc-adapter library has a bunch of web specific code that crashes on react native, so needs to be omitted for RN. Didn't run into any issues on Chrome, but if there's a more proper way to do this, I'm all ears.

davidzhao: I'm not sure if there's a better way.. will test this for safari and FF to ensure things are still working.

davidzhao (approved):
confirming it works on FF/Safari too.

PR #178, opened 2022-03-28
example: Use location pathname when saving search parameters

bekriebel:
If the sample is hosted from a path other than `/`, it would get reset
to the root path. Set the path based on the current location when adding
the search parameters.


lukasIO (approved):
thanks, that makes sense!

PR #182, opened 2022-03-31
force video play when attaching

lukasIO:
I'm not sure if this will fix all our Safari related problems, but it should mitigate the issues @ackermann reported. 

I was able to reproduce videos not starting when Safari was quickly switched to background shortly before rendering the first video frame. This PR fixes this specific problem.

Still wasn't able to reproduce a freeze frame in Safari.


davidzhao (approved):
nice!

PR #189, opened 2022-04-05
Fix #188

### LINKED ISSUE #188: Unable to use livekit client with http: protocol.
Hi, I unable to use livekit with http: protocol. I have site that serve both on https, and http. Everything works fine on https, but when I open it with http protocol livekit never connected. I don't need something special in case of http protcol I only need for data channels, so I don't want to use media devices. The problem is inside [livekit-client](https://github.com/livekit/client-sdk-js/blob/HEAD/src/room/Room.ts#L293)

```js
        if (isWeb()) {
          window.addEventListener('beforeunload', this.onBeforeUnload);
          navigator.mediaDevices.addEventListener('devicechange', this.handleDeviceChange); **// <-- HERE**
        }

        resolve(this);
```

With http protocol navigator.mediaDevices are not defined, so resolve is never called.

caiiiycuk:
With http protocol navigator.mediaDevices are not defined, so resolve is never called.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -290,7 +290,9 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
         // also hook unload event
         if (isWeb()) {
           window.addEventListener('beforeunload', this.onBeforeUnload);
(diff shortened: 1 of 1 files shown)


lukasIO:
hi @caiiiycuk 
thanks for the fix, makes sense!

PR #190, opened 2022-04-05
refactor all log messages to use same format, overhaul logExtension

lukasIO:
motivation: 
a more structured approach to logging in order to facilitate sending logs to third party services. 

structuring the logs with a message string and an optional context object turned out to be the most flexible approach while not forcing us to rewrite all messages. 
some logging implementations that utilize a similar structure: 
- https://github.com/pinojs/pino/blob/master/docs/api.md
- https://github.com/gajus/roarr
- https://github.com/DataDog/browser-sdk/blob/main/packages/logs/README.md

I opted to go for the ordering of arguments `(msg: string, context?: object)` instead of `(context?: object, msg: string)`, but not sure anymore if it's the right bet, as the implementations mentioned above seem to switch things around.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -137,7 +137,7 @@ export default class LocalParticipant extends Participant {
    * way to manage the common tracks (camera, mic, or screen share)
    */
   private async setTrackEnabled(source: Track.Source, enabled: boolean): Promise<void> {
-    log.debug('setTrackEnabled', source, enabled);
+    log.debug('setTrackEnabled', { source, enabled });
     const track = this.getTrack(source);
     if (enabled) {
       if (track) {
@@ -145,7 +145,7 @@ export default class LocalParticipant extends Participant {
       } else {
         let localTrack: LocalTrack | undefined;
         if (this.pendingPublishing.has(source)) {
-          log.info('skipping duplicate published source', source);
+          log.info('skipping duplicate published source', { source });
           // no-op it's already been requested
           return;
         }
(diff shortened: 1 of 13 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -420,7 +420,7 @@ export default class LocalParticipant extends Participant {
     if (!this.engine.publisher) {
       throw new UnexpectedConnectionState('publisher is closed');
     }
-    log.debug(`publishing ${track.kind} with encodings`, encodings, ti);
+    log.debug(`publishing ${track.kind} with encodings`, { encodings, ti });
```
boks1971, line 423: Maybe call the second one `trackInfo` just to be more clear?

src/room/participant/LocalParticipant.ts
```diff
@@ -490,7 +486,7 @@ export default class LocalParticipant extends Participant {
             this.engine.publisher?.pc.removeTrack(sender);
             this.engine.negotiate();
           } catch (e) {
-            log.warn('unpublishTrack', 'failed to remove track', e);
+            log.warn('failed to remove track', { error: e });
```
boks1971, line 489: Is it possible to get file/line of the caller. I am guessing a log like this is common and having the context of function name helps?

src/room/participant/LocalParticipant.ts
```diff
@@ -618,11 +614,10 @@ export default class LocalParticipant extends Participant {
     }
     const pub = this.videoTracks.get(update.trackSid);
     if (!pub) {
-      log.warn(
-        'handleSubscribedQualityUpdate',
-        'received subscribed quality update for unknown track',
-        update.trackSid,
-      );
+      log.warn('received subscribed quality update for unknown track', {
+        method: 'handleSubscribedQualityUpdate',
```
boks1971, line 618: I see you are using `method` here as an indicator. I am guessing this log line is unique and does not need context. But, it would be nice to have file/line/function name in some way. Not sure that is even possible with TS.

boks1971 (approved):

src/room/participant/LocalParticipant.ts
lukasIO: you are right, forgot to readd the methodName here within the context.

src/room/participant/LocalParticipant.ts
boks1971: Thank you @lukasIO . Guess it can be an improvement (or not :-) ) down  the line.

davidzhao (approved):

PR #191, opened 2022-04-06
Check if mediaDevices exists before using it on disconnect

caiiiycuk:


lukasIO:
@caiiiycuk thanks, would you mind changing the if clause to use optional chaining instead? `navigator.mediaDevices?.removeEventListener`.
To keep it consistent also for when subscribing to the event listener.
I didn't want to bother for a single line change yesterday, but as it is popping up in more places, it feels right to align with the coding style of the repo.

caiiiycuk:
Done

Issue #129, opened 2022-01-19
Publish library on CDN


davidzhao:
To make it easier to just embed it inside of a plain vanilla HTML page

lukasIO:
now available here https://unpkg.com/livekit-client

PR #198, opened 2022-04-08
Bundle library using rollup

lukasIO:
Initial motivation was #129 and that subsequently lead to 
- use rollup as a bundler
- create a umd and esm bundle 
- follow best practices for package.json entry files
- use babel (instead of tsc) for transpilation in order to target browser compatibility more specifically
- remove webpack and replace with rollup dev server for /example

some resources/projects used as inspiration:
- https://vitejs.dev/guide/build.html#library-mode
- https://github.com/davidcetinkaya/embla-carousel/tree/master/packages/embla-carousel
- https://github.com/mrdoob/three.js/tree/dev/utils/build
- https://levelup.gitconnected.com/code-splitting-for-libraries-bundling-for-npm-with-rollup-1-0-2522c7437697

The rollup build process started to show some warning around our `protobuf.js` dependency. Specifically some circular dependencies within that library and the frowned-upon usage of `eval` within the library. 
Rollup seems to deal well with the circular dependencies, to workaround the usage of `eval` there are some annotations/additional context in `rollup.config.js` where basically every instance of `eval` gets replaced with `undefined`.


package.json
```diff
@@ -2,19 +2,29 @@
...
   "license": "Apache-2.0",
   "scripts": {
-    "build": "tsc",
-    "build:watch": "tsc --watch",
+    "build": "rollup --config",
+    "build:watch": "rollup --watch --config rollup.config.js",
     "build-docs": "typedoc",
     "proto": "protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto --ts_proto_opt=esModuleInterop=true --ts_proto_out=./src/proto --ts_proto_opt=outputClientImpl=false,useOptionals=true -I./protocol ./protocol/livekit_rtc.proto ./protocol/livekit_models.proto",
-    "sample": "cd example && webpack serve",
-    "build-sample": "cd example && webpack && cp styles.css index.html dist/",
```
lukasIO, line 17: I removed the `build-sample` command (without `dist` present it was failing for me) but can re-add with a rollup equivalent if that command still makes sense/we want to continue to provide it

davidzhao (approved):
This is really fantastic. I'm reading about rollups and have learned something new today!

davidliu (approved):
So, something with rollup is screwing up react-native. Can't figure out what it is, but the javascript engine it's using doesn't like it. When debugging (which uses the computer's browser javascript engine), it works again.

For now, since react-native is still in alpha, I think we'll lock down the livekit-client versions that it can use until we figure out a more permanent solution.

PR #200, opened 2022-04-13
specify document as root for intersection observer

lukasIO:
addresses https://github.com/livekit/livekit-react/issues/41

here's an online reproduction for safari showcasing the fix in an isolated context: 
https://svelte.dev/repl/d1acaeee97f44b11bddff072ea3a6d5a?version=3.47.0

if you change `root: document` to `root: null` you can see that isIntersecting displays `false`. Setting it back to `document` shows `true` for isIntersecting.
Behaviour doesn't change in chrome.

unfortunately I couldn't reproduce https://github.com/livekit/livekit-react/issues/41 on my machine.
This fix feels like a good move regardless in order to get consistent behaviour for all browsers.

### DIFF
--- a/src/room/utils.ts
+++ b/src/room/utils.ts
@@ -54,7 +54,11 @@ export const getResizeObserver = () => {
 
 let intersectionObserver: IntersectionObserver | null = null;
 export const getIntersectionObserver = () => {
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
fantastic work!

PR #203, opened 2022-04-19
Send adaptive stream as a param in join request

boks1971:


src/api/SignalClient.ts
```diff
@@ -485,5 +489,9 @@ function createConnectionParams(token: string, info: ClientInfo, opts?: ConnectO
     params.set('publish', opts.publishOnly);
   }
 
+  if (opts?.adaptiveStream !== undefined) {
```
lukasIO, line 492: is it intended that we have a distinct `adaptiveStream=0` state and if so, would it make sense to always default to that? 
the question being shouldn't we just omit the adaptiveStream param if it is false?

src/options.ts
```diff
@@ -73,6 +73,9 @@ export interface RoomConnectOptions {
    * publish only mode
    */
   publishOnly?: string;
+
+  /** adapt subscribed video streams to visibility/video tile size, defaults to false */
+  adaptiveStream?: boolean;
```
lukasIO, line 78: I fear it might be a bit confusing to have two options for adaptiveStream. Once in RoomOptions and another one in ConnectOptions. 
If we leave them as a room option only we could extract it on each connect call from the room options object, wdyt?

src/api/SignalClient.ts
boks1971: We do not need to add `adaptiveStream=0` explicitly? Maybe change this to
```
if (opts?.adaptiveStream) {                                                                 
    params.set('adaptive_stream', '1');                                         
}
```
What do you think?

src/options.ts
boks1971: That's where I was going initially. But, I got all tangled up. Let me have another look.

boks1971:
@lukasIO Let me know how this commit looks - https://github.com/livekit/client-sdk-js/pull/203/commits/940de8e3163daf404bdac82bf7d53f4a5be16c1d. Actually, might be easier to read the whole PR now as it is simpler :-)

PR #205, opened 2022-04-21
add method for pausing and resuming track upstream (aka server-side-only-mute)

lukasIO:
Usecase: stop track from consuming upstream bandwidth, but keep it running locally (all with minimal time to resume).
Track gets paused for all other participants.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -625,12 +625,12 @@ export default class LocalParticipant extends Participant {
   };
 
   private onTrackUpstreamPaused = (track: LocalTrack) => {
-    log.info('upstream paused');
+    log.debug('upstream paused');
     this.onTrackMuted(track, true);
   };
 
   private onTrackUpstreamResumed = (track: LocalTrack) => {
-    log.info('upstream resumed');
+    log.debug('upstream resumed');
     this.onTrackMuted(track, track.isMuted);
   };
 
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -385,8 +385,8 @@ export default class LocalParticipant extends Participant {
     track.on(TrackEvent.Muted, this.onTrackMuted);
     track.on(TrackEvent.Unmuted, this.onTrackUnmuted);
     track.on(TrackEvent.Ended, this.onTrackUnpublish);
-    track.off(TrackEvent.UpstreamHalted, this.onTrackUpstreamHalted);
-    track.off(TrackEvent.UpstreamResumed, this.onTrackUpstreamResumed);
+    track.on(TrackEvent.UpstreamHalted, this.onTrackUpstreamHalted);
+    track.on(TrackEvent.UpstreamResumed, this.onTrackUpstreamResumed);
 
     // create track publication from track
     const req = AddTrackRequest.fromPartial({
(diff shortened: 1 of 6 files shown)


src/room/track/LocalTrack.ts
```diff
@@ -162,4 +168,22 @@ export default class LocalTrack extends Track {
...
     this.emit(TrackEvent.Ended, this);
   };
+
+  async pauseUpstream() {
+    this.emit(TrackEvent.UpstreamPaused, this);
+    this._isUpstreamPaused = true;
+    if (!this.sender) {
+      throw new TrackInvalidError('unable to pause upstream for an unpublished track');
+    }
+    await this.sender.replaceTrack(getEmptyMediaStreamTrack());
```
lukasIO, line 178: additionally to signalling the server a pause event, this is replacing the current mediastreamtrack with an empty one. (we keep the original mediastreamtrack untouched however to quickly resume to it).
Not sure if makes sense to keep the empty track inside or if the server-side-mute would also stop the the video data flowing upstream to the server (e.g. if dynacast is off).

src/room/track/LocalTrack.ts
```diff
@@ -162,4 +168,22 @@ export default class LocalTrack extends Track {
     }
     this.emit(TrackEvent.Ended, this);
   };
+
+  async pauseUpstream() {
+    this.emit(TrackEvent.UpstreamPaused, this);
```
davidzhao, line 173: should we check if it's already paused and skip the event if already paused?

src/room/participant/LocalParticipant.ts
```diff
@@ -620,6 +624,16 @@ export default class LocalParticipant extends Participant {
     this.engine.updateMuteStatus(track.sid, muted);
   };
 
+  private onTrackUpstreamPaused = (track: LocalTrack) => {
+    log.info('upstream paused');
```
davidzhao, line 628: should this be a debug level log?

src/room/track/LocalTrack.ts
davidzhao: this is neat! using empty track should be fine here. server will not be able to pause the upstream if MediaStreamTrack keeps on producing.

src/room/track/LocalTrack.ts
```diff
@@ -162,4 +168,22 @@ export default class LocalTrack extends Track {
     }
     this.emit(TrackEvent.Ended, this);
   };
+
+  async pauseUpstream() {
```
davidzhao, line 172: since it's public, could we add a bit of docs explaining what it's used for?

davidzhao (approved):
this is a clean API to accomplish this use case. well done!

boks1971 (approved):

src/room/utils.ts
```diff
@@ -75,3 +75,21 @@ export function getClientInfo(): ClientInfo {
   });
   return info;
 }
+
+let emptyMediaStreamTrack: MediaStreamTrack | undefined;
+
+export function getEmptyMediaStreamTrack() {
```
lukasIO, line 81: @davidliu do you have an idea how we could solve this for the react-native client? 
I guess the canvas-hack to get an empty media stream track won't work there...

davidliu: We can just create blank tracks on the native side. There's no methods available for it currently, but it's not a huge lift to make.

davidliu:
What does this do over the basic `enabled = false` on the track?

davidzhao:
> What does this do over the basic `enabled = false` on the track?

It keeps the track active (so that a client could render video locally). but to the server and remote participants, it would be muted.  This handles a specific case where you'd want to mute only the remote track.

PR #212, opened 2022-04-28
Fix bundling issues

lukasIO:
Trying to resolve a bug which a user reported in the build process of vite lead to nextjs builds failing for another user. 

The best practices to support both es modules and commonjs still seem to have pitfalls here and there. 
Two approaches that look the most promising: 
- use the `.mjs` extension for the es module to make it unmistakably clear that it actually is a module
- create a package.json file as part of the build step and place that into a folder next to the bundled module. This would mean we need a file structure like `/dist/esm/bundle.js` with `/dist/esm/package.json` right next to it. setting `type: module` in this package.json file would then only affect the bundle.js file in the /esm folder and signal to other bundlers that they should treat it as a module.

This PR implements the first solution and is tested with the latest nextjs and sveltekit frameworks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "livekit-client",
-  "version": "0.18.3",
+  "version": "0.18.4-RC1",
   "description": "JavaScript/TypeScript client SDK for LiveKit",
   "main": "./dist/livekit-client.umd.js",
   "unpkg": "./dist/livekit-client.umd.js",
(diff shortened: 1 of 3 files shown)


davidzhao (approved):

davidzhao:
I didn't quite follow the details of the issue. but I agree that this seems better than the alternative

lukasIO:
for context, this is a writeup outlining the pitfalls https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html

Issue #213, opened 2022-04-29
Vue project  error Module parse failed: Unexpected token


gzh6473:
In the following snippet, the inability to parse results in an exception, the error is Module parse failed: Unexpected token

`const opts = {
        ...options,
    };`

`const opts = {
            ...(_a = this.roomOptions) === null || _a === void 0 ? void 0 : _a.publishDefaults,
            ...options,
        };`

`this.options.audioCaptureDefaults = {
            ...audioDefaults,
            ...options === null || options === void 0 ? void 0 : options.audioCaptureDefaults,
        };
        this.options.videoCaptureDefaults = {
            ...videoDefaults,
            ...options === null || options === void 0 ? void 0 : options.videoCaptureDefaults,
        };
        this.options.publishDefaults = {
            ...publishDefaults,
            ...options === null || options === void 0 ? void 0 : options.publishDefaults,
        };`

lukasIO:
hi @gzh6473,

could you share what version of vue you are using so that we can reproduce and find the right fix?

lukasIO:
@gzh6473 could you try with version `0.18.4-RC5` and see if it fixes your issues?

gzh6473:
> hi @gzh6473,
> 
> could you share what version of vue you are using so that we can reproduce and find the right fix?

2.5.2

PR #218, opened 2022-05-03
Fix target browser support

lukasIO:
With recent issues like https://github.com/livekit/client-sdk-js/issues/213 around supported syntax, this change introduces an additional babel plugin to transform object rest spreads.
This should broaden compatibility even wider, without any significant drawbacks.

The babel config stays the same otherwise (default browserlist targets). By explicitly passing `ts` as a file extension to the babel rollup plugin, this PR makes babel also run on typescript input files, which previously lead to babel not doing its job properly for us, as all our input files are typescript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "livekit-client",
-  "version": "0.18.3",
+  "version": "0.18.4-RC1",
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
nice!

PR #221, opened 2022-05-06
add av1 svc codec

cnderrauber:
add av1 svc codec support, only enabled with chrome launch with --force-fieldtrials=WebRTC-DependencyDescriptorAdvertised/Enabled/

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -69,7 +69,12 @@
               </select>
             </div>
             <div>
-              <select id="preferred-quality" class="custom-select" style="width: auto" onchange="appActions.handlePreferredQuality(event)">
+              <select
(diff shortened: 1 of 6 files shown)


cnderrauber:
> I opened a PR against this branch, because the layout started breaking with more options (#222) feel free to merge it in here if that makes sense.
> 
> Wondering if the "Preferred Video Quality" selector should be further down inside the `actions-area`. I assumed the options on top are room/connect time options and preferred video quality seems to be changeable on the fly while already connected.

Make sense, need rearrange it to the actions

cnderrauber:
@lukasIO updated, will merge this pr after  your pr merged

PR #223, opened 2022-05-06
Restrict access to mediaStreamTrack when using adaptiveStream

lukasIO:
breaking change: track.mediaStreamTrack is now readonly and will throw an error when accessed on `RemoteVideoTrack`s without attached elements and adaptiveStream is being used.

unclear what to do about `track.mediaStream` which is there to support react native. I marked it as `internal`, but the footgun is still there, when doing something like `track.mediaStream.getVideoTracks()`. 
Should we do the same thing for `mediaStream` as this PR is doing for `mediaStreamTrack`? 
@davidliu does/will react-native support adaptiveStream?


davidzhao (approved):
great move! I think it's fine to leave mediaStream available.

PR #224, opened 2022-05-10
Fix robotic sounding audio for FF 100

davidzhao:
With FF 100, if channelCount is set to 1, the resulting audio sounds
robotic. Leaving it to defaults resolves the issue.


boks1971 (approved):

jibon57:
@davidzhao thank you. It's working perfectly! :tada:

PR #226, opened 2022-05-10
Abort connection attempt when calling disconnect while connecting

lukasIO:
Still need to test if this will solve the react-strict mode related problems.


davidzhao (commented):
solid strategy! are you thinking that the Room object is still one-time use? i.e. it would get thrown a way and a new one created? Or would we re-use it across the same `useRoom` hook?

davidzhao (approved):

davidzhao (approved):
great fix!

PR #229, opened 2022-05-12
Apply participant volume when microphone track is added

lukasIO:
Previously if `participant.setVolume` was called and no tracks were present, the volume info was lost.

### DIFF
--- a/src/room/participant/RemoteParticipant.ts
+++ b/src/room/participant/RemoteParticipant.ts
@@ -19,6 +19,8 @@ export default class RemoteParticipant extends Participant {
 
   signalClient: SignalClient;
 
+  private volume?: number;
+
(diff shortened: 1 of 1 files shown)


theomonnom (approved):

lukasIO:
definitely good to think about what the desired behaviour should be. should participant volume always override track volume? I assume yes.

the change in this PR wouldn't cause an issue with it though as the track is being created just before in 
https://github.com/livekit/client-sdk-js/blob/fba99753df8d679e553d3d0f1554d98c20c738d8/src/room/participant/RemoteParticipant.ts#L149
so there is no way of a user to have called setVolume on that track before

davidzhao (approved):
lg!

PR #232, opened 2022-05-16
Rename RoomState to ConnectionState

lukasIO:
`ConnectionState` is more descriptive about what kind of state it actually holds and we already have a `RoomState` type in the livekit-react repo, so this additionally avoids confusion. 
Old name is just deprecated for now and not yet removed.

### DIFF
--- a/src/room/events.ts
+++ b/src/room/events.ts
@@ -29,7 +29,7 @@ export enum RoomEvent {
   /**
    * Whenever the connection state of the room changes
    *
-   * args: ([[RoomState]])
+   * args: ([[ConnectionState]])
    */
   StateChanged = 'stateChanged',
 
--- a/src/room/events.ts
+++ b/src/room/events.ts
@@ -31,7 +31,12 @@ export enum RoomEvent {
    *
    * args: ([[ConnectionState]])
    */
-  StateChanged = 'stateChanged',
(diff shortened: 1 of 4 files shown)


src/room/events.ts
```diff
@@ -29,9 +29,14 @@ export enum RoomEvent {
...
-   * args: ([[RoomState]])
+   * args: ([[ConnectionState]])
    */
-  StateChanged = 'stateChanged',
+  ConnectionStateChanged = 'connectionStateChanged',
+
+  /**
+   * @deprecated StateChanged has been renamed to ConnectionStateChanged
+   */
+  StateChanged = 'connectionStateChanged',
```
davidzhao, line 39: with 1.0 change, we could remove this? since it won't be fired anywhere.

davidzhao (approved):

src/room/events.ts
lukasIO: only thing to keep in mind is that if we don't release another version before 1.0 (likely) it will simply break people's apps if they are currently using `RoomEvent.StateChanged`.

davidzhao: right.. you mean their existing app would fail to compile if they were using this. I guess it's fine if we left it in there until 1.1.

PR #234, opened 2022-05-20
Add adaptiveStreamSetting for pausing video when tab in background

lukasIO:
Adds an additional setting to adaptive stream for pausing video when the browser tab is in the background/hidden.
Defaults to true.

`adaptiveStream: { pauseVideoInBackground: true}`

### DIFF
--- a/src/room/track/RemoteVideoTrack.ts
+++ b/src/room/track/RemoteVideoTrack.ts
@@ -1,12 +1,7 @@
 import { debounce } from 'ts-debounce';
 import { TrackEvent } from '../events';
 import { computeBitrate, monitorFrequency, VideoReceiverStats } from '../stats';
-import {
-  getIntersectionObserver,
-  getResizeObserver,
-  isMobile,
-  ObservableMediaElement,
-} from '../utils';
+import { getIntersectionObserver, getResizeObserver, ObservableMediaElement } from '../utils';
 import RemoteTrack from './RemoteTrack';
 import { attachToElement, detachTrack, Track } from './Track';
 import { AdaptiveStreamSettings } from './types';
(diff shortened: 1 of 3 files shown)


src/room/track/Track.ts
```diff
@@ -179,8 +183,20 @@ export class Track extends (EventEmitter as new () => TypedEventEmitter<TrackEve
...
+  protected appVisibilityChangedListener = () => {
+    if (this.backgroundTimeout) {
+      clearTimeout(this.backgroundTimeout);
+    }
+    // delay app visibility update if it goes to hidden
+    // update immediately if it comes back to focus
+    if (document.visibilityState === 'hidden') {
+      this.backgroundTimeout = setTimeout(
+        () => this.handleAppVisibilityChanged(),
+        BACKGROUND_REACTION_DELAY,
```
davidzhao, line 195: is the purpose of the timeout so that if you are switching tabs, it doesn't pause too aggressively?

lukasIO: excactly. delay pausing when going into hidden state, but jump straight back if leaving hidden state

src/room/track/RemoteVideoTrack.ts
```diff
@@ -188,8 +183,6 @@ export default class RemoteVideoTrack extends RemoteTrack {
   protected async handleAppVisibilityChanged() {
     await super.handleAppVisibilityChanged();
     if (!this.isAdaptiveStream) return;
-    // on desktop don't pause when tab is backgrounded
-    if (!isMobile()) return;
```
lukasIO, line 192: @davidzhao this most likely caused the safari visibility bug. Am not able to reproduce the issue on this branch.

lukasIO:
instead of a boolean we could also let this setting have three different options like `mobile | always | never` in order to differentiate between desktop and mobile usage.

davidzhao (approved):
lg!

PR #235, opened 2022-05-20
Refactor ElementInfo for extension

davidliu:
Allows for RN to submit their own elements for observing.

### DIFF
--- a/src/room/track/RemoteVideoTrack.ts
+++ b/src/room/track/RemoteVideoTrack.ts
@@ -86,7 +86,7 @@ export default class RemoteVideoTrack extends RemoteTrack {
       this.elementInfos.find((info) => info.element === element) === undefined
     ) {
 
-      let elementInfo = new HTMLElementInfo(element);
+      const elementInfo = new HTMLElementInfo(element);
       elementInfo.handleResize = () => { this.debouncedHandleResize() }
       elementInfo.handleVisibilityChanged = () => { this.updateVisibility() }
       this.elementInfos.push(
@@ -117,7 +117,6 @@ export default class RemoteVideoTrack extends RemoteTrack {
       this.elementInfos.push(
         elementInfo
       );
-      console.log("observe: elementInfos: " + this.elementInfos)
       elementInfo.observe();
       this.debouncedHandleResize();
       this.updateVisibility();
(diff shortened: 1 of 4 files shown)


src/room/track/RemoteVideoTrack.ts
```diff
@@ -85,24 +85,45 @@ export default class RemoteVideoTrack extends RemoteTrack {
...
+    }
+    this.hasUsedAttach = true;
+    return element;
+  }
 
+  /**
+   * Observe an ElementInfo for changes when adaptive streaming.
+   * @param elementInfo
+   */
+  observeElementInfo(elementInfo: ElementInfo) {
```
lukasIO, line 99: wondering if these new methods should be part of the public API or not. 
I assume RN would call `observeElementInfo` directly instead of calling `attach()`, right?

If they need to be public for RN, we could still mark the methods with `@internal`.

lukasIO (approved):
This is great! Just one comment.

src/room/track/RemoteVideoTrack.ts
davidliu: `@internal` sounds good. Does it only hide it from documentation or are there any other effects?

lukasIO: Yeah, it will not appear in the documentation.

Additionally it has different effects depending on how we set up type resolvers. 
If we resolve our types against emitted `.d.ts` (which was the case previously) then methods marked with `@internal` would not get included in those type files and which would lead to your IDE showing an error if you try to call that method.
But currently we have TS source maps in place which let you jump to the original source code. In this case nothing happens apart from your IDE showing an `@internal` hint.

src/room/track/RemoteVideoTrack.ts
```diff
@@ -85,24 +85,46 @@ export default class RemoteVideoTrack extends RemoteTrack {
...
       // if the tab is backgrounded, the initial resize event does not fire until
       // the tab comes into focus for the first time.
       this.debouncedHandleResize();
+      this.updateVisibility();
     }
-    this.hasUsedAttach = true;
-    return element;
+  }
+
+  stopObservingElementInfo(elementInfo: ElementInfo) {
```
lukasIO, line 121: I think we should also mark this as internal

davidliu: Oh nice catch, forgot that.

PR #240, opened 2022-05-27
simulcast codecs support

cnderrauber:


example/index.html
```diff
@@ -68,6 +68,19 @@ <h2>Livekit Sample App</h2>
                 <option value="av1">AV1</option>
               </select>
             </div>
+            <div>
```
lukasIO, line 71: `handlePreferredQuality` is duplicated now, it also exists further down in the actions

src/room/track/LocalVideoTrack.ts
```diff
@@ -248,6 +276,65 @@ export default class LocalVideoTrack extends LocalTrack {
   }
 }
 
+async function setPublishingLayersForSender(
+  sender: RTCRtpSender,
+  senderEncodings: RTCRtpEncodingParameters[],
+  qualities: SubscribedQuality[],
+) {
+  console.log('setPublishingLayersForSender', sender, qualities, senderEncodings);
```
lukasIO, line 284: we should probably use our log method here instead of the `console.log`

lukasIO (approved):
didn't test the actual functionality, just went through the diff and there are just some nit picks.

src/room/participant/LocalParticipant.ts
```diff
@@ -409,16 +415,38 @@ export default class LocalParticipant extends Participant {
...
+        const simOpts = { ...opts };
+        simOpts.simulcast = true;
+        simOpts.scalabilityMode = undefined;
+        simEncodings = computeVideoEncodings(
+          track.source === Track.Source.ScreenShare,
+          width,
+          height,
+          simOpts,
+        );
+        const simulcastTrack = track.addSimulcastTrack(compatibleCodecForSVC, simEncodings);
```
davidzhao, line 436: perhaps we should use h.264 as the backup codec? thinking for performance reasons, at least H.264 could leverage HW encoder.

src/room/participant/LocalParticipant.ts
```diff
@@ -427,7 +455,7 @@ export default class LocalParticipant extends Participant {
         height,
         opts,
       );
-      req.layers = videoLayersFromEncodings(req.width, req.height, encodings);
+      req.layers = videoLayersFromEncodings(req.width, req.height, simEncodings ?? encodings);
```
davidzhao, line 458: if using av1/vp9, our default bitrates will be wrong. we'll need to test and compute reasonable bitrates for AV1 and/or VP9 for all of the resolutions we are targeting.

src/room/participant/LocalParticipant.ts
```diff
@@ -714,6 +763,7 @@ export default class LocalParticipant extends Participant {
     }
     const cap = RTCRtpSender.getCapabilities(kind);
     if (!cap) return;
+    log.info('get capabilities', cap);
```
davidzhao, line 766: this should be removed and/or is a debug log.

speaking of which, how do we handle cases where user desires AV1, but browser does not support DD? how is that detection performed?

src/room/track/LocalVideoTrack.ts
```diff
@@ -43,8 +64,15 @@ export default class LocalVideoTrack extends LocalTrack {
   }
 
   stop() {
+    log.info(`stop track ${this.mediaStreamTrack.id}`);
```
davidzhao, line 67: if this is for debugging, let's use debug level. would be ideal to avoid info level logs unless it's directly related to end-user action.

src/room/track/RemoteTrackPublication.ts
```diff
@@ -202,6 +202,7 @@ export default class RemoteTrackPublication extends TrackPublication {
       settings.quality = VideoQuality.HIGH;
     }
 
+    log.debug('emit settings', settings);
```
davidzhao, line 205: I believe we are logging this twice? once here and once in RemoteParticipant.

src/room/track/RemoteVideoTrack.ts
```diff
@@ -144,6 +144,11 @@ export default class RemoteVideoTrack extends RemoteTrack {
     return detachedElements;
   }
 
+  /** @internal */
+  getDecoderImplementation(): string {
```
davidzhao, line 148: nit: avoid returning empty string when we don't know the answer

```suggestion
  getDecoderImplementation(): string | undefined {
```

davidzhao:
@davidliu added you as a reviewer too to confirm there's nothing here that'll break react-native. We won't need to worry about AV1 for RN for awhile.. just wanted to make sure this PR doesn't break it.

src/room/participant/LocalParticipant.ts
```diff
@@ -456,9 +484,25 @@ export default class LocalParticipant extends Participant {
...
+      track.codec = opts.videoCodec;
     }
+
+    const localTrack = track as LocalVideoTrack;
+    simulcastTracks?.forEach((simulcastTrack) => {
+      const simTransceiverInit: RTCRtpTransceiverInit = { direction: 'sendonly' };
+      if (simulcastTrack.encodings) {
+        simTransceiverInit.sendEncodings = simulcastTrack.encodings;
+      }
+      const simTransceiver = this.engine.publisher!.pc.addTransceiver(
```
davidliu, line 498: include an `await` on addTransceiver for RN compatibility.

davidliu (approved):

src/room/participant/LocalParticipant.ts
cnderrauber: good idea, will file a issue on this

davidzhao (approved):

PR #242, opened 2022-05-31
Apply track permissions automatically on connect/reconnect/resume

lukasIO:
Allows to set permissions prior to a successful `room.connect()` call and applies them once the engine is connected. 
Also re-applies permissions after `EngineEvent.Resumed` and `EngineEvent.Reconnected`.


davidzhao (approved):
lg!

PR #243, opened 2022-06-01
Improve initial connection speed with protocol 8

cnderrauber:
client change for https://github.com/livekit/livekit/pull/747

### DIFF
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -158,11 +158,8 @@ const appActions = {
       startTime = Date.now();
       await room.connect(url, token, connectOptions);
       const elapsed = Date.now() - startTime;
-      const signalStableElaspsed = room.engine.signalStableTimestamp - startTime;
       appendLog(
-        `successfully connected to ${room.name} in ${Math.round(
-          elapsed,
-        )}ms, signal cost ${signalStableElaspsed}ms`,
+        `successfully connected to ${room.name} in ${Math.round(elapsed)}ms`,
         room.engine.connectedServerAddress,
       );
     } catch (error: any) {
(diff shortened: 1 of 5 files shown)


src/room/RTCEngine.ts
```diff
@@ -93,6 +93,8 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private connectedServerAddr?: string;
 
+  private signalStableTime: number = 0;
```
cnderrauber, line 96: this field for statistics, do we need this?

example/sample.ts
```diff
@@ -144,14 +139,30 @@ const appActions = {
...
         },
-      );
+      )
+      .on(RoomEvent.TrackSubscribed, (_1, _2, participant: RemoteParticipant) => {
+        renderParticipant(participant);
+        renderScreenShare();
+      })
+      .on(RoomEvent.SignalConnected, async () => {
+        if (shouldPublish) {
+          await Promise.all([
```
cnderrauber, line 149: move publish to SignalConnnect event handle, a slight improve on publish speed.

example/sample.ts
```diff
@@ -505,12 +516,13 @@ function renderParticipant(participant: Participant, remove: boolean = false) {
...
-      const startTime = Date.now();
+      const renderStartTime = Date.now();
       // measure time to render
       videoElm.onloadeddata = () => {
-        const elapsed = Date.now() - startTime;
+        const elapsed = Date.now() - renderStartTime;
+        const fromJoin = Date.now() - startTime;
         appendLog(
-          `RemoteVideoTrack ${cameraPub?.trackSid} (${videoElm.videoWidth}x${videoElm.videoHeight}) rendered in ${elapsed}ms`,
+          `RemoteVideoTrack ${cameraPub?.trackSid} (${videoElm.videoWidth}x${videoElm.videoHeight}) rendered in ${elapsed}ms, total cost ${fromJoin}ms from join`,
```
davidzhao, line 525:
```suggestion
          `RemoteVideoTrack ${cameraPub?.trackSid} (${videoElm.videoWidth}x${videoElm.videoHeight}) rendered in ${elapsed}ms, ${fromJoin}ms from start`,
```

example/sample.ts
```diff
@@ -531,6 +543,12 @@ function renderParticipant(participant: Participant, remove: boolean = false) {
   if (micEnabled) {
     if (!(participant instanceof LocalParticipant)) {
       // don't attach local audio
+      audioELm.onloadeddata = () => {
+        const fromJoin = Date.now() - startTime;
+        appendLog(
+          `RemoteAudioTrack ${micPub?.trackSid} rendered total cost ${fromJoin}ms from join`,
```
davidzhao, line 549:
```suggestion
          `RemoteAudioTrack ${micPub?.trackSid} played ${fromJoin}ms from start`,
```

src/room/RTCEngine.ts
davidzhao: I'm not sure this is meaningful for external users. IMO what's meaningful is the end to end connection time. so I vote to not add this.

example/sample.ts
lukasIO: yeah, publishing within the connect function would be great! 
I am not sure that simply having `camera: true, microphone: true` would suffice most of the scenarios.
I think it would be great if this speed up could be leveraged also for more complex setups, either by passing in localTracks directly or using `VideoCaptureOptions`. 
Ideally it would allow for all three scenarios.

davidzhao: sounds good, we can come up with a better API after this PR is merged.

davidzhao (approved):

PR #244, opened 2022-06-02
Improved handling of full reconnect

davidzhao:
Previously, we did not handle computer going to sleep. There are a few tricky cases that caused a errors during resume to go undetected.

Also improved handling of duplicate participants to enforce unique identity on the client side.


lukasIO (approved):
this is great :)

PR #249, opened 2022-06-05
Moving StreamState to Track for improved usability

davidzhao:
Similarly to `isSubscribed`, declaring StreamState on both local and
remote tracks to ensure application UI doesn't have to differentiate
between local and remote tracks.


hiroshihorie:
Should it be always `active` for local tracks ?

davidzhao:
> Should it be always `active` for local tracks ?

yup, it should!

PR #262, opened 2022-06-10
add browser name; refine preferred codec

cnderrauber:


src/room/utils.ts
```diff
@@ -67,11 +67,33 @@ export interface ObservableMediaElement extends HTMLMediaElement {
...
+
+  return browser;
+}
+
 export function getClientInfo(): ClientInfo {
   const info = ClientInfo.fromPartial({
     sdk: ClientInfo_SDK.JS,
     protocol: protocolVersion,
     version,
+    browser: browserName(),
```
davidzhao, line 96: does it have to be set on the client side? the server should determine this from UserAgent string if possible.

cnderrauber: removed

davidzhao (approved):
Looks good. can you also add a `yarn changeset` comment describing the change? it'll make generating a changelog easier

cnderrauber:
> Looks good. can you also add a `yarn changeset` comment describing the change? it'll make generating a changelog easier
Done

Issue #261, opened 2022-06-09
Feature request: Make reconnection logic configureable


coonmoo:
First of all thanks for the great work here!

We have a use-case where the client can be disconnected from the network for more than 60 seconds.
In this case the livekit client gives up reconnecting and stops all tracks.

Instead of hardcoding it here: https://github.com/livekit/client-sdk-js/blob/a492d213ac9c2e20caaf7147cb4930c65bff839b/src/room/RTCEngine.ts#L28

I would like to be able to configure it myself & specify if the tracks should be stopped.
[Example how SignalR handles that.](https://github.com/dotnet/aspnetcore/blob/8b30d862de6c9146f466061d51aa3f1414ee2337/src/SignalR/clients/ts/signalr/src/DefaultReconnectPolicy.ts)

I'm looking for something like this

```
export class MyReconnectPolicy implements IRetryPolicy {
   public nextRetryDelayInMilliseconds(retryContext: RetryContext): number | null {
      if (retryContext.previousRetryCount > 10) {
         retryContext.stopTracks = false;
         return null;
      }

      return retryContext.previousRetryCount * 1000;
   }
}

```

```
this.room = new Room({
 reconnectPolicy: new MyReconnectPolicy()
});
```

davidzhao:
yeah, this is a great suggestion! we could provide a few common reconnect policies in the SDK while making it easy for users to override that behavior.

are you interested in creating a PR for this?

PR #266, opened 2022-06-12
Reconnect policy

taskbit:
My motivation was that the existing reconnection logic gave up reconnecting after the LiveKit node was forcefully restarted.
Also in response to #261 I revisited the reconnection handling.

I run into an issue where the  websocket connection is being closed after trying to resume with `?reconnect=1` 
probably initiated on the server side.

I added a workaround [here](https://github.com/taskbit/client-sdk-js/blob/47ca91f95f2f325d9319cc9821fda3f1d1c0b7d1/src/room/RTCEngine.ts#L584)

### DIFF
--- a/src/room/IReconnectPolicy.ts
+++ b/src/room/IReconnectPolicy.ts
@@ -23,17 +23,3 @@ export interface IReconnectContext {
    */
   readonly retryReason?: Error;
 }
-
-export class ReconnectContext implements IReconnectContext {
-  public readonly elapsedMs: number;
-
-  public readonly retryCount: number;
-
-  public readonly retryReason?: Error;
-
-  constructor(elapsedMilliseconds: number, retryCount: number, retryReason?: Error) {
-    this.elapsedMs = elapsedMilliseconds;
-    this.retryCount = retryCount;
-    this.retryReason = retryReason;
-  }
-}
--- a/src/room/IReconnectPolicy.ts
+++ b/src/room/IReconnectPolicy.ts
@@ -1,0 +1,39 @@
+/** Controls reconnecting of the client */
+export interface IReconnectPolicy {
+  /** Called after disconnect was detected
+   *
+   * @returns {number | null} Amount of time in milliseconds to delay the next reconnect attempt, `null` signals to stop retrying.
+   */
+  nextRetryDelayInMs(context: IReconnectContext): number | null;
+}
+
+export interface IReconnectContext {
+  /**
+   * Number of failed reconnect attempts
+   */
+  readonly retryCount: number;
+
+  /**
+   * Elapsed amount of time in milliseconds since the disconnect.
+   */
+  readonly elapsedMs: number;
+
+  /**
+   * Reason for retrying
+   */
+  readonly retryReason?: Error;
+}
+
+export class ReconnectContext implements IReconnectContext {
+  public readonly elapsedMs: number;
+
+  public readonly retryCount: number;
+
+  public readonly retryReason?: Error;
+
+  constructor(elapsedMilliseconds: number, retryCount: number, retryReason?: Error) {
+    this.elapsedMs = elapsedMilliseconds;
+    this.retryCount = retryCount;
+    this.retryReason = retryReason;
+  }
+}
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -23,7 +23,7 @@ import { EngineEvent } from './events';
 import PCTransport from './PCTransport';
 import { isFireFox, isWeb, sleep } from './utils';
 import { RoomOptions } from '../options';
-import { IReconnectContext, IReconnectPolicy } from './IReconnectPolicy';
+import { ReconnectContext, ReconnectPolicy } from './ReconnectPolicy';
 import DefaultReconnectPolicy from './DefaultReconnectPolicy';
 
 const lossyDataChannel = '_lossy';
@@ -96,7 +96,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private attemptingReconnect: boolean = false;
 
-  private reconnectPolicy: IReconnectPolicy;
+  private reconnectPolicy: ReconnectPolicy;
 
   constructor(private options: RoomOptions) {
     super();
@@ -512,7 +512,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }, delay);
   };
 
-  private getNextRetryDelay(context: IReconnectContext) {
+  private getNextRetryDelay(context: ReconnectContext) {
     try {
       return this.reconnectPolicy.nextRetryDelayInMs(context);
     } catch (e) {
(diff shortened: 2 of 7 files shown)


src/room/RTCEngine.ts
```diff
@@ -448,60 +473,57 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+          await this.resumeConnection(signalEvents);
         }
         this.reconnectAttempts = 0;
         this.fullReconnectOnNext = false;
       } catch (e) {
+        log.debug('received reconnect error', { error: e });
         this.reconnectAttempts += 1;
-        let reconnectRequired = false;
-        let recoverable = true;
-        if (e instanceof UnexpectedConnectionState) {
```
davidzhao, line 461: Can you share the thought process behind this change?

src/room/RTCEngine.ts
```diff
@@ -448,60 +473,57 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+            this.fullReconnectOnNext = true;
+            requireSignalEvents = true;
+          }
         }
 
-        const duration = Date.now() - this.reconnectStart;
-        if (this.reconnectAttempts >= maxReconnectRetries || duration > maxReconnectDuration) {
-          recoverable = false;
-        }
+        duration = Date.now() - this.reconnectStart;
```
davidzhao, line 493: this logic is now repeated from the top. can you think of a way to avoid this duplication?

src/room/RTCEngine.ts
```diff
@@ -448,60 +473,57 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-          // cannot resume
-          reconnectRequired = true;
-        }
-
-        // when we flip from resume to reconnect, we need to reset reconnectAttempts
-        // this is needed to fire the right reconnecting events
-        if (reconnectRequired && !this.fullReconnectOnNext) {
-          this.fullReconnectOnNext = true;
-          this.reconnectAttempts = 0;
+        let requireSignalEvents = false;
```
davidzhao, line 485: I don't know if this logic is correct. we should only emit a reconnecting/resuming event on the very first attempt. I think this alters the behavior quite a bit.

src/room/Room.ts
```diff
@@ -139,9 +139,8 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       return;
     }
 
-    this.engine = new RTCEngine();
+    this.engine = new RTCEngine(this.options);
 
-    this.engine.client.signalLatency = this.options.expSignalLatency;
```
davidzhao, line 144: is moving this required by this PR? It would be good to keep PRs focused.

davidzhao (commented):
Thank you for the contribution. I have a few suggestions and questions about parts of it.

src/room/RTCEngine.ts
```diff
@@ -416,17 +421,37 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
     }
+
     log.debug(`${connection} disconnected`);
     if (this.reconnectAttempts === 0) {
       // only reset start time on the first try
       this.reconnectStart = Date.now();
     }
 
-    const delay = this.reconnectAttempts * this.reconnectAttempts * 300;
+    const disconnect = (duration: number) => {
```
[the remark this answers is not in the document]
taskbit: Code de-duplication, there are 2 paths requiring to emit the Disconnect event.

src/room/RTCEngine.ts
taskbit: I had a wrong thought about the UnexpectedConnectionState error, reverted the change.

src/room/RTCEngine.ts
taskbit: Previously code set `reconnectAttempts `to 0 to force re-emiting the events if `fullReconnectOnNext `is false and `reconnectRequired `is true.
I was wondering about that as well as this can emit the events twice. First & second reconnect attempt.

I decided to keep that behavior as I'm not aware of the original reason.
But instead of setting `reconnectAttempts `to 0 after the first attempt I found it cleaner to signal the events in a different way because it's the 2nd reconnect attempt we should not reset the counter as this alters the logic in the reconnect policy by re-evaluating `retryCount `with "0" another time.

I don't see an alteration of the previous behavior regarding events emitted.

src/room/Room.ts
taskbit: This is related to the PR but not a functional requirement. I passed the `RoomOptions `to the `RTCEngine `constructor to get hold of the `reconnectPolicy`. 

As `RTCEngine `is now aware of the `RoomOptions `it makes more sense in my opinion to apply the configuration setting of the `SignalClient `where it's created.

davidzhao:
thanks for your clarifications, I'll give it a try tomorrow and will keep you posted!

src/room/IReconnectPolicy.ts
```diff
@@ -0,0 +1,25 @@
+/** Controls reconnecting of the client */
+export interface IReconnectPolicy {
```
davidzhao, line 2: naming nit: we don't prefix interfaces with I, keeping the same pattern.

```suggestion
export interface ReconnectPolicy {
```

src/room/IReconnectPolicy.ts
```diff
@@ -0,0 +1,25 @@
...
+  /** Called after disconnect was detected
+   *
+   * @returns {number | null} Amount of time in milliseconds to delay the next reconnect attempt, `null` signals to stop retrying.
+   */
+  nextRetryDelayInMs(context: IReconnectContext): number | null;
+}
+
+export interface IReconnectContext {
```
davidzhao, line 10: naming nit:

```suggestion
export interface ReconnectContext {
```

src/room/RTCEngine.ts
davidzhao: I double checked the listeners and can confirm that we want to fire the signal again going from resume to reconnect. So your implementation is right.

src/room/RTCEngine.ts
```diff
@@ -540,25 +569,44 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
 
+    this.emit(EngineEvent.SignalResumed);
     this.subscriber.restartingIce = true;
 
+    let waitForPcCompleted = false;
+    const waitForPc = this.waitForPCConnected()
+      .then(() => {
+        waitForPcCompleted = true;
+      })
+      .catch(() => Error('failed to wait for peer connection'));
```
davidzhao, line 590: is this intending to throw? I'm unfamiliar with this syntax.

src/room/RTCEngine.ts
```diff
@@ -540,25 +569,44 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+
+    while (!waitForPcCompleted) {
+      // TODO: this is a workaround to check if the socket was closed while we are reconnecting
```
davidzhao, line 593: can you describe a repro step to trigger a socket closure while reconnecting?

src/room/RTCEngine.ts
```diff
@@ -540,25 +569,44 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+      // if signal client reconnects with ?reconnect=1 ws.onclose is almost immediately called after ws.onopen
+      if (!this.client.isConnected) {
+        this.fullReconnectOnNext = true;
```
davidzhao, line 596: does it need to be set here again? throwing should be sufficient to move this into the right state.

davidzhao (commented):
great progress here, added a few notes.

src/room/RTCEngine.ts
taskbit: Thanks good catch, there is throw missing.

src/room/RTCEngine.ts
taskbit: I decided to revert as it's not related to the changes in the PR, I would classify it as a separate bug.

src/room/RTCEngine.ts
taskbit: You are correct no need to set here again.

davidzhao (approved):
fantastic work here! this is a meaty PR and really improves the flexibility of reconnection. Thank you for making LiveKit better!

taskbit:
@davidzhao thank you for working on this awesome project and the detailed code review, appreciate it!

PR #267, opened 2022-06-16
publish new codec when subscriber need

cnderrauber:
client change for https://github.com/livekit/livekit/pull/762

### DIFF
--- a/.changeset/loud-buses-swim.md
+++ b/.changeset/loud-buses-swim.md
@@ -2,4 +2,4 @@
 'livekit-client': patch
 ---
 
-publish new codec when subscriber need
+Start publishing backwards-compatible codec only when subscribers need it
--- a/.changeset/loud-buses-swim.md
+++ b/.changeset/loud-buses-swim.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+publish new codec when subscriber need
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -108,14 +108,6 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     this.identityToSid = new Map();
     this.options = options || {};
 
-    switch (this.options?.publishDefaults?.videoCodec) {
-      case 'av1':
-      case 'vp9':
-        this.options.publishDefaults.simulcast = undefined;
-        break;
-      default:
-    }
-
     this.options.audioCaptureDefaults = {
       ...audioDefaults,
       ...options?.audioCaptureDefaults,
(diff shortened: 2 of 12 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -734,7 +795,14 @@ export default class LocalParticipant extends Participant {
       return;
     }
     if (update.subscribedCodecs.length > 0) {
-      pub.videoTrack?.setPublishingCodecs(update.subscribedCodecs);
+      if (!pub.videoTrack) {
+        return;
+      }
+      const newCodecs = await pub.videoTrack.setPublishingCodecs(update.subscribedCodecs);
```
davidzhao, line 801: how does the subscriber subscribe to a codec that isn't published yet?

cnderrauber: The publisher will declare some `simulcast codecs` but not publish them at beginning, and sfu will create a`DumpReceiver` for each  codecs used for negotiate with subscribers( Like sfu.WebrtcReceiver but no track bind to it), when it is subscribed by a client, publisher will publish a track and `DumbReceiver` upgrade to WebrtcReciever

cnderrauber: Now we don't have api for user to set `backup codecs`, but add vp8 as backup by default when user select svc(av1) codec, do we need a method for user to set them manually?

src/room/participant/LocalParticipant.ts
```diff
@@ -465,37 +461,38 @@ export default class LocalParticipant extends Participant {
...
+          simEncodings = computeVideoEncodings(
+            track.source === Track.Source.ScreenShare,
+            width,
+            height,
+            simOpts,
+          );
+        }
+
+        // set vp8 codec as backup for any other codecs
+        if (opts.videoCodec && opts.videoCodec !== 'vp8') {
```
cnderrauber, line 482: @davidzhao set vp8 as backup for other codecs

.changeset/loud-buses-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+publish new codec when subscriber need
```
davidzhao, line 5:
```suggestion
Start publishing backwards-compatible codec only when subscribers need it
```

src/room/Room.ts
```diff
@@ -108,14 +108,6 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     this.identityToSid = new Map();
     this.options = options || {};
 
-    switch (this.options?.publishDefaults?.videoCodec) {
```
davidzhao, line 111: discussed with Jie and decided to not surface this option to users right now.

davidzhao (approved):

PR #274, opened 2022-06-21
Added generic type to the Room class so RoomEventCallbacks can be extended

prattcmp:
This change is for developers who want to extend Room and RoomEventCallbacks. RoomEventCallbacks cannot be extended with additional callbacks because the Room class is not a generic. 

It is not a breaking change.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -55,7 +55,7 @@ export const RoomState = ConnectionState;
  *
  * @noInheritDoc
  */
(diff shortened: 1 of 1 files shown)


lukasIO:
hi @prattcmp,
thanks for the PR! 
Generally open to let developers extend the RoomEventCallbacks, but it TS is complaining about your changes with
 `Type 'Tsub' does not satisfy the constraint 'EventMap'.`. Are you seeing this, too?

lukasIO:
closing due to inactivity

PR #275, opened 2022-06-22
Fix missed TrackPublished events.

davidzhao:
When initial participant info is sent down along with tracks, we would
miss emitting TrackPublished events for them. This is due an incorrect
assumption that participants would first join without tracks.

Instead of relying on if it's the first update, we now use Room connection
state to determine if those TrackPublished events should be emitted or not.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -500,6 +500,20 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     stream: MediaStream,
     receiver?: RTCRtpReceiver,
   ) {
+    // don't fire onSubscribed when connecting
+    // WebRTC fires onTrack as soon as setRemoteDescription is called on the offer
+    // at that time, ICE connectivity has not been established so the track is not
+    // technically subscribed.
+    // We'll defer these events until when the room is connected or eventually disconnected.
+    if (this.state === ConnectionState.Connecting || this.state === ConnectionState.Reconnecting) {
+      setTimeout(() => {
+        this.onTrackAdded(mediaTrack, stream, receiver);
+      }, 10);
+      return;
+    }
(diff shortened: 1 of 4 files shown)


davidzhao:
found a related bug that I'm fixing together. protocol 8 causes us to fire TrackSubscribed prior to Room connection state going to connected.

lukasIO (approved):

davidliu:
One small thing is that other participant events might fire from updating the info. When implementing for android, participant metadata event fired for me.

davidzhao:
> One small thing is that other participant events might fire from updating the info. When implementing for android, participant metadata event fired for me.

Yeah, I've noticed that too. will fix that too before merging.

src/room/Room.ts
```diff
@@ -500,6 +500,20 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     stream: MediaStream,
     receiver?: RTCRtpReceiver,
   ) {
+    // don't fire onSubscribed when connecting
+    // WebRTC fires onTrack as soon as setRemoteDescription is called on the offer
+    // at that time, ICE connectivity has not been established so the track is not
+    // technically subscribed.
+    // We'll defer these events until when the room is connected or eventually disconnected.
+    if (this.state === ConnectionState.Connecting || this.state === ConnectionState.Reconnecting) {
+      setTimeout(() => {
```
cloudwebrtc, line 509: @davidzhao If it is to ensure that the track can be rendered (can receive RTP) when onTrackAdded is triggered, would it be more accurate to listen iceConnectionState instead of delaying 10 seconds?

davidzhao: right, more precisely, we can wait for `ConnectionStateChanged` event, and trigger the update when it flips to Connected.

PR #279, opened 2022-06-23
Only restart tracks if they are internally managed by the SDK

lukasIO:
Not very happy about the naming with `managed` and `isTrackManaged` because it's quite ambiguous. Could mean both "externally managed" and "internally managed".

### DIFF
--- a/.changeset/shiny-eyes-clean.md
+++ b/.changeset/shiny-eyes-clean.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Only restart tracks if they are internally managed by the sdk
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -578,7 +578,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
         if (!track.isMuted) {
           if (
             (track instanceof LocalAudioTrack || track instanceof LocalVideoTrack) &&
-            track.trackIsManaged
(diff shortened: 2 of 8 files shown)


davidzhao (approved):

lukasIO:
@davidzhao one thing where I could argue for both cases is after a call to `track.replaceTrack`. 
Right now calling this function does not mark the track as `userProvided` but maybe it should. 

Reason to argue it should: User is in fact providing a raw MediaStreamTrack. We have no control over it directly. So we should assume it's managed externally.

But: If a user has our APIs previously for creating tracks then it might be that they expect us to manage them also after they called replaceTrack. So maybe the right option is to not change the `userProvided`  state and just default to however the original LocalTrack has been created.

davidzhao:
> @davidzhao one thing where I could argue for both cases is after a call to `track.replaceTrack`. Right now calling this function does not mark the track as `userProvided` but maybe it should.
> 
> Reason to argue it should: User is in fact providing a raw MediaStreamTrack. We have no control over it directly. So we should assume it's managed externally.
> 
> But: If a user has our APIs previously for creating tracks then it might be that they expect us to manage them also after they called replaceTrack. So maybe the right option is to not change the `userProvided` state and just default to however the original LocalTrack has been created.

Ah good point.. makes sense if *they* have replaced the Track with something else, then we shouldn't try to reacquire it. I think for users that are doing things that are a bit more advanced, we can let them write a bit more code to handle these corner cases :)

PR #282, opened 2022-06-24
Improve duplicate connect

davidzhao:
When a user calls connect multiple times, the behavior is now consistent
where it returns a promise so caller wait to see if connection is
successful or otherwise.


lukasIO (approved):
lgtm!

PR #286, opened 2022-06-27
Use RoomEvent for emitting event on the room

lukasIO:
Should not have had any effect as the event strings are the same, but seems cleaner this way.


davidzhao (approved):

davidzhao:
hah I didn't see this PR and made the same change in mine

PR #292, opened 2022-06-29
Use boolean value for publish only option

lukasIO:
(would be a minor bump)


hiroshihorie:
If I remember correctly , this was supposed to be a string that gets appended to the identity like "identity#publish" so it should remain a string ?
@davidzhao or @cnderrauber Can you please confirm ? 🙂

PR #293, opened 2022-06-29
Respect stopLocalTrackOnUnpublish in event of engine disconnect

lukasIO:
The option name does not exactly say it would do the same for disconnect events.
Any suggestion on handling it differently? Maybe a separate option for `stopLocalTrackOnDisconnect`?


davidzhao (approved):
lg!

PR #300, opened 2022-07-01
Queue and await parallel calls to mute/unmute for local tracks

lukasIO:
Introduces an additional dependency on 'async-await-queue'. 
Keeping this PR clean, but we might be able to reuse the same lib for the signal request queue in order to keep the overall bundle size down as much as possible.

### DIFF
--- a/.changeset/thick-ways-turn.md
+++ b/.changeset/thick-ways-turn.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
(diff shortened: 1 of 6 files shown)


davidzhao (approved):
lgtm, this should be compatible with react native right?

lukasIO:
I don't have any reason to believe otherwise.
 @davidliu do you see any problem regarding this change with react native?

PR #301, opened 2022-07-01
Re-use external queue lib for signalling queue

lukasIO:
Instead of having two different queue implementations, this PR gets rid of ours and replaces it with the one used in #300


davidzhao (approved):
lgtm!

PR #304, opened 2022-07-02
Forward leave reason of disconnected events

lukasIO:
client implementation for https://github.com/livekit/protocol/pull/139

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -15,6 +15,7 @@ import {
 import {
   ConnectionQualityUpdate,
   JoinResponse,
+  LeaveRequest_LeaveReason,
   SimulateScenario,
   StreamStateUpdate,
   SubscriptionPermissionUpdate,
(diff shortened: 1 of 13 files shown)


src/room/Room.ts
```diff
@@ -15,6 +15,7 @@ import {
 import {
   ConnectionQualityUpdate,
   JoinResponse,
+  LeaveRequest_LeaveReason,
```
davidzhao, line 18: what do you think about exporting a type-defed version of this? i.e. `type DisconnectedReason = LeaveRequest_LeaveReason`, it seems a bit more friendly for users to import.

lukasIO: sounds good! was missing the export anyways...

davidzhao: or.. maybe I should just name that DisconnectReason in the proto? 🤔 wdyt?

lukasIO: renaming in the proto seems slightly cleaner, but no strong opinion either way.

davidzhao: yeah, you are right.. it'll be consistent across platforms.. I'll do that and then merge.

davidzhao: yeah, you are right.. it'll be consistent across platforms. will update.

davidzhao (approved):
lg!

PR #306, opened 2022-07-03
Presence of Track subscription to take priority over allowed status.

davidzhao:
Client may end up with a copy of incorrect RemoteTrackPublication.allowed
status, if rapid successions of subscribe/unsubscribe/permission updates
occur.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -375,7 +375,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       this.engine.close();
     }
 
-    this.handleDisconnect(stopTracks);
+    this.handleDisconnect(stopTracks, DisconnectReason.CLIENT_INITIATED);
     /* @ts-ignore */
     this.engine = undefined;
   };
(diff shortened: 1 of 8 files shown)


boks1971 (approved):

lukasIO (approved):

src/room/Room.ts
```diff
@@ -358,7 +358,8 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
   /**
    * disconnects the room, emits [[RoomEvent.Disconnected]]
    */
-  disconnect = (stopTracks = true) => {
+  disconnect = async (stopTracks = true) => {
```
lukasIO, line 361: we should probably mention the fact that disconnect is async now in the changeset

davidzhao: so I made this change because connection basics started failing. it looks like the request queue is causing a change in timing so without async/await, it would proceed to close the engine before attempting to send `leave` to the server.

lukasIO: makes total sense, I missed this in the other PR. not saying we'd have to advertise as a behavioural change but just as a heads up maybe?

PR #308, opened 2022-07-04
Try to re-aquire mediastreamtrack when device has been disconnected, …

lukasIO:
…pause upstream if no device could be acquired

The function is currently deeply nested. We can figure out the best course of action first and then I'll clean it up.

The basic idea to react to the 'ended' event is:
- Unpublish if it's a screen track (-> stop initiated by stopping via browser UI)
- Don't do anything if the camera/microphone track is user provided
- Try to restart the track if it has ended (-> will effectively re-acquire with the given constraints and might switch to a different device)
- If reacquiring fails, replace the current mediastreamtrack with an empty one and set it to muted on the server side (-> calling `pauseUpstream`)
- If the track ended because media permissions were revoked, attach an eventlistener to try to restart the track once permissions have been granted again (currently not supported in firefox and safari)

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -32,7 +32,7 @@ import {
 } from '../track/options';
 import { Track } from '../track/Track';
 import { constraintsForOptions, mergeDefaultOptions } from '../track/utils';
-import { isFireFox } from '../utils';
+import { isFireFox, isWeb } from '../utils';
 import Participant from './Participant';
 import { ParticipantTrackPermission, trackPermissionToProto } from './ParticipantTrackPermission';
 import { computeVideoEncodings, mediaTrackToLocalTrack } from './publishUtils';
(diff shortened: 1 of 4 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -894,11 +894,45 @@ export default class LocalParticipant extends Participant {
...
+              }
+            };
+            throw new Error('GetUserMedia Permission denied');
+          }
+        }
+        log.debug('track ended, attempting to use a different device');
+        await track.restartTrack();
+      } catch (e) {
+        log.warn(`could not restart track, pausing upstream instead`);
+        await track.pauseUpstream();
```
davidzhao, line 933: should pauseUpstream replace the MediaStreamTrack that's attached to `Track` as well? I think that would ensure we are not listening to events from the old MediaStreamTrack.

src/room/participant/LocalParticipant.ts
```diff
@@ -894,11 +894,45 @@ export default class LocalParticipant extends Participant {
...
+  private handleTrackEnded = async (track: LocalTrack) => {
+    if (
+      track.source === Track.Source.ScreenShare ||
+      track.source === Track.Source.ScreenShareAudio
+    ) {
+      log.debug('unpublishing local track due to TrackEnded', {
+        track: track.sid,
+      });
+      this.unpublishTrack(track);
+    } else if (track.isUserProvided) {
```
davidzhao, line 906: should we pauseUpstream here too? if a user track is ended, I think we wouldn't know how to send data associated with it. without indicating it's muted somehow, I think we might get a bunch of PLIs from the receiver.

src/room/participant/LocalParticipant.ts
lukasIO: good question. we could do that. I assumed that `ended` is the last event we ever see of a particular mediastreamtrack. The spec doesn't make it overly clear, but it sounds to me that way:
> The [MediaStreamTrack](https://w3c.github.io/mediacapture-main/#dom-mediastreamtrack) object's source will no longer provide any data, either because the user revoked the permissions, or because the source device has been ejected, or because the remote peer permanently stopped sending data.

lukasIO: I think `pauseUpstream` might be misleading if it also replaces the original track, as we will loose the ability to `resumeUpstream`, maybe it makes sense to add a `stopUpstream` function which mutes on server side plus sets the actual track to undefined?

davidzhao: right, that's a fair point. you are right it shouldn't do that. handling resume on replaceTrack covers that case I think.

PR #311, opened 2022-07-07
Avoid multiple calls to getUserMedia for getLocalDevices

lukasIO:
fixes iOS safari video preview disappearing. 
Alternative fix on the SDK level to https://github.com/livekit/livekit-react/pull/60.
With this fix, we don't have to change anything in the react packages.


src/room/track/create.ts
```diff
@@ -30,7 +31,15 @@ export async function createLocalTracks(
 
   const opts = mergeDefaultOptions(options, audioDefaults, videoDefaults);
   const constraints = constraintsForOptions(opts);
-  const stream = await navigator.mediaDevices.getUserMedia(constraints);
```
davidzhao, line 33: this seems like it'd be problematic when we require different constraints? It seems that each attempt could be different.

davidzhao (commented):
I wonder if there's an alternative way to trigger the permission dialog without impacting normal devices.

Is it worth trying this approach:
* in getDevices, calling navigator.mediaDevices.getUserMedia() but with constraints that wouldn't match any device.. for example, deviceId: "dummy".

Thinking out loud, the problem only exists if we acquire the same device twice, so if we could make it so that the getUserMedia attempt in discovery grants permissions, but fails to acquire an actual device.

lukasIO:
nice idea! will test that approach with a dummy deviceId.

src/room/track/create.ts
davidzhao: ok, I totally misread the code.. it wasn't re-using it, but just caching the previous one.. clever!

this should work as well.

davidzhao: though I think there are still some edge cases as it relates to audio/video types.

lukasIO:
@davidzhao your suggestion works like a charm also on iOS. It's a great workaround!

davidzhao (approved):
nice!!

lukasIO:
@davidzhao the map approach seems to work across browsers as long as the safari case is handled separately. tested across firefox/chrome/safari and iOS safari

davidzhao (approved):
awesome work!

PR #319, opened 2022-07-12
Throw error on unexpected connection state before signalling add track

lukasIO:
Prior to this fix it would be possible for the `engine.addTrack` request to be queued even though the publisher isn't defined yet.


davidzhao (approved):
lg. though the server would only leave tracks like these in a pending state, and would not actually consider them as published.

lukasIO:
closing as unneeded due to comment above.

PR #321, opened 2022-07-13
Improve reconnect timeout handling

davidzhao:
Previously multiple attempts could be queued - resulting in more frequent
retries than desired.

### DIFF
--- a/.changeset/late-mice-explode.md
+++ b/.changeset/late-mice-explode.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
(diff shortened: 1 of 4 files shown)


hiroshihorie (approved):

lukasIO (approved):
we could also change 
https://github.com/livekit/client-sdk-js/blob/bba2ab39b4cf7e3ea23a920181e9403417d8f00a/src/room/DefaultReconnectPolicy.ts#L26
to 

```ts 
if (context.retryCount >= this._retryDelays.length) return null;
```
just to be safe

PR #323, opened 2022-07-13
Remove event listeners from local participant on disconnect

### LINKED ISSUE #320: subsequent calls to room.connect cause event listeners to accumulate
https://github.com/livekit/client-sdk-js/blob/6f2d7fa1880b96bd95a99b19f5194330d760bd01/src/room/Room.ts#L257-L288

This happens when a user switches rooms (room.disconnect -> room.connect)

image

lukasIO:
fixes #320

### DIFF
--- a/.changeset/seven-suits-chew.md
+++ b/.changeset/seven-suits-chew.md
@@ -2,4 +2,4 @@
 'livekit-client': patch
 ---
 
-Reset local participant on disconnect
+Remove internal event listeners from local participant on disconnect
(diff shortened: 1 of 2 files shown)


davidzhao (commented):
I think resetting the local participant would remove any event listeners that the user has set on it.. are we able to only clean up the handlers that we've set?

lukasIO:
true, resetting would be unexpected. changed it to unsubscribing from the specific events!

davidzhao (approved):

PR #327, opened 2022-07-14
set av1 bitrate to 70% preset

cnderrauber:


cnderrauber:
#334 contain this, close it

PR #329, opened 2022-07-15
Log warning if multiple tracks of the same source are published

lukasIO:


src/room/participant/LocalParticipant.ts
```diff
@@ -434,6 +434,23 @@ export default class LocalParticipant extends Participant {
     if (opts.source) {
       track.source = opts.source;
     }
+    const existingTrackOfSource = Array.from(this.tracks.values()).find(
+      (publishedTrack) => track instanceof LocalTrack && publishedTrack.source === track.source,
+    );
+    if (existingTrackOfSource) {
+      try {
+        // throw an Error in order to capture the stack trace
+        throw Error(`publishing a second track with the same source: ${track.source}`);
```
davidzhao, line 443: excellent!

src/room/participant/RemoteParticipant.ts
```diff
@@ -220,6 +220,25 @@ export default class RemoteParticipant extends Participant {
     // always emit events for new publications, Room will not forward them unless it's ready
     newTracks.forEach((publication) => {
       this.emit(ParticipantEvent.TrackPublished, publication);
+      const existingTrackOfSource = Array.from(this.tracks.values()).find(
+        (publishedTrack) => publishedTrack.source === publication.source,
+      );
+      if (existingTrackOfSource) {
+        try {
+          // throw an Error in order to capture the stack trace
+          throw Error(
```
davidzhao, line 229: maybe just a regular warning here? since the stack trace isn't going to be useful here.

lukasIO: thought about it too, unclear why I left it in 🤔 
will remove!

davidzhao (approved):
lgtm!

PR #332, opened 2022-07-18
Clean up simulcast codecs in unpublish

lukasIO:


src/room/participant/LocalParticipant.ts
```diff
@@ -723,6 +723,15 @@ export default class LocalParticipant extends Participant {
...
+        if (track instanceof LocalVideoTrack) {
+          for (const [, trackInfo] of track.simulcastCodecs) {
+            if (trackInfo.sender) {
+              this.engine.publisher.pc.removeTrack(trackInfo.sender);
+              trackInfo.sender = undefined;
+            }
+          }
+          track.simulcastCodecs.clear();
+        }
         this.engine.negotiate();
```
davidzhao, line 735: how should it be handled if we are able to remove the primary track, but not simulcasted ones? it seems that we should still negotiate regardless right?

perhaps moving this outside of this try/catch block, and negotiating `if (hasChanged)`

davidzhao (approved):

cnderrauber (approved):

PR #334, opened 2022-07-20
Add experimental fallback codec support

lukasIO:
This is a draft/suggestion of how to expose custom multi codec publishing encodings to users. 

Naming (polycast) of course also up for debate, but as everything is marked as `@experimental` it should also be fine to rename afterwards


cnderrauber (approved):

davidzhao:
thinking a bit about the API, and just reading the options, it doesn't seem intuitive to me what I should be doing there. Let's discuss @lukasIO; maybe there's a simple version of it that could work for the majority of the cases?

src/room/participant/LocalParticipant.ts
```diff
@@ -492,32 +497,31 @@ export default class LocalParticipant extends Participant {
       req.height = height ?? 0;
       // for svc codecs, disable simulcast and use vp8 for backup codec
       if (track instanceof LocalVideoTrack) {
-        if (opts?.videoCodec === 'vp9' || opts?.videoCodec === 'av1') {
+        if (opts.backupCodec && (opts?.videoCodec === 'vp9' || opts?.videoCodec === 'av1')) {
           // set scalabilityMode to 'L3T3' by default
           opts.scalabilityMode = opts.scalabilityMode ?? 'L3T3';
```
davidzhao, line 502: I think we need to set scalability mode as long as it's AV1, even if a backupCodec isn't set.

We should also remove VP9 references, since we've decided to focus on AV1 and not worry about VP9

src/room/participant/LocalParticipant.ts
```diff
@@ -598,16 +602,9 @@ export default class LocalParticipant extends Participant {
...
+    videoCodec: BackupVideoCodec,
     options?: TrackPublishOptions,
   ) {
-    const opts: TrackPublishOptions = {
-      ...this.roomOptions?.publishDefaults,
-      ...options,
-    };
-    // clear scalabilityMode setting for backup codec
-    opts.scalabilityMode = undefined;
-    opts.videoCodec = videoCodec;
```
davidzhao, line 610: does this get set elsewhere? it gets used [here](https://github.com/livekit/client-sdk-js/pull/334/files#diff-cf84874acd5677364a55762127f033eb5a9e7ca344a666b6305c3d76452bff02R647)

src/room/participant/publishUtils.ts
```diff
@@ -181,6 +243,20 @@ export function determineAppropriateEncoding(
...
+  // presets are based on the assumption of vp8 as a codec
+  // for other codecs we adjust the maxBitrate if no specific videoEncoding has been provided
+  // TODO make the bitrate multipliers configurable per codec
+  if (codec) {
+    switch (codec) {
+      case 'av1':
+        encoding.maxBitrate = encoding.maxBitrate * 0.7;
+        break;
+      case 'h264':
+        encoding.maxBitrate = encoding.maxBitrate * 1.1;
```
davidzhao, line 255: aren't H.264 and VP8 at similar efficiencies?

davidzhao (commented):
the interface looks much easier to follow!

src/room/participant/publishUtils.ts
lukasIO: yeah, we can ignore it by default. I think the efficiency ultimately comes down to which h.264 profile is used

src/room/participant/LocalParticipant.ts
lukasIO: the link doesn't work, but generally get's set in `computeTrackBackupEncodings` https://github.com/livekit/client-sdk-js/blob/becc9ea417ba0de73541a95e47944ed05cbbf3eb/src/room/participant/publishUtils.ts#L209
Does that resolve the issue you saw?

davidzhao: ah ha I see it. Just didn't expect that `computeXXX` would mutate one of its arguments.

davidzhao (approved):

src/room/participant/LocalParticipant.ts
lukasIO: that's fair. I moved it there in order to keep the arguments untouched wherever possible and then only have a distinct function that needs the mutation to actually do it.

src/room/participant/LocalParticipant.ts
```diff
@@ -671,18 +664,19 @@ export default class LocalParticipant extends Participant {
...
-      this.engine.publisher.setTrackCodecBitrate(
-        req.cid,
-        videoCodec,
-        encodings[0].maxBitrate / 1000,
-      );
-    }
+    this.setPreferredCodec(transceiver, track.kind, videoCodec);
+    track.setSimulcastTrackSender(videoCodec, transceiver.sender);
+
+    // TODO do we want to support av1 as a backup codec?
```
cnderrauber, line 670: I think we will not use av1 as backup, it should be primary codec if exist

cnderrauber (approved):
looks good!

Issue #356, opened 2022-07-24
Room.setAndEmitConnectionState rejects a promise with Error: "disconnected from Room"  that cannot be handled


aravindanve:
## Summary

The method `Room.setAndEmitConnectionState` rejects a promise with Error: "disconnected from Room" that cannot be handled, at least in the following scenarios, and possibly more:

1. Reconnection fails and max attempts are exhausted
2. When the room is in reconnecting state and I call disconnect

https://github.com/livekit/client-sdk-js/blob/9c999845a34984a24dc4b3b27a0c09746e2ba3c8/src/room/Room.ts#L1052-L1056

<img width="1676" alt="Screenshot 2022-07-24 at 11 52 03 AM" src="https://user-images.githubusercontent.com/5883448/180635362-879995db-7131-4453-a6d3-730c9dcb3e40.png">

#### 1. Reconnection failed and max attempts exhausted

It would be nice to be able to handle this gracefully. For now I'm using a Custom reconnect policy:

```js
import { DefaultReconnectPolicy } from "livekit-client";

export class CustomReconnectPolicy extends DefaultReconnectPolicy {
  constructor({ handleGiveUp }) {
    this.handleGiveUp = handleGiveUp;
  }

  nextRetryDelayInMs(context) {
    const result = super.nextRetryDelayInMs(context);
    if (result !== null) {
      return result;
    } else {
      return this.handleGiveUp(context);
    }
  }
}

```

But that results in the following problem

#### 2. When the room is in reconnecting state and I call disconnect

I handle reconnection failure by directing the user away from the conference page in my `CustomReconnectPolicy.handleGiveUp` callback. But during clean up, when I call `room.disconnect().catch(console.error)`, `Room.setAndEmitConnectionState` rejects the `connectFuture` that results in an unhandled rejection.

https://github.com/livekit/client-sdk-js/blob/9c999845a34984a24dc4b3b27a0c09746e2ba3c8/src/room/Room.ts#L1052-L1056

## Solutions

Do not reject a promise that is not returned or is inaccessible via the public API.

#### 1. Reconnection failed and max attempts exhausted

Additionally, emit the reason for disconnect here https://github.com/livekit/client-sdk-js/blob/9c999845a34984a24dc4b3b27a0c09746e2ba3c8/src/room/RTCEngine.ts#L469

#### 2. When the room is in reconnecting state and I call disconnect

[...]

PR #359, opened 2022-07-25
Only reject connection promise if it was triggered by a call to connect()

lukasIO:
addresses (and fixes partly) #356

### DIFF
--- a/.changeset/breezy-chefs-know.md
+++ b/.changeset/breezy-chefs-know.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Only reject connection promise if it was triggered by a call to connect()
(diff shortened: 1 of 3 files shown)


davidzhao (approved):

davidzhao (commented):
lg structurally. One scenario I don't believe it handles is:

1. reconnection attempt is in progress
2. user calls await connect()
3. reconnection actually failed

in this case I think that connect attempt would be stuck. should there be a check in handleDisconnected to check if it's an ongoing reconnect?

davidzhao:
> you are right! Rather than checking whether it's an ongoing reconnect in `handleDisconnected`, what do you think about simply rejecting pending connect promises inside of `handleDisconnected`? Seems like that would be cleaner than my current solution.

I think rejecting `reconnectPromise` would work great, since if we do reject connectPromise, it'll lose the ability to return specific error types in `connect`.

davidzhao (approved):
looks good!

PR #363, opened 2022-07-28
Clear pending track resolver on track unpublish

lukasIO:
- decreases publication timeout to 10s
- new `engine.removeTrack` method that clears `pendingTrackResolver` track entry (and rejects publication promise if `unpublish` is called while track is not fully published (no response from server)).

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -72,7 +72,9 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private _isClosed: boolean = true;
 
-  private pendingTrackResolvers: { [key: string]: (info: TrackInfo) => void } = {};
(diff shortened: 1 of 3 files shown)


src/room/RTCEngine.ts
```diff
@@ -166,18 +168,40 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+        reject: () => {
+          clearTimeout(publicationTimeout);
+          reject('Cancelled publication by calling unpublish');
+        },
       };
       this.client.sendAddTrack(req);
     });
   }
 
+  removeTrack(sender: RTCRtpSender) {
```
davidzhao, line 190: nice!

davidzhao (approved):
awesome work!

PR #376, opened 2022-08-04
Fix setting name on RemoteParticipant creation

lukasIO:


davidzhao (approved):
lg. what about also setting name in `updateInfo` ?

lukasIO:
@davidzhao `updateInfo` does this already!
https://github.com/livekit/client-sdk-js/blob/main/src/room/participant/Participant.ts#L176

PR #377, opened 2022-08-04
Add ping pong heartbeat for signal connection

lukasIO:
would - theoretically - work without a protocol bump, as the ping timeout never starts if the join response doesn't include `pingInterval` and `pingTimeout`, but might still make sense to update the version.

### DIFF
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -542,7 +542,6 @@ export class SignalClient {
   }
 
   private clearPingTimeout() {
-    log.debug('clearing ping timeout');
     if (this.pingTimeout) {
       clearTimeout(this.pingTimeout);
     }
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -3,6 +3,7 @@ import 'webrtc-adapter';
 import log from '../logger';
 import {
   ClientInfo,
+  DisconnectReason,
   ParticipantInfo,
   Room,
   SpeakerInfo,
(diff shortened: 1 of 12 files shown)


src/api/SignalClient.ts
```diff
@@ -463,6 +492,37 @@ export class SignalClient {
   private handleWSError(ev: Event) {
     log.error('websocket error', ev);
   }
+
+  private resetPingTimeout() {
+    if (this.pingTimeoutDuration) {
```
boks1971, line 497: Guess this should be `pingTimeout`?

src/api/SignalClient.ts
```diff
@@ -152,6 +160,10 @@ export class SignalClient {
 
   async reconnect(url: string, token: string): Promise<void> {
     this.isReconnecting = true;
+    // clear ping timeout and reset it once reconnected
+    if (this.pingTimeoutDuration) {
```
boks1971, line 164: Guess this should be `pingTimeout`? Also, can this call `resetPingTimer()`?

src/api/SignalClient.ts
```diff
@@ -265,6 +282,10 @@ export class SignalClient {
     if (this.ws) this.ws.onclose = null;
     this.ws?.close();
     this.ws = undefined;
+    if (this.pingTimeout) {
```
boks1971, line 285: Maybe make an equivalent `stopPingTimeout` same as `stopPingInterval()` call it from everywhere rather than the if + clear in all the places?

davidzhao (approved):
lgtm!

PR #378, opened 2022-08-04
[ADD] support for boolean and exact values in facing mode

cscherban:


alist:
@cscherban could you explain what cases the `ConstrainDOMStringParameters` covers?

PR #382, opened 2022-08-05
Disable simulcast if an user provides an empty array for custom simulcast layers

lukasIO:
makes it possible to disable screen share simulcast by setting `screenShareSimulcastLayers` to `[]` while simulcast is still being used for camera publications (and vice versa)


boks1971 (approved):

davidzhao (approved):
lgtm!

PR #388, opened 2022-08-09
don't declare simulcastCodecs for firefox

cnderrauber:
fix issue that subscriber can't receive video stream when publisher prefer h264 codec in firefox.
we declare `h264`, `vp8` simulcast codecs when prefer h264 in sample app with **Firefox** (always use vp8 as backup codec), and subscriber match h264 first, while firefox can’t use `setCodecPreferences` (it not support that api)  to send a h264 stream. fixed by disable simulcast codecs feature for firefox.

### DIFF
--- a/.changeset/bright-waves-jump.md
+++ b/.changeset/bright-waves-jump.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+don't declare simulcast codecs for firefox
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
lgtm! 
just referencing #334 in order for us not to forget this change when merging 334.

Issue #392, opened 2022-08-11
Calling start() on a remote track multiple times creates multiple looping monitors


wcarle:
When calling `start()` on RemoteTrack it calls `this.monitorReceiver()` which then calls itself with:
```
    setTimeout(() => {
      this.monitorReceiver();
    }, monitorFrequency);
```

This means that if you call start again on a remote track it will create a new setTimout loop. This can lead to a case where there are many redundant monitorReceiver calls on the same RemoteTrack if you call start on a track multiple times

lukasIO:
closed via #393

PR #393, opened 2022-08-11
Fixes calling start() on a remote track multiple times

wcarle:
Prevent multiple redundant monitors from being started if start is called multiple times on a RemoteTrack 

Fixes issue: https://github.com/livekit/client-sdk-js/issues/392

### DIFF
--- a/.changeset/nine-crabs-vanish.md
+++ b/.changeset/nine-crabs-vanish.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Prevent multiple redundant monitors from being started if start is called multiple times on a RemoteTrack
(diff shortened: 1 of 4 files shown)


lukasIO:
hi @wcarle, 
thanks for the PR!
Looking at the code (not particularly your changes), I'm thinking it might make sense to base the monitor on an interval, rather than timeouts. 
Then a reference to the interval could be used to track if monitoring has been started already and it wouldn't need the extra variable to track if monitoring has been started.

wcarle:
@lukasIO Yeah that's a good call, it's much cleaner that way. Just pushed that change

lukasIO:
great, thank you! 
I think there's only one thing left to do, which is clearing the interval at the top of the `stop` function on the `RemoteTrack`

wcarle:
@lukasIO Done!

PR #395, opened 2022-08-15
Populate metadata when creating participant

lukasIO:
Fixes a bug that got introduced by #365 where the `ParticipantConnected` event could have been missing the `participantMetadata`.


davidzhao (approved):
lg

PR #396, opened 2022-08-16
Send network parameter with current network type

davidzhao:
This is only supported on Chrome-like browsers at the moment.


lukasIO:
by setting it via connect parameters we lose the ability to react to network changes. 
Could imagine it would be useful to be able to have visibility of the changes while connected as well?

lukasIO (approved):

PR #401, opened 2022-08-18
Move docs publishing into release action

lukasIO:


.github/workflows/release.yaml
```diff
@@ -32,8 +32,16 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      
 
-      # - name: Send a Slack notification if a publish happens
-      #   if: steps.changesets.outputs.published == 'true'
-      #   # You can do something when a publish happens.
-      #   run: my-slack-bot send-notification --message "A new version of ${GITHUB_REPOSITORY} was published!"
\ No newline at end of file
+      - name: Build Docs
```
davidzhao, line 37: would this run each time instead of only at publish time? Is that the intention?

If intention is only to publish docs upon released, I think we'd need to perform this step under `publish:` of `changesets/action@v1`

lukasIO: because of the line underneath
`if: steps.changesets.outputs.published == 'true'` 
the step only runs if the changesets job actually published something. (this is using the same pattern as their official example: https://github.com/changesets/action#with-publishing

davidzhao: I didn't realize that. fantastic!

davidzhao (approved):

PR #403, opened 2022-08-18
Add fallback for `addTrack` if `addTransceiver` is not supported by a device

lukasIO:
the work in this PR is largely based upon the work in #373 (thank you @cscherban !)
main difference implementation wise is that there's no support for `addStream`. 
apart from that it's mostly refactoring.

To increase compatibility further, we will have to add some more specific browser targets to the babel transpilation (will open a separate PR for that).

note: using `addTrack` currently breaks dynacast as this check is failing https://github.com/livekit/client-sdk-js/blob/main/src/room/track/LocalVideoTrack.ts#L322

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -538,7 +538,26 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     throw new UnexpectedConnectionState('Required webRTC APIs not supported on this device');
   }
 
-  async createTransceiverRTCRtpSender(
+  async createSimulcastSender(
+    track: LocalVideoTrack,
+    simulcastTrack: SimulcastTrackInfo,
+    opts: TrackPublishOptions,
+    encodings?: RTCRtpEncodingParameters[],
+  ) {
+    // store RTCRtpSender
+    // @ts-ignore
+    if (supportsTransceiver()) {
+      return this.createSimulcastTransceiverSender(track, simulcastTrack, opts, encodings);
+    }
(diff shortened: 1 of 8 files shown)


src/room/RTCEngine.ts
```diff
@@ -475,6 +479,142 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+  async createSender(
+    track: LocalTrack,
+    opts: TrackPublishOptions,
+    encodings?: RTCRtpEncodingParameters[],
+  ) {
+    if (supportsTransceiver()) {
+      return this.createTransceiverRTCRtpSender(track, opts, encodings);
+    }
+    if (supportsAddTrack()) {
+      console.log('using add-track fallback');
```
davidzhao, line 535: switch to our logger?

src/room/participant/LocalParticipant.ts
```diff
@@ -577,7 +567,7 @@ export default class LocalParticipant extends Participant {
     this.engine.negotiate();
 
     // store RTPSender
-    track.sender = transceiver.sender;
+    track.sender = await this.engine.createSender(track, opts, encodings);
```
davidzhao, line 570: I think this will need to happen before `this.engine.negotiate()` and `setTrackCodecBitrate`. tho today this works because of the debounce logic that we have in place.

davidzhao (commented):
This is a great cleanup!

cnderrauber (approved):

PR #404, opened 2022-08-18
Error on implicit type only import

lukasIO:
currently based on #403 as this includes a largish change. 
The option to set non-value-imports to error has the following advantage (taken from the typescript docs):
> This preserves all imports (the same as the preserve option), but will error when a value import is only used as a type. This might be useful if you want to ensure no values are being accidentally imported, but still make side-effect imports explicit.


davidzhao (approved):
looks good! would this have any impact to users of the library?

lukasIO:
there is no impact for users of the library. This change is only considered at build time by the typescript compiler and its default behaviour would have stripped implicit type-only imports at build time even before this change. 
The difference is that this change makes it explicit if we are interested in the type only or using the actual class/value.

PR #405, opened 2022-08-19
Specify more explicit browser targets

lukasIO:
...as part of the efforts to increase compatibility. 
Firefox version above 53 because below that it adds ~10kb on the gzipped bundle size. 
FF v53 was released in 2017, so I think this is an ok trade off. 

This PR is also removing the `defaults` setting of browserslist, which means some browsers like opera, baidu, kaios aren't considered at all anymore for transpilation. The reasoning behind this is, that they do not support the necessary webRTC APIs anyways to use the library.

### DIFF
--- a/.changeset/yellow-experts-joke.md
+++ b/.changeset/yellow-experts-joke.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Update browserslist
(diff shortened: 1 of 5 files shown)


davidzhao (approved):
awesome work!

PR #410, opened 2022-08-22
Move monitor APIs into parent classes

lukasIO:
follow up on https://github.com/livekit/client-sdk-js/pull/393 and unifies the approach to use an interval instead of setTimeout also for local tracks, making sure that the aren't multiple parallel monitoring loops running.


davidzhao (approved):
lgtm!

PR #412, opened 2022-08-23
Room.isRecording to surface if the room is being recorded

davidzhao:


lukasIO (approved):
lgtm! just wondering if we'd want to surface this as an event for convenience?

davidzhao:
> lgtm! just wondering if we additionally want to surface this as an event for convenience?

that's true.. let's think on that for a bit. thinking that it might be easier to think about to have a single event that indicates attributes about the room have changed.. rather than one event for each attribute.

lukasIO:
what other room attributes do you have in mind?
If there aren't too many, I think a granular event for recording could still make sense

lukasIO:
@davidzhao does the additional event look ok to you?

davidzhao (commented):
This looks good. Thanks @lukasIO

PR #414, opened 2022-08-24
add force relay configuration

cnderrauber:


src/proto/livekit_models.ts
```diff
@@ -89,9 +88,8 @@ export function trackSourceToJSON(object: TrackSource): string {
       return 'SCREEN_SHARE';
     case TrackSource.SCREEN_SHARE_AUDIO:
       return 'SCREEN_SHARE_AUDIO';
-    case TrackSource.UNRECOGNIZED:
     default:
-      return 'UNRECOGNIZED';
+      return 'UNKNOWN';
```
cnderrauber, line 92: looks I have use a different protoc version, which version are used to generate this?

lukasIO: oh, mine says `libprotoc 3.19.4`. 
can you try running `yarn install` to see if that changes the output?
I think recently some dev dependencies got updated by the renovate bot...

lukasIO (approved):

src/proto/livekit_models.ts
cnderrauber: it worked after `yarn install`, thanks!

PR #418, opened 2022-08-28
Fix issue with older browsers erroring upon setCodecPreferences

davidzhao:
On Chrome 96, using setCodecPreferences in the way that we do causes
the browser to fail on setLocalDescription, even for the offer that it
had created.

@cnderrauber With the latest version of the server, can we can fully rely on server-side setCodecPreferences? If so we could deprecate this on the client-side.

### DIFF
--- a/.changeset/rude-shrimps-run.md
+++ b/.changeset/rude-shrimps-run.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Fixed compatibility with older browsers with setCodecPreferences (Chrome 96)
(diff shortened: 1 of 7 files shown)


cnderrauber:
> On Chrome 96, using setCodecPreferences in the way that we do causes the browser to fail on setLocalDescription, even for the offer that it had created.


Failure on the offer it original created even without munged sdp? that's weird.


> @cnderrauber With the latest version of the server, can we can fully rely on server-side setCodecPreferences? If so we could deprecate this on the client-side.

 We can rely on server-side preference.

cnderrauber (approved):

davidzhao:
> Failure on the offer it original created even without munged sdp? that's weird.

Yeah it was strange. I disabled munging and it was still failing with the exact same error. The error also cannot be caught.

lukasIO (approved):

PR #419, opened 2022-08-29
fix video track lost for migration on safari

cnderrauber:
In case of session migration, safari have chance not fire the `stream.onremovetrack` event, so the RemoteTrack are not cleared, will be set again by the same remote track then cause the track detach and not attached, then some video tracks are disappeared in ui

### DIFF
--- a/src/room/track/RemoteTrackPublication.ts
+++ b/src/room/track/RemoteTrackPublication.ts
@@ -129,6 +129,9 @@ export default class RemoteTrackPublication extends TrackPublication {
   setTrack(track?: RemoteTrack) {
     const prevStatus = this.subscriptionStatus;
     const prevTrack = this.track;
+    if (!!prevTrack === !!track) {
+      return;
+    }
     if (prevTrack) {
       // unregister listener
       prevTrack.off(TrackEvent.VideoDimensionsChanged, this.handleVideoDimensionsChange);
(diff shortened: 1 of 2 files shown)


src/room/track/RemoteTrackPublication.ts
```diff
@@ -129,6 +129,9 @@ export default class RemoteTrackPublication extends TrackPublication {
   setTrack(track?: RemoteTrack) {
     const prevStatus = this.subscriptionStatus;
     const prevTrack = this.track;
+    if (!!prevTrack === !!track) {
```
lukasIO, line 132: the commit message sounds like you'd want to know if `track` and `prevTrack` are the same, but this would return always if `track` and `prevTrack` are either both set or both undefined. 
I'm wondering if there's any part where we'd expect `setTrack` to work even if a previous (different) track had been set.

cnderrauber: There is a case that old track not cleared and new track fired, then prevTrack and track are both set. Then cause prevTrack detached and new track not attached, that's why safari lost video tracks on UI. 
If `!!prevTrack === !!track` means they are either both set or both undefined, does the below code will subscribed not fired if they are both set.

```
    if (!!track !== !!prevTrack) {
      // when undefined status changes, there's a subscription changed event
      if (track) {
        this.emit(TrackEvent.Subscribed, track);
      } else if (prevTrack) {
        this.emit(TrackEvent.Unsubscribed, prevTrack);
      }
    }
```

I see @davidzhao add this code, any suggesstions about this?

PR #423, opened 2022-09-01
Add subscriptionStatusChanged event and isDesired getter for RemoteTrack Publications

lukasIO:


src/room/track/RemoteTrackPublication.ts
```diff
@@ -67,6 +67,10 @@ export default class RemoteTrackPublication extends TrackPublication {
     return super.isSubscribed;
   }
 
+  get isDesired(): boolean {
```
davidzhao, line 70: comments describing what the return value means would be helpful.

davidzhao (approved):

lukasIO:
Currently wishing there was an event to observe for a change in "desire to subscribe" to someone. Currently `updateSubscription` is being fired on track level, but it's targeted for internal usage. 
The use case scenario would be to know about when to update a participants list if selective subscriptions are being used instead of auto subscribe. 
What do you think about having an event fire on the participant that has the `track` and `isDesired` as arguments?

davidzhao:
> Thinking more about it, I wish there was an event to observe for a change in "desire to subscribe" to someone. Currently `updateSubscription` is being fired on track level, but it's targeted for internal usage. The use case scenario would be to know about when to update a participants list if selective subscriptions are being used instead of auto subscribe. What do you think about having an event fire on the participant that has the `track` and `isDesired` as arguments?

It does seem like there are quite a bit of events to pay attention to. Do you think it's something that we could fire a `subscriptionStatusChanged` ? where desired is one of the subscription statuses? That way we could have:
* unsubscribed
* desired
* not_allowed
* subscribed

lukasIO:
I think that's a great idea! 
Just trying to think this through, but I think you are right that we can get away without an extra state for `undesired` and just use `unsubscribed` in this case as well.

lukasIO:
latest changes need a bit more testing to ensure they are not breaking any previous assumptions

src/room/track/RemoteTrackPublication.ts
```diff
@@ -45,15 +51,19 @@ export default class RemoteTrackPublication extends TrackPublication {
...
 
   get subscriptionStatus(): TrackPublication.SubscriptionStatus {
-    if (this.subscribed === false || !super.isSubscribed) {
-      if (!this.allowed) {
-        return TrackPublication.SubscriptionStatus.NotAllowed;
-      }
+    if (!this.allowed) {
+      return TrackPublication.SubscriptionStatus.NotAllowed;
+    }
+    if (!this.subscribed) {
```
davidzhao, line 61: would this be better? since initial value inherits from autoSubscribe, which is true by default

```suggestion
    if (this.subscribed === false) {
```

src/room/track/RemoteTrackPublication.ts
```diff
@@ -67,6 +77,11 @@ export default class RemoteTrackPublication extends TrackPublication {
     return super.isSubscribed;
   }
 
+  // returns client's desire to subscribe to a track, also true if autoSubscribe is enabled
+  get isDesired(): boolean {
+    return !!this.subscribed;
```
davidzhao, line 82: is this equivalent to 
```suggestion
    return this.subscribed !== false;
```

`!!` is always harder for me to understand.. 😅

davidzhao (approved):

lukasIO:
@davidzhao had to update this again in order not to break previous assumptions. There's a new `PermissionStatus` enum, which we only use internally (for now) to check whether or not to emit a `TrackSubscriptionPermissionChanged` event

PR #427, opened 2022-09-07
fix video attach to incorrect element when transceiver reuse

cnderrauber:
When MediaStreamTrack been removed  from  MediaStream ,we emit event before setTrack, cause ui received `unsubscribed` event to render the participant. But the RemoteTrack holds a stale `MediaStreamTrack`, when transceiver reuse involved, the `MediaStreamTrack` might be added to other participant's `MediaStream`, that cause video track has been attached to incorrect element. If this participant don't have any subsequent track events fired, then ui don't have chance to render the correct track. This cause the video chaos issue on my local cluster kill node test

### DIFF
--- a/.changeset/thirty-eagles-wait.md
+++ b/.changeset/thirty-eagles-wait.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+fix video attach to incorrect element when tranceiver reuse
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
great find!

davidzhao (approved):
really good fix!

PR #431, opened 2022-09-09
Add `RoomEvent.Connected`, fix connect future

lukasIO:
Regarding the connect future fix, we've seen a couple of instances of `b.reject is not defined` when quickly connecting/disconnecting (one of these got raised with react's strict mode). 
When constructing a future from a promise, previously there was no way to actually reject it from the future. This is fixed by not accepting a promise, but rather the function that runs within the promise in the future constructor. 
Also removed the `!` from the resolve and reject, as we cannot be sure those are actually populated.
In order to clean up stuff, there's an `onFinally` optional callback in the constructor.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -171,7 +171,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       })
       .on(EngineEvent.Resumed, () => {
         this.setAndEmitConnectionState(ConnectionState.Connected);
-        this.reconnectFuture?.resolve();
+        this.reconnectFuture?.resolve?.();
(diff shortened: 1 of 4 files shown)


src/room/Room.ts
```diff
@@ -171,7 +171,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       })
       .on(EngineEvent.Resumed, () => {
         this.setAndEmitConnectionState(ConnectionState.Connected);
-        this.reconnectFuture?.resolve();
+        this.reconnectFuture?.resolve?.();
```
davidzhao, line 174: ah ha, TIL!

davidzhao (approved):

PR #432, opened 2022-09-09
send previous server offer when reconnecting

cnderrauber:


src/proto/google/protobuf/timestamp.ts
```diff
@@ -1,8 +1,8 @@
 /* eslint-disable */
-import Long from 'long';
```
cnderrauber, line 2: changed by protoc version? have already run `yarn install`, but still change these file

lukasIO: strange, this looks like prettier changes. can you run `yarn format` to see if that helps?

lukasIO (approved):

src/proto/google/protobuf/timestamp.ts
cnderrauber: @lukasIO No change after `yarn format` ..

src/room/Room.ts
```diff
@@ -1050,9 +1051,16 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
 
     this.engine.client.sendSyncState({
       answer: toProtoSessionDescription({
-        sdp: previousSdp.sdp,
-        type: previousSdp.type,
+        sdp: previousAnswer.sdp,
+        type: previousAnswer.type,
       }),
+      offer:
+        previousOffer !== null
```
davidzhao, line 1058: default value is undefined, which is diff from null. using "trueness" operator would be fine in this case.

```suggestion
        previousOffer ? toProtoSessionDescription
```

PR #433, opened 2022-09-09
Add participant track subscribe method

lukasIO:
A follow up on #423 
This is mainly an idea on how to approach the fact that currently we rely on TrackPublications to be present in order to perform selective subscriptions and simplify this for users. 
With a method on the remote participant we could automatically subscribe to certain tracks once their publications arrive.
```typescript
room.on(RoomEvent.ParticipantConnected, (participant) => {
  // users wouldn't have to wait for the publications to arrive to signal their desire to subscribe to certain tracks
  // method takes in optional sources, if none are provided it will subscribe/unsubscribe from all sources
  participant.subscribeToTracks();
}
```
The method in this PR uses `Track.Source` to do that, which probably doesn't cover all cases, but seems like the most convenient way for most users.


lukasIO:
Interesting thought! 
I think it's a great way to further simplify state handling, but actually only in case `identity` is already known ahead of participant connections (I'm not sure how often this is the case). 

I guess we would want to create a dummy participant object already on the room if a call to `setSubscribed` on the room happens before a participant actually joins? That would allow us to merge them once the Participant actually connects and would avoid handling that state both on the room and on the participant level.

PR #436, opened 2022-09-12
Use explicit defaults for room and connect options

lukasIO:
In order to not have to pipe through config parameters as optional until they are actually used and (even more so) to have a single place in the code base where the default options are defined.


davidzhao (approved):
great cleanup!

PR #437, opened 2022-09-13
Restore previous behaviour for when to trigger track permission updates

lukasIO:
The code path for the permission updates is getting a bit hard to reason about. 
The reason why it's currently necessary is that a new `SubscriptionStatus: Desired` has been introduced and the tricky part is to not break the previous assumptions around `TrackSubscriptionPermissionUpdate`s along with it, because that event emits `SubscriptionStatus` (which has an additional state now). 
This change seems to do the trick right now, but it feels like a clear separation between what a `SubscriptionStatus` and a `PermissionStatus` is, would be a good call generally. Not sure if we'd want to do that already though, because it would change the return type of the `TrackSubscriptionPermissionUpdate` event.

### DIFF
--- a/src/room/track/RemoteTrackPublication.ts
+++ b/src/room/track/RemoteTrackPublication.ts
@@ -59,7 +59,7 @@ export default class RemoteTrackPublication extends TrackPublication {
   }
 
   get subscriptionStatus(): TrackPublication.SubscriptionStatus {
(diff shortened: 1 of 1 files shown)


boks1971 (approved):
Agreed separation sounds like a good path forward. May have to do a major version bump, but that is a big hammer for what seems like a small change.

PR #439, opened 2022-09-14
Update Package.json

jebas0812:
While using livekit-client , export package.json error occur while fix this it will working


lukasIO:
hi @jebas0812 
when did the error occur for you? 
it would be great if you could fill out an issue template, so that we can reproduce the issue

lukasIO:
closing due to missing feedback and in the hope that #460 fixed your original issue

PR #440, opened 2022-09-17
Sync state settings should use desired instead of subscribed

davidzhao:


boks1971 (approved):

davidzhao:
If a migration fails on the initial attempt, Tracks will become unsubscribed. However, when client attempts to resume the connection again, it will actually cause it to unsubscribe from all tracks.

Using @lukasIO's new isDesired attribute to ensure we don't accidentally unsubscribe from everything.

PR #443, opened 2022-09-20
Decouple permission and subscription status

lukasIO:
includes a breaking API change for `SubscriptionPermissionChanged` events:

`SubscriptionStatus` now only returns `desired | subscribed | unsubscribed`
And the `TrackSubscriptionPermissionChanged` event now returns `PermissionStatus` as `allowed | not_allowed`.


davidzhao (approved):
looks great! much cleaner.

PR #445, opened 2022-09-23
Increase event listener limit for publicly available emitters

lukasIO:
and some js docs for `LocalTrack` constructors


davidzhao (approved):
thank you for adding the comments here

PR #446, opened 2022-09-23
Add experimental option to pipe attached audio tracks through webaudio API

lukasIO:


src/room/track/RemoteAudioTrack.ts
```diff
@@ -48,9 +65,41 @@ export default class RemoteAudioTrack extends RemoteTrack {
...
     return element;
   }
 
+  /**
+   * @internal
+   */
+  setAudioContext(audioContext: AudioContext) {
+    this.audioContext = audioContext;
+    if (this.attachedElements.length > 0) {
+      this.setupWebAudio(audioContext, this.attachedElements[0]);
```
davidzhao, line 85: if attached to multiple elements, do we only need to handle the first one?

davidzhao: or would it be a limitation that it only handles a single audio element?

davidzhao (approved):
lg, I'll figure out a way to make testing audio a bit easier via the CLI.

src/room/track/RemoteAudioTrack.ts
lukasIO: for webaaudio we only need one element, as we're actually not interested in the element, but rather in the mediastream attached with it. we wouldn't want to pipe the same audio track multiple times into web audio

lukasIO: but just realised detach logic is still missing...

davidzhao: good call!

PR #450, opened 2022-09-28
add forceStereoAudioSupport option to Room constructor

mmckegg:
As far as I could tell, there is currently no way in the livekit client to stream stereo audio. Even when constraints are specified as `channels: 2`, the stream will always end up mono on the other end.

According to https://github.com/w3c/webrtc-extensions/issues/63, munging the sdp and adding `stereo=1` is the only way to make this happen. There is discussion to make this the default, but currently this is not the case.

From my limited research, it seems that adding this option has no affect on mono streams, it just allows stereo streams over the transport.

**This PR adds an additonal option `forceStereoAudioSupport` to the `Room` constructor which when enabled adds the sdp munging for stereo support if not already in the sdp**

The change as above solves my immediate use case problem, so I'll be using this fork until a solution lands in the main project. I'm not sure if this is the right solution for the main project, maybe it would be better to just always add the `stereo=1` option. I'll leave this up to the maintainers to decide! Hopefully this PR provides a good starting point.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -44,6 +44,12 @@ export interface InternalRoomOptions {
   publishDefaults?: TrackPublishDefaults;
 
   /**
+   * whether to force inclusion of stereo=1 in opus audio codec sdp, defaults to false
+   * see https://github.com/w3c/webrtc-extensions/issues/63
+   */
+  forceStereoAudioSupport: boolean;
+
+  /**
    * should local tracks be stopped when they are unpublished. defaults to true
    * set this to false if you would prefer to clean up unpublished local tracks manually.
    */
--- a/src/room/PCTransport.ts
+++ b/src/room/PCTransport.ts
@@ -22,10 +22,13 @@ export default class PCTransport {
 
   trackBitrates: TrackBitrateInfo[] = [];
 
+  forceStereoAudioSupport: boolean = false;
+
   onOffer?: (offer: RTCSessionDescriptionInit) => void;
 
-  constructor(config?: RTCConfiguration) {
+  constructor(config?: RTCConfiguration, forceStereoAudioSupport?: boolean) {
     this.pc = new RTCPeerConnection(config);
+    this.forceStereoAudioSupport = forceStereoAudioSupport ?? false;
   }
 
   get isICEConnected(): boolean {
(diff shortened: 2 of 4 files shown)


lukasIO:
hi @mmckegg, 
have you tried publishing stereo with the latest livekit-server version (v1.2.3) ?
opus stereo support just landed in that version!

mmckegg:
Oh really! I will give that a go. If so, sorry about the PR noise. But the good news is that I am now a bit more familiar with the code base and might actually be able to contribute something useful in the future 😉

mmckegg:
Yep, upgrading solved it. Thanks @lukasIO, y'all are moving at an awesome speed! This project is making my life so much easier.

Good reminder for me to always check every day for new releases ❤️

bekriebel:
@mmckegg How were you able to get a stereo track to publish? I'm verifying that my audio track is capturing with 2 channels, but on my receiving end it still appears to be only getting a single channel of audio.

mmckegg:
@bekriebel first make sure you are running the latest livekit server (as mentioned above).

Need to be sure that `echoCancellation` is not enabled on getUserMedia audio sources. It is by default, unless you specifically disable it, and in most browsers, it [causes stereo audio to become mono](https://bugs.chromium.org/p/webrtc/issues/detail?id=10465).

Here's my getUserMedia and publish that seems to result in good quality audio:

```ts
const stream = await navigator.mediaDevices.getUserMedia({
    video: false,
    audio: {
      channelCount: { ideal: 2 },
      echoCancellation: { ideal: false },
    },
  })
  
const publishedTrack = await room.localParticipant.publishTrack(stream.getAudioTracks()[0], { audioBitrate: 64000 })
```

I found that 64000 is the bare minimum to make music sound acceptable. 96000 was pretty much spotify quality to my ears.

Hope this helps 🙃

bekriebel:
@mmckegg Huh... Yeah, I'm doing similar. I'm actually disabling `echoCancellation`, `autoGainControl`, and `noiseSuppression` and have the same ideal channel count. Locally, I can see that my `mediaStreamTrack` has two channels, but if I look at the `mediaStreamTrack` for the user on the remote side, `channelCount` is still `1`. This is on a server that I just upgraded to v1.2.3.

bekriebel:
This is the original issue I opened about this, and the information I provided there is still what I'm finding to be true: https://github.com/livekit/livekit/issues/970

Have you confirmed on a receiving client that the track is coming through as stereo?

mmckegg:
@bekriebel yep, it’s definitely stereo on the receiving client. I haven’t checked the channel, count but my headphones told me it was stereo 😎

bekriebel:
As mentioned in https://github.com/livekit/livekit/issues/970, something similar to this is actually needed after all. From my testing, this doesn't need to be sent on the SDP Offer, only the Answer. I'm not sure if there's any benefit to setting it on both. I'm also not sure if this should be put under a room setting, or if it should just be enabled by default.

PR #452, opened 2022-09-28
fix setting audioBitrate on unknown track sources

mmckegg:
If a track is created using `localPartipant.publishTrack` without a source specified (or source: unknown), then the `audioBitrate` option would be ignored. This was due to the code using track source instead of track type for the check.

However, if an app needs to have multiple audio tracks (which mine does), then `source: unknown` is required to prevent automatic unwanted de-duplication when also using microphone api.

**This PR changes the check for adding audioBitrate to use track type instead of source, and now the audioBitrate option is added for all audio tracks**

### DIFF
--- a/.changeset/fresh-pens-enjoy.md
+++ b/.changeset/fresh-pens-enjoy.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+fix setting audioBitrate on unknown track sources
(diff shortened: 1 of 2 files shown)


lukasIO:
hi @mmckegg,
thank you for the PR! 
I'm trying to follow your description where `Track.Source` is used for checking, but I can only see `Track.Kind` being used, which should work just fine with `Track.Source.Unknown`.

lukasIO:
you are right, this is rather unexpected! Will take a look at how to fix this

PR #453, opened 2022-09-28
Use exact source matching for `getTrack`

lukasIO:
A user reported a problem when using multiple audio tracks alongside the `setMicrophoneEnabled` method. 
What happened was that `getTrack`, which is used within `setTrackEnabled`, would return a "best guess" track based on the assumption that every audio track that's not named `screen` could potentially be a microphone track. 

This PR introduces an optional `exact` parameter for the `getTrack` function to force the function to only return exact source matches. 
The only place where this is being actively used is `setTrackEnabled`, where we can assume control over the local tracks and thus can be sure that an adequate source has been set.

thanks @mmckegg for the report!

### DIFF
--- a/src/room/participant/Participant.ts
+++ b/src/room/participant/Participant.ts
@@ -88,10 +88,11 @@ export default class Participant extends (EventEmitter as new () => TypedEmitter
   /**
    * Finds the first track that matches the source filter, for example, getting
    * the user's camera track with getTrackBySource(Track.Source.Camera).
-   * @param source
-   * @returns
+   * @param {Track.Source} source
+   * @param {boolean | undefined} exact - Specifies whether only exact source matches should be returned or if a best guess is acceptable
+   * @returns {TrackPublication | undefined }
    */
-  getTrack(source: Track.Source): TrackPublication | undefined {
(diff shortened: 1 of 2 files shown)


src/room/participant/Participant.ts
```diff
@@ -88,45 +88,50 @@ export default class Participant extends (EventEmitter as new () => TypedEmitter
...
-  getTrack(source: Track.Source): TrackPublication | undefined {
+  getTrack(source: Track.Source, exact?: boolean): TrackPublication | undefined {
     if (source === Track.Source.Unknown) {
       return;
     }
     for (const [, pub] of this.tracks) {
       if (pub.source === source) {
         return pub;
       }
-      if (pub.source === Track.Source.Unknown) {
```
davidzhao, line 102: I think we can remove this chunk of code now. This was done for backwards compatibility when we initially introduced Track Source. (before 1.0).

That would simplify things and avoiding introducing new params to the API.

lukasIO: even better!

davidzhao (approved):
lg!

PR #455, opened 2022-09-30
Retry initial signal connection

lukasIO:


src/api/SignalClient.ts
```diff
@@ -210,11 +223,21 @@ export class SignalClient {
...
+                await retryConnection();
+              }
+              reject(new ConnectionError(msg, resp.status));
             } else {
-              reject(new ConnectionError('Internal error'));
+              reject(new ConnectionError('Internal error', resp.status));
             }
           } catch (e) {
+            if (this.reconnectAttempts === 0) {
+              await retryConnection();
```
davidzhao, line 239: do we need to return after this to avoid the `reject` below from getting used?

davidzhao (commented):
wondering if retries should be handled within SignalClient, versus in RTCEngine where a bit more of the business logic would normally live. I think we could still handle the authentication error differently by throwing a different error after validate.

src/api/SignalClient.ts
lukasIO: the `retryConnection` funciton is using the `resolve` of the connect function. so if a reconnection is successful the promise gets resolved before it could reach `reject`

lukasIO:
sure, moving it into `RTCEngine` seems like a good idea, as this already has some reconnect logic in it!

src/room/Room.ts
```diff
@@ -357,12 +357,12 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
           navigator.mediaDevices?.addEventListener('devicechange', this.handleDeviceChange);
         }
         this.setAndEmitConnectionState(ConnectionState.Connected);
+        this.emit(RoomEvent.Connected);
         resolve();
       });
     };
     this.connectFuture = new Future(connectFn, () => {
       this.clearConnectionFutures();
-      this.emit(RoomEvent.Connected);
```
lukasIO, line 365: this was falsely emitting a connected event, even if the connection future was rejected

lukasIO:
@davidzhao I thought about reusing the reconnectPolicy, but felt it's more predictable if initial reconnect attempts can be configured separately. 
One downside is that currently there's no way to report back to the user that a join attempt has failed.

src/options.ts
```diff
@@ -54,6 +54,9 @@ export interface InternalRoomOptions {
    */
   reconnectPolicy: ReconnectPolicy;
 
+  /** specifies how often an initial join connection is allowed to retry (only applicable if server is not reachable) */
+  maxJoinAttempts: number;
```
davidzhao, line 58: nit, should this be available in `ConnectOptions` instead? since it's related to the connect action. Then we could just call it `maxRetries` ?

davidzhao (approved):

Issue #459, opened 2022-10-03
Add missing ESM types subpath export condition.


scott-lc:
### Describe the bug

The TypeScript 4.7+ `NodeNext` module resolver does not honor the legacy `types` property in the `package.json` file.  You need to add a `types` export condition as described [here](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing).

> Note that the "types" condition should always come first in "exports".

```jsonc
{
  "exports": {
    ".": {
      "types": "./dist/src/index.d.ts",
      "require": "./cjs/nats.js",
      "import": "./esm/nats.js"
    }
  }
```

or, since the "." export path is optional if there is only one element:

```jsonc
{
  "exports": {
      "types": "./dist/src/index.d.ts",
      "require": "./cjs/nats.js",
      "import": "./esm/nats.js"
  }
```

### Reproduction

```
git clone git@github.com:scott-lc/livekit-nodenext-issue.git
cd ./livekit-nodenext-issue
npm install
npm run build
```

Compilation failure - 

```
index.ts:1:22 - error TS7016: Could not find a declaration file for module 'livekit-client'. '.../livekit-nodenext-issue/node_modules/livekit-client/dist/livekit-client.esm.mjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/livekit-client` if it exists or add a new declaration (.d.ts) file containing `declare module 'livekit-client';`

1 import { Room } from "livekit-client";
                       ~~~~~~~~~~~~~~~~


Found 1 error in index.ts:1
```

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.50 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.17.0/bin/npm
    Watchman: 2022.09.19.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 104.0.5112.101
    Edge: 101.0.1210.32
    Firefox: 105.0.1
    Safari: 16.0
```

### Severity

blocking all usage of LiveKit

### Additional Information

Will submit PR shortly.

[...]

PR #460, opened 2022-10-03
Added missing ESM types subpath export condition.

scott-lc:
Fix for #459.


lukasIO:
hi @scott-lc,
this is great, thank you for the fix!

PR #462, opened 2022-10-03
Fix local participant events being disabled before all local tracks are unpublished

### LINKED ISSUE #461: RoomEvent.LocalTrackUnpublished events are not sent when disconnecting a Room
### Describe the bug

When calling `room.disconnect()`, events that the local participant send during disconnect are not received by listeners setup on the room. For example, `room.on(RoomEvent.LocalTrackUnpublished, ...)` will not be received because the [`handleDisconnect()` implementation in Room disables all of its event listeners on the local participant before unpublishing the tracks](https://github.com/livekit/client-sdk-js/blob/84f00b893d46bc7b1d0d6fa3785ace85f08f0584/src/room/Room.ts#L706).

This bug seems to have been introduced in [v1.2.3](https://github.com/livekit/client-sdk-js/releases/tag/v1.2.3) with #323.

### Reproduction

1. Create a room object and setup `RoomEvent.LocalTrackPublished` and `RoomEvent.LocalTrackUnpublished` event listeners on it.
2. Connect to the room using `room.connect()`.
3. Notice that the `RoomEvent.LocalTrackPublished` listener gets called.
4. Disconnect from the room.
5. Notice that the `RoomEvent.LocalTrackUnpublished` listener does not get called.

### Logs

_No response_

### System Info

```shell
N/A
```

### Severity

### Additional Information

_No response_

KallynGowdy:
Fixes #461

### DIFF
--- a/.changeset/five-clouds-beam.md
+++ b/.changeset/five-clouds-beam.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Fix local participant events being disabled before tracks unpublished
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -703,6 +703,16 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       });
     });
 
+    this.localParticipant.tracks.forEach((pub) => {
+      if (pub.track) {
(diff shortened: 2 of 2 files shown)


davidzhao (approved):
Thank you for catching & fixing this 🙏

PR #467, opened 2022-10-10
Update ice server from join response

cnderrauber:
When ice servers from join response changed while reconnected (fallback to tls relay), should use new ice servers to create peerconnection

### DIFF
--- a/.changeset/khaki-turkeys-work.md
+++ b/.changeset/khaki-turkeys-work.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


lukasIO (approved):

imredobos:
Hi @cnderrauber , @lukasIO , @davidzhao i feel like we are loosing the opportunity to set ice servers from client side with this change. I do not want to rely on Livekit's embedded nor configured TURN, I want to handle it in a separate workflow and set it in the client side before create the peer connection. What idea do you have to achive this?

PR #470, opened 2022-10-14
Support track level stereo and red setting

cnderrauber:


cnderrauber:
Fix https://github.com/livekit/livekit/issues/970

src/room/PCTransport.ts
```diff
@@ -233,5 +239,43 @@ function ensureAudioNack(
...
+        if (rtp.codec === 'opus') {
+          opusPayload = rtp.payload;
+          return true;
+        }
+        return false;
+      });
+
+      media.fmtp.some((fmtp): boolean => {
+        if (fmtp.payload === opusPayload) {
+          if (fmtp.config.includes('sprop-stereo=1')) {
```
boks1971, line 271: Does `server` have to send this?

src/room/participant/LocalParticipant.ts
```diff
@@ -214,6 +214,29 @@ export default class LocalParticipant extends Participant {
...
+          }
+          if (channelCount && channelCount > 1) {
+            if (!publishOptions) {
+              publishOptions = { red: false, dtx: false };
+            } else {
+              if (publishOptions.red !== true) {
+                publishOptions.red = false;
+              }
+              if (publishOptions.dtx !== true) {
+                publishOptions.dtx = false;
```
boks1971, line 234: This and above is already checking for not true and setting to false. Should the check be `if set to true, override it with false when stereo`?

src/room/PCTransport.ts
```diff
@@ -233,5 +239,43 @@ function ensureAudioNack(
         type: 'nack',
       });
     }
+
+    if (stereoMids.includes(media.mid!)) {
+      media.fmtp.some((fmtp): boolean => {
+        if (fmtp.payload === opusPayload) {
+          fmtp.config += ';stereo=1';
```
boks1971, line 246: Should this check if config already has `stereo=1` and add only if not there? i. e. will multiple pass through here keep adding `stereo=1`? Guess it probably does not affect functionality, but the SDP could look strange.

src/room/PCTransport.ts
cnderrauber: yeah, it should have it if sending a stereo track

src/room/participant/LocalParticipant.ts
cnderrauber: the red and dtx can be left to undefined, means use `default` values. So this only disable them if dtx and red not enabled explicitly

src/room/PCTransport.ts
cnderrauber: Test it will not be added repeatedly in chrome, but the check is safe for various browsers, will add it

boks1971 (approved):

src/room/participant/LocalParticipant.ts
lukasIO: should this logic maybe live in `publishTrack` instead of `setTrackEnabled` ? 
If users choose to use the `publishTrack` API directly, I guess they would have to handle all this by themselves otherwise?

davidzhao: yeah I agree this should be in publishTrack.

why do we need to set dtx or red to default false? It seems that we shouldn't change the default behavior of other unrelated params.

also, the logic would be easier to read with the `??` operator

```
publishOptions ??= {};
publishOptions.red ??= false
publishOptions.dtx ??= true
```

src/room/participant/LocalParticipant.ts
```diff
@@ -214,6 +214,29 @@ export default class LocalParticipant extends Participant {
         }
         this.pendingPublishing.add(source);
         try {
+          // disable red and dtx for stereo track if not enabled explicitly
+          let channelCount: ConstrainULong | undefined;
+          if (options) {
+            if ('channelCount' in options) {
```
davidzhao, line 220: when this is moved to publishTrack, determine the actual channelCount will be a challenge. we could take in a `stereo?: boolean` parameter in publishOptions too, to give user control in case we cannot detect correctly.

In Safari, we could get channelCount from [MediaStreamTrack.getSettings()](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings/channelCount). but Chrome doesn't seem to support that.. which means we may have to go look for it in the constraints.

davidzhao (changes requested):

src/room/participant/LocalParticipant.ts
lukasIO: it's also not fully fail proof but chrome should support `getConstraints()` that would return the channel count in case it was manually set.

src/room/participant/LocalParticipant.ts
cnderrauber: @lukasIO sounds good, will move it there.
@davidzhao For stereo audio tracks, it will use high bitrates encoding (128kbps or higher) and most scenario is music, RED will cost more bandwidth and DTX might cause noise in special case or device.

cnderrauber: solved by https://github.com/livekit/client-sdk-js/pull/470/commits/9b8a0a4dd1127ee88fd481fd3343f6af2218589f

lukasIO (approved):

davidzhao (commented):
looks good! remaining question about defaults

src/room/participant/LocalParticipant.ts
```diff
@@ -423,6 +418,31 @@ export default class LocalParticipant extends Participant {
...
+    if (isStereo) {
+      if (!options) {
+        options = {};
+      }
+      if (options.red === undefined || options.dtx === undefined) {
+        log.warn(
+          `Opus RED and DTX will be disabled for stereo tracks by default. Enable them explicitly to make it work.`,
+        );
+      }
+      options.red ??= false;
```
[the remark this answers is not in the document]
bekriebel: I haven't had the chance to catch up on this whole thread, but I'm excited to test it out later this week!

Regarding red support, I will say that from my testing - high bitrate stereo + red (with DTX, noise canceling, echo canceling, etc. all turned off), audio sounds AMAZING. Using similar bitrates with opus by itself doesn't sound nearly as good. For my "quality audio mode" I plan to leave red enabled.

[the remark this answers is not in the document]
davidzhao: thanks for the feedback @bekriebel ! that's wonderful to hear it made a significant difference in audio quality.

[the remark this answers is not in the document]
cnderrauber: It is hard to reproduce, but a few users report noise with special music cases before, and we close DTX to fix it. At least, opus decoder will generate `comfort noise` when received dtx packet, that is no needed and might harm to quality in music case. So I suggest disable dtx in stereo mode

davidzhao (approved):
lg!

PR #475, opened 2022-10-18
Fix stuck reconnect

lukasIO:
- adds an `onLine` listener to initiate a reconnect attempt if the browser detects that the network status has switched to `online`
- gets rid of `reconnectFuture` and instead overrides existing reconnection attempts when a new call to `connect` happens
- uses events (instead of the `reconnectFuture`) to defer adding tracks while connecting/reconnecting
- makes sure pending reconnects are disposed when recreating engine


davidzhao (approved):
lg!

PR #484, opened 2022-10-24
Add fps to subscribe setting

cnderrauber:


src/room/track/RemoteTrackPublication.ts
```diff
@@ -267,6 +283,7 @@ export default class RemoteTrackPublication extends TrackPublication {
       settings.quality = VideoQuality.HIGH;
     }
 
+    log.info('track update', settings);
```
lukasIO, line 286: should this be `debug` level instead?

src/room/track/RemoteTrackPublication.ts
```diff
@@ -143,6 +145,19 @@ export default class RemoteTrackPublication extends TrackPublication {
     this.emitTrackUpdate();
   }
 
+  setVideoFPS(fps: number) {
+    if (!this.isManualOperationAllowed()) {
```
lukasIO, line 149: should we add a check here to make sure `kind` is of kind video?

src/room/track/RemoteTrackPublication.ts
```diff
@@ -143,6 +145,23 @@ export default class RemoteTrackPublication extends TrackPublication {
     this.emitTrackUpdate();
   }
 
+  setVideoFPS(fps: number) {
```
lukasIO, line 148: nit: as we do not know for certain that this will actually result in the desired fps, maybe naming it `setPreferredVideoFPS` would be communicating more clearly what it is doing?

lukasIO (approved):

src/room/track/RemoteTrackPublication.ts
cnderrauber: Yeah, it is more clear, we have discussed this during the subscribed fps design, just want to keep same as `setVideoDimensions`, it is a preference too.

PR #492, opened 2022-10-26
Only emit metadatachanged if the metadata actually changed

lukasIO:


davidzhao (approved):
thanks for fixing this!

PR #494, opened 2022-10-26
Fix safari 16 screen share

lukasIO:


davidzhao (approved):
wow.. Safari strikes again

Issue #496, opened 2022-10-28
Revoking camera access in browser mutes remote track but doesn't mute local track


davideberlein:
#### Steps to reproduce
1.  start a call
2. publish a `MediaStreamTrack` using `room.localParticipant.publishTrack(videoTrack,...)`
3. revoke camera access permission (in chrome)

#### Expected behavior
- The LocalTrackPublication is muted and the `localTrackMuted` event is triggered
- The RemoteTrackPublication is muted for anyone subscribed to the published camera and the `trackMuted` event is triggered.

#### Actual behavior
- The LocalTrackPublication is **not** muted and *no* `localTrackMuted` event is triggered (however the `LocalTrackPublication.isUpstreamPaused` is set to true and the `LocalTrack.upstreamPaused` event is trigged)
- The RemoteTrackPublication *is* muted for anyone subscribed to the published camera and the `trackMuted` event *is* triggered.

### System Info

```shell
OS: Mac OSX 12.5.1 (21G83)
Browser: Chrome Version 106.0.5249.61 (Official Build) (arm64)
livekit-js: 1.5.0
```

### Severity

serious, but I can work around it

### Additional Information

It looks like `LocalParticipant.onTrackUpstreamPaused` handles `UpstreamPaused` the same way as `Muted` which is why the state is correctly propagated to the remote participants but not the local participant.

PR #498, opened 2022-10-28
Mute track if ended event has been fired

lukasIO:
fixes https://github.com/livekit/client-sdk-js/issues/496
Considering the fact that after the `ended` event we will not be able to reuse that track again ever, we might as well mute it to surface the fact that the local publication isn't currently active anymore.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -965,6 +965,7 @@ export default class LocalParticipant extends Participant {
       this.unpublishTrack(track);
     } else if (track.isUserProvided) {
       await track.pauseUpstream();
+      await track.mute();
     } else if (track instanceof LocalAudioTrack || track instanceof LocalVideoTrack) {
       try {
         if (isWeb()) {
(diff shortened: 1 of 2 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -965,6 +965,7 @@ export default class LocalParticipant extends Participant {
       this.unpublishTrack(track);
     } else if (track.isUserProvided) {
       await track.pauseUpstream();
+      await track.mute();
```
davidzhao, line 968: if we are muting the track, is pauseUpstream still needed?

davidzhao (approved):

src/room/participant/LocalParticipant.ts
lukasIO: the only reason I left it in is to not break the behaviour for people who have relied on those events to be emitted. 
But realised just now, that it's marked as an internal event anyways, so should be safe to remove!

davidzhao: sounds good!

PR #515, opened 2022-11-28
Add audio analyser util function

lukasIO:


example/sample.ts
```diff
@@ -112,7 +114,16 @@ const appActions = {
...
       })
-      .on(RoomEvent.LocalTrackPublished, () => {
+      .on(RoomEvent.LocalTrackPublished, (pub) => {
+        const track = pub.track as LocalAudioTrack;
+
+        if (track instanceof LocalAudioTrack) {
+          const { calculateVolume } = createAudioAnalyser(track);
+
+          setInterval(() => {
+            console.log(calculateVolume());
```
davidzhao, line 124: nit: this would seem pretty noisy in the console. maybe less frequently or as a bonus, hooking it up to a jank meter?

src/room/utils.ts
```diff
@@ -236,3 +239,62 @@ export class Future<T> {
...
+  minDecibels?: number;
+  maxDecibels?: number;
+};
+
+/**
+ * Creates and returns an analyser web audio node that is attached to the provided track.
+ * Additionally returns a convenience method `calculateVolume` to perform instant volume readings on that track.
+ * Call the returned `cleanup` function to close the audioContext that has been created for the instance of this helper
+ */
+export function createAudioAnalyser(
```
davidzhao, line 255: should we handle track attachment bug in here? i.e. log a warning if it's not yet attached.

davidzhao (approved):

src/room/utils.ts
lukasIO: In my tests, the attachment didn't matter. I had a separate path for RemoteTracks initially that would have handled this case, but it turns out it also works for the analysing use case without any attachments!

PR #523, opened 2022-12-02
Include exception message in connection error

lukasIO:


src/room/Room.ts
```diff
@@ -342,7 +342,12 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       } catch (err) {
         this.recreateEngine();
         this.handleDisconnect(this.options.stopLocalTrackOnUnpublish);
-        reject(new ConnectionError('could not establish signal connection'));
+        let errorMessage = '';
+        if (err instanceof Error) {
+          errorMessage = err.message;
+          log.debug(`error trying to establish signal connection`, { error: err });
+        }
+        reject(new ConnectionError(`could not establish signal connection ${errorMessage}`));
```
boks1971, line 350: maybe a semi-colon after `could not establish signal connection:` to separate what is our log and what comes in the error message. Not sure about front end conventions, but I like to do that so that it is easy visual separation on what we are logging and what is coming from elsewhere.

boks1971 (approved):

PR #526, opened 2022-12-06
Unlock before returning

lukasIO:


davidzhao (approved):

src/room/Room.ts
```diff
@@ -402,6 +402,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     try {
       if (this.state === ConnectionState.Disconnected) {
         log.debug('already disconnected');
+        unlock();
```
dorr91, line 405: Hey @lukasIO I think the unlock in the `finally` clause should catch this early return. Does unlocking twice cause issues?

lukasIO: ah, of course, you're right. was overly eager to catch those early returns... 
shouldn't cause any problems, but will revert in order to keep it clean.

PR #531, opened 2022-12-08
Add method to populate an offline room with simulated participants

lukasIO:
Would be great to add additional features in the future that mimic remote participant behaviour. 
As a first attempt, this keeps remote participants static.

### DIFF
--- a/src/room/utils.ts
+++ b/src/room/utils.ts
@@ -181,20 +181,39 @@ let emptyVideoStreamTrack: MediaStreamTrack | undefined;
 
 export function getEmptyVideoStreamTrack() {
   if (!emptyVideoStreamTrack) {
-    const canvas = document.createElement('canvas');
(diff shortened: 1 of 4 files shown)


src/room/utils.ts
```diff
@@ -181,22 +181,41 @@ let emptyVideoStreamTrack: MediaStreamTrack | undefined;
...
-    if (!emptyVideoStreamTrack) {
-      throw Error('Could not get empty media stream video track');
-    }
-    emptyVideoStreamTrack.enabled = false;
+    emptyVideoStreamTrack = getDummyVideoStreamTrack();
   }
   return emptyVideoStreamTrack;
 }
 
+export function getDummyVideoStreamTrack(
```
davidzhao, line 189: nit: 
```suggestion
export function createDummyVideoStreamTrack(
```

davidzhao (approved):
lg!

PR #534, opened 2022-12-12
Improve autoplay when multiple streams are attached to the same media element

lukasIO:
A couple of improvements for auto playback. 
Thanks @davideberlein for the suggestions!

1. Emits `RoomEvent.AudioPlaybackStatusChanged` not only when tracks emit it, but also if `acquireAudioContext` fails 
2. Sets video elements to `muted` if there's no audio stream attached to it to make sure autoplay is working
3. in case there is an audio stream present, it will try to start playing. if it fails it sets the element to muted and thus makes sure that at least the video is playing.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -1045,6 +1045,12 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     if (this.options.expWebAudioMix) {
       this.participants.forEach((participant) => participant.setAudioContext(this.audioContext));
     }
+
+    const newContextIsRunning = this.audioContext?.state === 'running';
+    if (newContextIsRunning !== this.canPlaybackAudio) {
+      this.audioEnabled = newContextIsRunning;
+      this.emit(RoomEvent.AudioPlaybackStatusChanged, newContextIsRunning);
+    }
   }
 
   private createParticipant(id: string, info?: ParticipantInfo): RemoteParticipant {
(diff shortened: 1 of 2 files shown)


src/room/track/Track.ts
```diff
@@ -121,7 +121,7 @@ export abstract class Track extends (EventEmitter as new () => TypedEventEmitter
     // we'll want to re-attach it in that case
     attachToElement(this._mediaStreamTrack, element);
 
-    if (element instanceof HTMLAudioElement) {
+    if ((element.srcObject as MediaStream).getAudioTracks().length > 0) {
       // manually play audio to detect audio playback status
       element
         .play()
```
davideberlein, line 127: I think you would need to do the following in line `131`
```ts
.catch(e => {
     this.emit(TrackEvent.AudioPlaybackFailed, e);
     element.muted = true;
     return element.play();
})
```

src/room/Room.ts
```diff
@@ -1045,6 +1045,12 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     if (this.options.expWebAudioMix) {
```
davideberlein, line 1045: I think in line `537` you would have to add:
```ts
try {
      await Promise.all(elements.map((e) => {
            e.muted = false;
            e.play();
      }));
      this.handleAudioPlaybackStarted();
    } catch (err) {
      this.handleAudioPlaybackFailed(err);
      throw err;
    }
```

src/room/track/Track.ts
davideberlein: I agree, either you have to prohibit using the same video element for both or do it this way

davidzhao (approved):
lgtm

PR #539, opened 2022-12-21
Add internal method to await data channel connection

lukasIO:


src/room/RTCEngine.ts
```diff
@@ -915,39 +915,54 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-    }
-
-    if (!this.publisher) {
-      throw new ConnectionError('publisher connection not set');
+  /**
+   * @internal
+   */
+  async ensureDataTransportReady(
+    kind: DataPacket_Kind,
+    subscriber: boolean = this.subscriberPrimary,
```
boks1971, line 923: Wow you can default using class variable. TIL

src/room/RTCEngine.ts
```diff
@@ -915,39 +915,54 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
 
     throw new ConnectionError(
-      `could not establish publisher connection, state ${this.publisher?.pc.iceConnectionState}`,
+      `could not establish ${transportName} connection, state ${primaryTransport.pc.iceConnectionState}`,
     );
   }
 
+  private async ensurePublisherConnected(kind: DataPacket_Kind) {
+    if (!this.subscriberPrimary) {
+      return;
```
boks1971, line 960: Do we want to return here? IIRC, all upstream data always goes via publisher PC data channel. Only for the downstream data, it comes down via either depending on which one is primary.

lukasIO: Returning is the behaviour `ensurePublisherConnected` always had. 
I think the assumption is if the primary PC is the publisher, then we'll have the publisher connected anyways already? 
but yeah, we still cannot be sure that the data channel is in an `open` state at that point 🤔

src/room/RTCEngine.ts
```diff
@@ -915,39 +915,54 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
 
-    if (!this.publisher.isICEConnected && this.publisher.pc.iceConnectionState !== 'checking') {
+    if (
+      this.subscriberPrimary &&
+      !this.publisher?.isICEConnected &&
+      this.publisher?.pc.iceConnectionState !== 'checking'
+    ) {
       // start negotiation
       this.negotiate();
     }
```
boks1971, line 938: Can this be done as
```
if transport == this.publisher && !transport.isICEConnected && !transport.pc.iceConnectionState !== 'checking'
```
?

I got quite confused reading it. Seems like this is trying to set up publisher PC as long as it is not set up and subscriber is primary. But, the intention of the function seems to be ensure that a peer connection is set up when necessary. So, if `ensureDataTransportConnected` is called to ensure that subscriber PC data transport is connected, it would still negotiate the publisher PC. Am I reading this right?

The change I am suggesting keeps it purely based on what is supplied in `subscriber` argument. If `subscriber = true`, publisher PC will not be touched. If called with `subcriber = false`, publisher PC will be set up if necessary.

lukasIO: ah yes, you are right, that's an oversight. your suggestion sounds good!

src/room/RTCEngine.ts
```diff
@@ -915,39 +915,54 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
   }
 
-  private async ensurePublisherConnected(kind: DataPacket_Kind) {
-    if (!this.subscriberPrimary) {
-      return;
```
lukasIO, line 920: @davidzhao this PR gets rid of this early return. The reasoning being that even if `!this.subscriberPrimary` we don't know whether the associated data channel is already in an `open` state, so it's safer to perform the check anyways.
Wanted to double-check with you if there were any intentions behind this early return that I might be unaware of?

davidzhao: Initially when I had this in here, this function relied on PeerConnection state before deciding passing. Hence when the room is connected, publisher connection should be established, I didn't think it was needed to wait.

Since then, we changed to checking for data channel state.. so this is still useful. Thanks for catching this!

davidzhao (approved):

src/room/RTCEngine.ts
lukasIO: yeah, thanks @boks1971 for catching this 😅

PR #543, opened 2022-12-30
Wait for dimensions to become available prior to publishing

davidzhao:
Video dimensions are required for simulcast publishing to take place. When certain transformations are applied to the MediaStreamTrack, it could take a bit of time for getSettings() to return the correct dimensions.

Currently this will fallback to the old behavior. In future versions, we could make dimensions a requirement.


lukasIO (approved):
LGTM!
Also making it a requirement in the future sounds reasonable!

PR #549, opened 2023-01-02
Clear negotiation handler if engine is closed during negotiation

lukasIO:


src/room/RTCEngine.ts
```diff
@@ -185,8 +185,12 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   async close() {
     const unlock = await this.closingLock.lock();
+    if (this.isClosed) {
```
davidzhao, line 188: should it unlock as well?

lukasIO: thanks for catching this!

davidzhao (approved):
lgtm!

PR #553, opened 2023-01-11
add reconnect response to update configuration while reconnecting

cnderrauber:


src/api/SignalClient.ts
```diff
@@ -279,14 +272,28 @@ export class SignalClient {
               this.startPingInterval();
             }
             resolve(resp.message.join);
-          } else {
+          } else if (opts.reconnect) {
+            // in reconneting, any message received means signal reconnected
```
lukasIO, line 276: typo 
```suggestion
            // in reconnecting, any message received means signal reconnected
```

src/api/SignalClient.ts
```diff
@@ -264,6 +256,7 @@ export class SignalClient {
         }
 
         if (!this.isConnected) {
+          let continueProcess = false;
```
lukasIO, line 259: nit: very subjective, but the meaning of `continueProcess` wasn't immediately clear to me. Maybe something like `shouldProcessMessage` would have been more obvious to me right away.

lukasIO (approved):

PR #555, opened 2023-01-13
Enable nack for audio track only if offer want it

cnderrauber:


.changeset/rare-rivers-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Enable nack for audio track only offer wants it
```
lukasIO, line 5: typo
```suggestion
Enable nack for audio track only if offer wants it
```

lukasIO (approved):

PR #556, opened 2023-01-22
Allow for background timers to be overriden

davidliu:


src/room/timers.ts
```diff
@@ -1,7 +1,9 @@
 /**
- * Timers that can be overridden with platform specific implementations.
+ * Timers that can be overridden with platform specific implementations
+ * that ensure that they are fired. These should be used when it is critical
```
lukasIO, line 3: nice! that makes it a lot clearer!

lukasIO (approved):

PR #557, opened 2023-01-23
Experimental end-to-end encryption support

lukasIO:


initpwn:
Status of this PR?

src/e2ee/worker/e2ee.worker.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+}
+
+function setupCryptorErrorEvents(cryptor: Cryptor) {
+  cryptor.on('cryptorError', (error) => {
+    const msg: ErrorMessage = {
+      kind: 'error',
+      data: { error },
+    };
+    postMessage(msg);
```
max-b, line 188: I believe this is an instance of attempting to pass an error w/ postMessage which browsers generally don't allow?
https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage
https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

src/e2ee/worker/Cryptor.ts
```diff
@@ -0,0 +1,601 @@
...
+      .pipeThrough(transformStream)
+      .pipeTo(writable)
+      .catch((e) => {
+        const errorMsg: ErrorMessage = {
+          kind: 'error',
+          data: {
+            error: new E2EEError(e.message, E2EEErrorReason.InternalError),
+          },
+        };
+        postMessage(errorMsg);
```
max-b, line 148: Same issue as https://github.com/livekit/client-sdk-js/pull/557/files#r1156612686 I believe

src/e2ee/worker/e2ee.worker.ts
lukasIO: thanks for the reminder 🙏

yarn.lock
```diff
@@ -4466,6 +4454,12 @@ glob@^7.0.3, glob@^7.1.3, glob@^7.1.4:
     once "^1.3.0"
     path-is-absolute "^1.0.0"
 
+<<<<<<< HEAD
```
cloudwebrtc, line 4457: There are some merge conflicts in the .lock file, you may need to override yarn.lock

src/room/participant/Participant.ts
```diff
@@ -68,6 +68,10 @@ export default class Participant extends (EventEmitter as new () => TypedEmitter
 
   private _connectionQuality: ConnectionQuality = ConnectionQuality.Unknown;
 
+  get isEncrypted() {
+    return this.tracks.size > 0 && Array.from(this.tracks.values()).every((tr) => tr.isEncrypted);
```
davidzhao, line 72: so participant.isEncrypted is only true when every track is encrypted?

lukasIO: yeah, that seems the like the least confusing option. In theory all tracks published by a participant should be encrypted anyways after enabling e2ee. 
Anything that you would think is unexpected with that approach?

davidzhao: Nope.. just thinking through how this option would be used. But that makes sense to me.

cloudwebrtc (approved):
lgtm

lukasIO:
confirmed this is working in conjunction with track-processors. Will cut a new patch release for livekit-client first before merging this in as a minor change

PR #562, opened 2023-01-30
Add getDisplayMedia support check

kilimnik:
I noticed that on Android, the `getDisplayMedia` function is undefined. Therefore I added a more verbose error message.


lukasIO:
That's great! Thank you for your contribution :)

lukasIO (approved):

PR #568, opened 2023-02-01
Handle unknown signal messages from the server

davidzhao:
When we switched to [using oneof](https://github.com/livekit/client-sdk-js/pull/379) in generated protobufs, it broke the ability to handle unknown signal messages from the server. This is due to the protobuf parser returning `undefined` for `message` field, versus a new value for `$case`.

### DIFF
--- a/.changeset/honest-bikes-know.md
+++ b/.changeset/honest-bikes-know.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Fixed handling of unknown signal messages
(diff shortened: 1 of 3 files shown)


boks1971 (approved):

cnderrauber (approved):
Thanks for fixing this! A pitfall in typescript for me again :)

davidzhao:
> Thanks for fixing this! A pitfall in typescript for me again :)

not your bug at all! It was mine (as usual 🙃 ).

PR #572, opened 2023-02-03
Rehydrate state after signal reconnection

lukasIO:
builds on top of https://github.com/livekit/protocol/pull/292

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -685,7 +685,6 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     if (this.reconnectAttempts === 0) {
       // only reset start time on the first try
       this.reconnectStart = Date.now();
-      this.emit(EngineEvent.Resuming);
     }
 
     const disconnect = (duration: number) => {
@@ -711,6 +710,9 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
 
     log.debug(`reconnecting in ${delay}ms`);
+    if (this.reconnectAttempts === 0) {
+      this.emit(EngineEvent.Resuming);
+    }
 
     this.clearReconnectTimeout();
     this.reconnectTimeout = CriticalTimers.setTimeout(
@@ -843,7 +845,11 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
 
     log.info(`resuming signal connection, attempt ${this.reconnectAttempts}`);
-    if (emitResuming) {
+    if (
+      emitResuming &&
+      // for the first reconnect attempt the event is being emitted in handleDisconnect
+      this.reconnectAttempts !== 0
+    ) {
       this.emit(EngineEvent.Resuming);
     }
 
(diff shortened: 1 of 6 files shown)


src/room/RTCEngine.ts
```diff
@@ -842,7 +843,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
 
     log.info(`resuming signal connection, attempt ${this.reconnectAttempts}`);
-    if (emitResuming || this.reconnectAttempts === 0) {
+    if (emitResuming) {
```
boks1971, line 846: Will this be emitted twice now? Here and also in the `handleDisconnect` change ☝️ ?

src/room/RTCEngine.ts
```diff
@@ -853,14 +854,15 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
         this.publisher.pc.setConfiguration(rtcConfig);
         this.subscriber.pc.setConfiguration(rtcConfig);
       }
+
+      this.emit(EngineEvent.SignalResumed, res ?? undefined);
```
boks1971, line 858: for my own education: why does TS need the `??` here. If `res` is not there, why can't it just be passed as is?

src/room/Room.ts
```diff
@@ -200,10 +201,13 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
         this.updateSubscriptions();
       })
-      .on(EngineEvent.SignalResumed, () => {
+      .on(EngineEvent.SignalResumed, (reconnectResp) => {
         if (this.state === ConnectionState.Reconnecting) {
           this.sendSyncState();
         }
+        if (reconnectResp) {
+          this.rehydrateStateAfterReconnect(reconnectResp);
+        }
```
boks1971, line 210: Should this before the above `sendSyncState`? Given that things were working before this does not make the change necessary, but feel code will read better if the response is handled first and then action taken (unless handling response does something to the state which will affect operation of `sendSyncState`)

src/room/Room.ts
```diff
@@ -743,20 +747,35 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     }
   };
 
+  private rehydrateStateAfterReconnect(resp: JoinResponse | ReconnectResponse) {
```
boks1971, line 750: as this is handling both `JoinResponse` and `ReconnectResponse`, can this be called just `hydrateState`?

boks1971: And the comments changed to not use `rehydrate`?

boks1971 (approved):
lgtm! but please do not weight my review much as I do not know all the nuances here :-)

src/room/RTCEngine.ts
lukasIO: it's just a type conflict that res can be either `ReconnectResponse` or `void` but TS complains that `void` and `undefined` are not compatible as `void` could theoretically be any value, not just undefined (https://www.typescriptlang.org/docs/handbook/2/functions.html#return-type-void). 
Agree however that it looks weird, probably changing the return type of connect to `ReconnectResponse | undefined` is the right move.

src/room/Room.ts
lukasIO: even though it can take a JoinResponse, the function is only called either after resuming or a full reconnect. On a full reconnect we would still get a `JoinResponse` which is the confusing part.
Due to the limited scope I thought it makes sense to keep it explicit.

src/room/RTCEngine.ts
lukasIO: yeah, that's theoretically possible, but couldn't find a path that would trigger it twice. will address.

cnderrauber (approved):

PR #573, opened 2023-02-03
Add reconnect reason and signal rtt calculation

cnderrauber:


src/room/RTCEngine.ts
```diff
@@ -1001,7 +1022,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
       const negotiationTimeout = setTimeout(() => {
         reject('negotiation timed out');
-        this.handleDisconnect('negotiation');
+        this.handleDisconnect('negotiation', false, ReconnectReason.REASON_SIGNAL_DISCONNECTED);
```
lukasIO, line 1025: just double checking, is this the correct reason for a negotiation time out? 
doesn't seem like this has anything to do with the signal connection at this stage?

src/room/RTCEngine.ts
```diff
@@ -766,7 +784,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       }
 
       if (recoverable) {
-        this.handleDisconnect('reconnect', requireSignalEvents);
+        this.handleDisconnect('reconnect', requireSignalEvents, ReconnectReason.REASON_UNKOWN);
```
lukasIO, line 787: would it be useful (from the server's perspective) to know with what error/exception the reconnect attempt failed?

lukasIO (approved):

src/room/RTCEngine.ts
cnderrauber: When transport get answer from server and call `setRemoteDescription`, it will fire `NegotiationComplete` and clear the timeout timer, so if the timeout timer is fired, it means no answer arrived.

src/room/RTCEngine.ts
cnderrauber: Currently it don't care the detail of error/exception.

PR #578, opened 2023-02-07
Reset ping when receiving pong instead of clearing

lukasIO:
sorry @cnderrauber, missed this in the review!


cnderrauber (approved):
Thanks!

PR #487, opened 2022-10-24
Fix `localTrack.replaceTrack` mute handling

lukasIO:
syncs muted state with the enabled state of the newly provided track when calling `replaceTrack`

fixes #486

PR #579, opened 2023-02-07
Fix `LocalTrack.restart` mute handling

HermanBilous:
This PR is quite similar to #487, which originally was intended to fix the issue current PR is addressing: having `LocalTrack.isMuted` set to `true`, while having `MediaTrack.enabled` set to `true` as well.

### DIFF
--- a/.changeset/sour-falcons-shop.md
+++ b/.changeset/sour-falcons-shop.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Sync muted state with `MediaStreamTrack.enabled` when calling `LocalTrack.restart`
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -193,6 +193,9 @@ export default abstract class LocalTrack extends Track {
 
     this._mediaStreamTrack = newTrack;
 
+    // sync muted state with the enabled state of the newly created stream
+    this._mediaStreamTrack.enabled = !this.isMuted;
+
     await this.resumeUpstream();
 
     this.attachedElements.forEach((el) => {


lukasIO:
hi @HermanBilous 
thanks for the PR! 
The current assumption is that restarting a track would always put the track in an unmuted state.
Can you tell a bit more about your use case for this change?

HermanBilous:
hi @lukasIO, thanks for taking a look. My case is the following: I have a react app built with react-core and livekit-client. A user connects to a room, we publish an audio track and mute it right away (to make consecutive unmutes quicker). We also display a list of actively speaking participants on the screen, relying on isSpeaking property from `Participant` in react-core.

The issue is, sometimes a user shows up as actively speaking even though they didn't activate their mic. I'm not sure what is the "natural" way to reproduce this issue. However, denying and then granting browser's (chrome 109.0.5414.119 on macOS) permission to a mic on the page produces the same result, you will appear as speaking to other users despite not actually enabling your mic.

HermanBilous:
I haven't changed dtx value, so it's enabled as by default.
I'm doing the following:
```
room?.localParticipant.setMicrophoneEnabled(true).then((track) => track?.mute());
```

With the change from this PR, the app behaves as expected.
Do you think it's correct that the `mediaStreamTrack.enabled` and  `LocalTrack.isMuted` are out of sync?

lukasIO:
> Do you think it's correct that the mediaStreamTrack.enabled and LocalTrack.isMuted are out of sync?
No, they shouldn't be out of sync.
But muting the track shouldn't even be calling `restartTrack`, so I'm wondering what else might be at play here.

Are you able to provide a minimal reproduction that showcases this incorrect behaviour?

HermanBilous:
Sure, here's a sandbox.
https://codesandbox.io/s/infallible-pateu-dpje5x?file=/src/App.tsx

lukasIO:
thanks for the sandbox! that helped isolate the issue. 
The problem I found is that `restartTrack` would be triggered on permission change even if the track is not muted.
Can you check if https://github.com/livekit/client-sdk-js/pull/581 fixes the problem for you, too?

PR #580, opened 2023-02-08
Switch disconnect to the unload event instead of beforeunload

wcarle:
Because the room disconnect is called on the beforeunload event users aren't able to add their own beforeunload confirmation warning. This is because the user will be disconnected from the room even if they decide to stay on the page.

For example, when a recording is in progress in our application we want to warn a user before they leave to remind them to stop the recording.

This switches the room disconnect to the unload event so in our client code we are able to add the confirm navigation warning.

As far as I know there shouldn't be an issue with the disconnect happening on unload instead of beforeunload, but let me know.

### DIFF
--- a/.changeset/spotty-horses-shave.md
+++ b/.changeset/spotty-horses-shave.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Switch disconnect to the unload event instead of beforeunload
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -392,7 +392,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
         this.abortController?.signal.removeEventListener('abort', abortHandler);
         // also hook unload event
         if (isWeb()) {
-          window.addEventListener('beforeunload', this.onBeforeUnload);
+          window.addEventListener('unload', this.onUnload);
           navigator.mediaDevices?.addEventListener('devicechange', this.handleDeviceChange);
         }
         this.setAndEmitConnectionState(ConnectionState.Connected);
@@ -542,7 +542,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     }
   }
 
-  private onBeforeUnload = async () => {
+  private onUnload = async () => {
     await this.disconnect();
   };
 
@@ -837,7 +837,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       this.audioContext = undefined;
     }
     if (isWeb()) {
-      window.removeEventListener('beforeunload', this.onBeforeUnload);
+      window.removeEventListener('unload', this.onUnload);
       navigator.mediaDevices?.removeEventListener('devicechange', this.handleDeviceChange);
     }
     this.setAndEmitConnectionState(ConnectionState.Disconnected);


lukasIO:
hi @wcarle,
that's a good point. 
It seems generally recommended not to use the `unload` event at all anymore. 
`pagehide` seems to be preferable. (https://developer.chrome.com/blog/page-lifecycle-api/#legacy-lifecycle-apis-to-avoid)
I haven't done any tests yet to determine whether `pagehide` fires the disconnect rather reliably, but am happy to take a closer look if you could confirm that switching to `pagehide` would mitigate the issue you're having?

wcarle:
@lukasIO reading a bit further in that doc to the section on beforeunload, it actually has the same limitations as unload:
https://developer.chrome.com/blog/page-lifecycle-api/#the-beforeunload-event

So technically neither of them are good as end of session signals in the same way:

> The beforeunload event has a similar problem to the unload event, in that when present it prevents browsers from caching the page in their [back/forward cache](https://developer.chrome.com/blog/page-lifecycle-api/#back-forward-cache).
> 
> The difference between beforeunload and unload, though, is that there are legitimate uses of beforeunload. For instance, when you want to warn the user that they have unsaved changes they'll lose if they continue unloading the page.
> 
> Since there are valid reasons to use beforeunload but using it prevents pages from being added to the back/forward cache, it's recommended that you only add beforeunload listeners when a user has unsaved changes and then remove them immediately after the unsaved changes are saved.

So if I'm reading that correctly, both events have the same downsides, the only reason they say beforeunload is ok is for the specific case of the unsaved changes warning

davidzhao:
I really should've read the docs on `beforeunload` prior to using it in the SDK. It seems `pagehide` is the better path forward. @wcarle I believe `beforeunload` is still fired before `pagehide`. So in that case it should satisfy your needs?

Ocupe:
If this [article](https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/) is still accurate, then there is a possibility that none of the events under discussion will work reliably for mobile browsers.

> You cannot rely on `pagehide`, `beforeunload`, and `unload` events to fire on mobile platforms. This is not a bug in your favorite browser; this is due to how all mobile operating systems work. An active application can transition into a "background state" via several routes:
> - The user can click on a notification and switch to a different app.
> - The user can invoke the task switcher and move to a different app.
> - The user can hit the "home" button and go to homescreen.
> - The OS can switch the app on users behalf - e.g. due to an incoming call.

On mobile browsers `visibilitychange` it the last reliably fired event:
`addEventListener("visibilitychange", (event) => {});`

[This claim is supported by MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event):

> Transitioning to hidden is the last event that's reliably observable by the page, so developers should treat it as the likely end of the user's session (for example, for [sending analytics data](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon)).

On desktop, switching tabs should obviously not disconnect the participant from the LiveKit-session. So maybe we need to implement different session-ended-detection-logics for mobile and desktop?

lukasIO:
@wcarle are you interested in changing this PR to reflect @davidzhao 's suggestions?

If not, I'd start working on this in a separate PR.

lukasIO:
superseded by https://github.com/livekit/client-sdk-js/pull/627
thanks @wcarle for the report!

PR #581, opened 2023-02-09
Only restart track after permission change if not muted

lukasIO:
thanks to @HermanBilous for surfacing this issue!

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -999,7 +999,9 @@ export default class LocalParticipant extends Participant {
               // detect granted change after permissions were denied to try and resume then
               currentPermissions.onchange = () => {
                 if (currentPermissions.state !== 'denied') {
-                  track.restartTrack();
+                  if (!track.isMuted) {
+                    track.restartTrack();
+                  }
                   currentPermissions.onchange = null;
                 }
               };
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -1011,8 +1011,10 @@ export default class LocalParticipant extends Participant {
             // permissions query fails for firefox, we continue and try to restart the track
           }
         }
-        log.debug('track ended, attempting to use a different device');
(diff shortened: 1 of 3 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -999,7 +999,9 @@ export default class LocalParticipant extends Participant {
               // detect granted change after permissions were denied to try and resume then
               currentPermissions.onchange = () => {
                 if (currentPermissions.state !== 'denied') {
-                  track.restartTrack();
+                  if (!track.isMuted) {
```
davidzhao, line 1002: what is the correct behavior if it's muted after permission change? do we need to update our state internally? (i.e. replace uptrack with placeholder?)

related, when the user unmutes, what is the behavior there?

lukasIO: I don't think we need to replace the track at that point, we can keep the ended mediastreamtrack as the muted one.
For video the track get's restarted anyways when `unmute` is being called. 
For audio this also happens, as long as `stopOnMute` is set to true. 
Not sure if we additionally want to handle audio tracks that are not being stopped, i.e. where we `unmute` does not trigger a restart. 
Acquiring permissions/restarting the track while a user's track is muted sounds unexpected imo

davidzhao: got it. so to focus on the case with audio when `stopOnMute` is set to false.

the sequence will be like this
1. user mutes
2. permission went to denied and then granted
3. we did not restart the track here
4. when the user unmutes, would that media track still be valid?

lukasIO: No, the mediatrack would not be valid any more. 
What do you think about a check for `mediastreamtrack.ended` within the `unmute` function of the audio track? 
If the track's status is `ended` we could then trigger `restartTrack`

davidzhao: ah that makes sense.. sgtm!

davidzhao (approved):

PR #585, opened 2023-02-28
Prevent concurrent setParameter call to avoid exception

davidzhao:
if multiple get/setParameter are called concurrently, certain timing of events could lead to the browser throwing an exception in `setParameter`, due to a missing `getParameter` call.


lukasIO (approved):
lgtm!

Issue #590, opened 2023-03-01
enable 'selfBrowserSurface' in screen share


ShubhamJuIT:
### Describe the problem

When doing Share Screen, the tab on which we are currently not visible in screen sharing options in chrome, because chrome disabled it by default. 

This issue is only happening in chrome only. Chrome disabled it bydefault to disable mirro imaging issue.

### Describe the proposed solution

Proposed solution is to pass 'selfBrowserSurface': 'include' when calling getMediaDevice()

 

### Importance

nice to have

davidzhao:
I don't think the right solution is to override a default that Chrome decided to set. But I do agree it could make sense as an option for those that need this

ShubhamJuIT:
@davidzhao Yes. For my requirement it is important. If you add this, it would be immensely helpful, to reduce some code duplicity and compatibility as well.

ShubhamJuIT:
Thanks and Cheers to whole livekit team. 😄

PR #592, opened 2023-03-02
Add support for experimental getDisplayMedia options

lukasIO:
fixes https://github.com/livekit/client-sdk-js/issues/590

and additionally adds all the other new experimental options that chrome introduced


davidzhao (approved):
lgtm!

PR #593, opened 2023-03-03
Disable red encoding by default for stereo track

cnderrauber:
As the stereo audio track has high bitrates(128kbps or more), we don't want to consume 3x bitrates for it by enabling red encoding by default because that would cause higher packet loss if the network is poor, so disable it. The user can enable it explicitly if required.

### DIFF
--- a/src/room/track/options.ts
+++ b/src/room/track/options.ts
@@ -28,8 +28,7 @@ export interface TrackPublishDefaults {
   audioBitrate?: number;
 
   /**
-   * dtx (Discontinuous Transmission of audio), defaults to true.
-   * If the audio track is stereo, need enable it explicitly.
+   * dtx (Discontinuous Transmission of audio), enabled by default for mono tracks.
    */
   dtx?: boolean;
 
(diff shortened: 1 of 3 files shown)


src/room/track/options.ts
```diff
@@ -28,12 +28,14 @@ export interface TrackPublishDefaults {
   audioBitrate?: number;
 
   /**
-   * dtx (Discontinuous Transmission of audio), defaults to true
+   * dtx (Discontinuous Transmission of audio), defaults to true.
+   * If the audio track is stereo, need enable it explicitly.
```
davidzhao, line 32:
```suggestion
   * dtx (Discontinuous Transmission of audio), enabled by default for mono tracks.
```

src/room/track/options.ts
```diff
@@ -28,12 +28,14 @@ export interface TrackPublishDefaults {
...
    */
   dtx?: boolean;
 
   /**
    * red (Redundant Audio Data), defaults to true
+   * If the audio track is stereo, need enable it explicitly.
```
davidzhao, line 38:
```suggestion
   * red (Redundant Audio Data), enabled by default for mono tracks.
```

davidzhao (approved):

PR #595, opened 2023-03-03
Fixed incorrect state with resume then reconnect

davidzhao:
When a resume sequence fails, a full reconnect is attempted. There were a few issues there with that sequence
- We did not always fire EngineEvent.Restarting, so Room missed tearing down existing participants
- With selective subscriptions, when existing `isDesired` isn't cleared, it will not send a subscribe request when requested
- New tracks were not republished successfully when reconnected (due to sender not being reset early enough)

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -187,7 +187,6 @@ export default abstract class LocalTrack extends Track {
     log.debug('re-acquired MediaStreamTrack');
 
     if (this.sender) {
-      log.debug('replacing track');
       // Track can be restarted after it's unpublished
       await this.sender.replaceTrack(newTrack);
     }
@@ -203,7 +202,6 @@ export default abstract class LocalTrack extends Track {
     this.mediaStream = mediaStream;
     this.constraints = constraints;
     this.emit(TrackEvent.Restarted, this);
-    log.debug('restarted track');
     return this;
   }
 


src/room/RTCEngine.ts
```diff
@@ -731,12 +735,12 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
     this.clearReconnectTimeout();
     this.reconnectTimeout = CriticalTimers.setTimeout(
-      () => this.attemptReconnect(signalEvents, disconnectReason),
+      () => this.attemptReconnect(disconnectReason),
       delay,
     );
   };
 
-  private async attemptReconnect(signalEvents: boolean = false, reason?: ReconnectReason) {
+  private async attemptReconnect(reason?: ReconnectReason) {
```
davidzhao, line 743: The logic for `signalEvents` was really difficult to follow. The reason why it was there is so that Room doesn't have to fire `Reconnecting` multiple times.

It makes more sense to enforce at the Room layer, instead of here.

RTCEngine always fires Resuming or Restarting events, so that we do not miss the opportunity to unwind

src/room/RTCEngine.ts
```diff
@@ -883,6 +879,11 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
     this.emit(EngineEvent.SignalResumed);
 
+    if (this.shouldFailNext) {
```
davidzhao, line 882: this is for testing with `simulateScenario('resume-reconnect')`

src/room/participant/LocalParticipant.ts
```diff
@@ -721,31 +721,34 @@ export default class LocalParticipant extends Participant {
...
+              }
               trackInfo.sender = undefined;
             }
           }
           track.simulcastCodecs.clear();
         }
       } catch (e) {
         log.warn('failed to unpublish track', { error: e, method: 'unpublishTrack' });
-      } finally {
-        await this.engine.negotiate();
```
davidzhao, line 743: this await caused `track.sender` to not get unset in time. instead other logic ran and threw an exception using the wrong track sender.

src/room/participant/RemoteParticipant.ts
```diff
@@ -246,7 +245,7 @@ export default class RemoteParticipant extends Participant {
           (publishedTrack) => publishedTrack.source === publication?.source,
         );
         if (existingTrackOfSource && publication.source !== Track.Source.Unknown) {
-          log.warn(
+          log.debug(
```
davidzhao, line 248: it is possible to use tracks of multiple Source, so we'll stop sounding the alarm bells about it.

davidzhao:
Sorry for tagging y'all. since this is a pretty tricky case and important to get right (resume and reconnect attempts), wanted to make sure we get this right here as well as across other SDKs.

src/room/RTCEngine.ts
lukasIO: as of now the room layer check to only emit `Reconnecting` once is still missing, right?

lukasIO (approved):
great findings!

theomonnom (approved):
Looks good, I'll check that for the Rust SDK 👌
Is it the issue I had at our last meeting?

davidzhao:
This PR actually had a regression as well. addressed in a follow up #600

PR #600, opened 2023-03-03
Always fire EngineEvent.Resuming

davidzhao:
Revert previous change in behavior (and premature optimization) of skipping resuming event. Doing so breaks migration


theomonnom (approved):
lgtm :)

davidzhao:
@theomonnom I've added another commit to this to address events not being emitted at the right time. merging this in now.

PR #609, opened 2023-03-07
Reject publish future if engine disconnects

lukasIO:
forgot to handle the case when engine disconnects during a publish attempt.


davidzhao (approved):

src/room/participant/LocalParticipant.ts
```diff
@@ -142,6 +143,10 @@ export default class LocalParticipant extends Participant {
     this.updateTrackSubscriptionPermissions();
   };
 
+  private handleDisconnected = () => {
+    this.reconnectFuture?.reject?.('Got disconnected during publishing attempt');
```
davidzhao, line 147: should we set this to undefined as well?

lukasIO: good catch! yes!

PR #617, opened 2023-03-13
option to enable/disable adaptiveStream for an element

jibon57:
By default adaptiveStream can be enable/disable using global variable. In certain case may need to disable/enable adaptive steam for single tracks. For example one video element need to put in PiP but because of adaptiveStream if user move to another tab or window then the stream will be pause. To prevent this I had made `stopObservingElement` public & another new function `startObservingElement` for triggering the element manually.

### DIFF
--- a/.changeset/silent-zoos-run.md
+++ b/.changeset/silent-zoos-run.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+option to enable/disable adaptiveStream for an element
--- a/src/room/track/RemoteVideoTrack.ts
+++ b/src/room/track/RemoteVideoTrack.ts
@@ -195,7 +195,25 @@ export default class RemoteVideoTrack extends RemoteTrack {
     return receiverStats;
   }
 
-  private stopObservingElement(element: HTMLMediaElement) {
+  /**
+   * Start observing an Element for changes.
+   * @param element
+   */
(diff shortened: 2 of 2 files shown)


jibon57:
Yes, that will be fine. But I think have option to manually trigger to stop/start observation will be also good idea. In certain case it will be helpful. Also the changed I have done won't affect any other code of this library.

jibon57:
@lukasIO any thoughts on this? I can't use pip.

lukasIO:
Hi jibon, 
sorry for the late reply! Very happy to add the PiP event listeners to make adaptive stream aware of it, but I'm not entirely convinced that exposing the methods in this PR is the right move. 

It will increase API surface for a very narrow use case, plus a bunch of testing would be needed to make sure that every call sequence would still behave as expected. On top of that we would arguably have to introduce additional state that makes `adaptiveStream` a track based setting (because that's what watching/unwatching does effectively) while currently it's a room level setting.

lukasIO:
@jibon57 opened https://github.com/livekit/client-sdk-js/pull/622 to address this and also noticed that you should probably work around the issue completely when setting `{ adaptiveStream: { pauseVideoInBackground: false } }` on the room options

jibon57:
@lukasIO having option to disable adaptive Stream for any video track would be helpful for certain cases not just for pip. It's fine for me as I can do customization for my application.

PR #618, opened 2023-03-13
Specify tsconfig rootDir

max-b:
From [the docs](https://www.typescriptlang.org/tsconfig#rootDir):
> Default: The longest common path of all non-declaration input files. If [composite](https://www.typescriptlang.org/tsconfig#composite) is set, the default is instead the directory containing the tsconfig.json file.

In this repo's case, the "the longest common path of all non-declaration input files" happens to be `./` so the correct behavior is occuring. However, when attempting to build this in alternative ways (we're linking it into a `node_modules` folder so that we can do local development), that changes and the build is no longer predictable.

This change just makes explicit what the expected/desired behavior is.


lukasIO (approved):
thanks, that makes sense!

PR #622, opened 2023-03-21
Don't pause videos when element is in pictureInPicture mode

lukasIO:


src/room/track/RemoteVideoTrack.ts
```diff
@@ -284,17 +287,29 @@ export interface ElementInfo {
 class HTMLElementInfo implements ElementInfo {
   element: HTMLMediaElement;
 
-  visible: boolean;
+  get visible(): boolean {
+    console.log({ pip: this.isPiP, intersect: this.isIntersecting });
```
davidzhao, line 291: remove debug console log?

src/room/track/RemoteVideoTrack.ts
```diff
@@ -284,17 +287,29 @@ export interface ElementInfo {
...
 
+  private isPiP: boolean;
+
+  private isIntersecting: boolean;
+
   constructor(element: HTMLMediaElement, visible?: boolean) {
     this.element = element;
-    this.visible = visible ?? isElementInViewport(element);
+    this.isIntersecting = visible ?? isElementInViewport(element);
+    this.isPiP = false; // we cannot be sure, but default to false as a first guess
```
davidzhao, line 312: what happens if the element is already in pip when we are binding to it? would we ever know that it's in PIP?

davidzhao (approved):

davidzhao (approved):
awesome!

PR #623, opened 2023-03-21
Emit `RoomEvent.LocalAudioSilenceDetected` if a `LocalAudioTrack` is comp…

### LINKED ISSUE #584: Detect when a user's local microphone isn't working and fire event
### Describe the problem

Sometimes the user's microphone may be malfunctioning at the hardware level so it's unable to produce audio.

This can be observed in a few different ways:
* No audio packets transmitted for the track (visible from webrtc stats)
* Server did not receive the track (SFU requires packets to fully register the track)

I think the right way of handling it is to fire an event so that the application could render UI that notifies the end user. We've seen that the problem can be resolved by resetting the hardware (such as unplugging and replugging a USB mic)

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

lukasIO:
…letely silent after publishing

closes #584

### DIFF
--- a/.changeset/nasty-apes-agree.md
+++ b/.changeset/nasty-apes-agree.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Emit `RoomEvent.LocalAudioSilenceDetected` if a `LocalAudioTrack` is silent after publishing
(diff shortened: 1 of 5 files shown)


davidzhao (approved):
lgtm

Issue #589, opened 2023-03-01
No audio on Safari when someone joined and the browser tab is inactive


laszlominczer:
### Describe the bug

When I joined a room with disabled microphone, and after changing the browser tab on Safari, I didn't hear the newly joined participants audio tracks. But when I changed back to the active room's tab the audio is working again.

### Reproduction

Open the sample app(latest version) in Safari
Join  a room 
Disable the audio
Switch to another browser tab in Safari
Join with another participant in another browser or a different tab
Say something into the microphone

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 49.69 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.2.0 - ~/.nvm/versions/node/v18.2.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.12.2 - ~/.nvm/versions/node/v18.2.0/bin/npm
  Browsers:
    Chrome: 110.0.5481.177
    Firefox: 109.0
    Safari: 16.2
```

### Severity

annoyance

lukasIO:
This unfortunately seems to be safari specifically blocking auto play on elements in backgrounded tabs. 
Have you observed any other audio/video solution that has made this work?

davidzhao:
Safari strikes again :(

I guess we could pre-warm a bunch of audio elements?

laszlominczer:
Thanks for your responses, @lukasIO I tried Zoom and Google Meets also, and those were working well.

davidzhao:
This should be fixed by #733 @lukasIO ?

lukasIO:
Not sure, if this addresses the same issue. #733 is solving a (different) auto-play issue specifically for iOS Safari, while this issue affects primarily macOS due to tab switching.

svajunas-budrys:
Still have this problem, are there any plans to deal with it in the near future?

svajunas-budrys:
This seem to work. Thanks a lot!

PR #624, opened 2023-03-21
Prewarm an audio element pool when connecting

lukasIO:
This PR
- creates an audio element pool
- tries to play silent audio when connecting and triggers `audioPlaybackFailed` if that doesn't work
- adds playback of silent pool audio nodes to `startAudio`

potentially fixes https://github.com/livekit/client-sdk-js/issues/589, but there are some caveats: 
- user still needs to click `startAudio` when first connecting (also this will fire a lot more now)
- only works if users do not supply their own HTML element to the `attach` function, because we can only use prewarmed elements if we can choose which element it can be

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   },
   "dependencies": {
     "async-await-queue": "^1.2.1",
+    "create-silent-audio": "^0.1.2",
     "events": "^3.3.0",
     "loglevel": "^1.8.0",
     "protobufjs": "^7.0.0",
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -59,6 +59,7 @@ import {
   createDummyVideoStreamTrack,
   Future,
   getEmptyAudioStreamTrack,
+  isSafari,
   isWeb,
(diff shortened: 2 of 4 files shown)


davidzhao (commented):
sorry just getting around to this.. re: startAudio, is that due to us creating a bunch more audio elements ahead of time?

I'm wondering if this extra piece of code is worth only a partial fix to the issue. Would using WebAudio help solve the case of attaching to existing audio elements too? Perhaps that'd be a better path to go down.

lukasIO:
> is that due to us creating a bunch more audio elements ahead of time
yeah, the initiated elements still need to be "unlocked" with a user gesture at one point. 

> I'm wondering if this extra piece of code is worth only a partial fix to the issue. 
I have the same reservation about this change. 

> Would using WebAudio help solve the case of attaching to existing audio elements too?
currently we're still adding actual audio elements (albeit muted) for attach calls in the webaudio path. I guess it would help as muted audio elements might not get blocked, but I'll have to do some tests to verify this

PR #625, opened 2023-03-21
fix: emit mute status event for unsubscribed user

wjaykim:
Hi!

This PR changes some behavior about muted status event.

Before this change, event about muted status was not emitted if the track publication is not subscribed, despite of its internal muted status is changed.

I don't know if it is livekit's intended behavior. But in my use case I want to know the change of muted status even when user didn't subscribed the track.

This change will affect these events:

- `ParticipantEvent.Muted`
- `ParticipantEvent.Muted`
- `TrackEvent.Muted`
- `TrackEvent.Muted`

### DIFF
--- a/src/room/track/RemoteTrackPublication.ts
+++ b/src/room/track/RemoteTrackPublication.ts
@@ -208,7 +208,11 @@ export default class RemoteTrackPublication extends TrackPublication {
   updateInfo(info: TrackInfo) {
     super.updateInfo(info);
     this.metadataMuted = info.muted;
-    this.track?.setMuted(info.muted);
+    if (this.track) {
+      this.track.setMuted(info.muted);
+    } else {
+      this.emit(info.muted ? TrackEvent.Muted : TrackEvent.Unmuted);
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/room/track/RemoteTrackPublication.ts
```diff
@@ -207,8 +207,16 @@ export default class RemoteTrackPublication extends TrackPublication {
   /** @internal */
   updateInfo(info: TrackInfo) {
     super.updateInfo(info);
+    if (this.metadataMuted === info.muted) {
```
lukasIO, line 210: I think it would be great to move this check into the `else` clause further down for the following reason:
while it's not intended that way, it's possible that users use `setMuted` directly on the track in which case metadataMuted and the actual mute status of the track might get out of sync. by not altering the path before `track.setMuted` is called we would make sure that a remote update would still process the track correctly in all cases.

wjaykim: That makes sense! I'll make the change.

lukasIO (approved):
Thanks for the PR, this makes a lot of sense!

PR #629, opened 2023-03-23
React-native client info

davidliu:


davidliu:
@lukasIO Changed to using `navigator.product`, but no luck on the `navigator.platform` for detecting the OS. :(

src/room/utils.ts
```diff
@@ -122,6 +122,22 @@ export function isWeb(): boolean {
...
 
+export function isReactNative(): boolean {
+  // navigator.product is deprecated on browsers, but will be set appropriately for react-native.
+  return navigator.product == 'ReactNative';
+}
+
+export function getReactNativeOs(): string | undefined {
+  let reactNative;
+  try {
+    reactNative = require('react-native');
```
lukasIO, line 133: still a bit wary about having this `require` in here. Not sure if there are any toolchains that might get stuck on this when being imported as a plain JS lib. 
I know it's not a nice solution either, but could we pollute global scope with a flag that indicates this on the RN side, instead of explicitly `require`ing here?

davidzhao: maybe wrap that require with `isReactNative` ?

lukasIO: I fear that wouldn't help in cases where the build tool statically analyses all `require` statements in the code.
related: https://github.com/facebook/react-native/issues/1331

davidliu: Switched over to using the global scope to pass the platform info.

lukasIO (approved):
nice! 
and thank you 😉

PR #633, opened 2023-03-29
Give up on reconnections if token is invalid

lukasIO:
Reconnect attempts will now fail immediately in case the reason for reconnection failing was the user being unauthorized to join

### DIFF
--- a/.changeset/plenty-cherries-invite.md
+++ b/.changeset/plenty-cherries-invite.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Give up on reconnections if token is invalid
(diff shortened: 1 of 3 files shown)


davidzhao (commented):
I think this change itself looks good. Though double checking when we'd consider it NotAllowed, it seems the current logic has a bug:

```
            const resp = await fetch(`http${url.substring(2)}/validate${params}`);
            if (!resp.ok) {
              const msg = await resp.text();
              reject(new ConnectionError(msg, ConnectionErrorReason.NotAllowed, resp.status));
```

it assumes any validate errors to be NotAllowed, including if it's a 500. I think we'd need to tighten that and compare if response code is 4xx (i.e. user error)

lukasIO:
very good point! I had missed that.

davidzhao (approved):
lgtm!

PR #634, opened 2023-03-30
Allow manual operation on audio publications even with adaptiveStream enabled

lukasIO:
Previously we were restricting the `setEnabled` API also for audio publications in case `adaptiveStream` is enabled.
`adaptiveStream` only affects video publications, however.

### DIFF
--- a/.changeset/polite-eels-swim.md
+++ b/.changeset/polite-eels-swim.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Allow manual operation on audio publications with adaptiveStream enabled


src/room/track/RemoteTrackPublication.ts
```diff
@@ -238,8 +238,8 @@ export default class RemoteTrackPublication extends TrackPublication {
   }
 
   private isManualOperationAllowed(): boolean {
-    if (this.isAdaptiveStream) {
-      log.warn('adaptive stream is enabled, cannot change track settings', {
+    if (this.kind === Track.Kind.Video && this.isAdaptiveStream) {
+      log.warn('adaptive stream is enabled, cannot change video track settings', {
         trackSid: this.trackSid,
       });
       return false;
```
boks1971, line 245: On a different note, should there an `adpativeStream` mode which includes audio also? I guess it can get confusing. I was thinking a participant level `adaptiveStream`. If any of the video tile of a participant is visible (camera or screen share), audio is also enabled. If none are visible, audio also disabled in this mode?

lukasIO: I think this might make sense for a narrow set of use cases, but also wondering if it's worth the mental overhead to reason about different adaptiveStream modes to support it. 
FWIW, the adaptiveStream setting already can be passed in as a config object, so it's not too far fetched to think about `adaptiveStream: { includeAudio: true }` or something along those lines

boks1971: Yeah, it will increase cognitive load. Don't think it is worth it.

PR #642, opened 2023-04-06
Region URL Provider

lukasIO:


src/room/RTCEngine.ts
```diff
@@ -47,6 +47,7 @@ import {
   supportsSetCodecPreferences,
   supportsTransceiver,
 } from './utils';
+import { RegionUrlProvider, isCloud } from './track/RegionUrlProvider';
```
davidzhao, line 50: nit: should this file belong in `track/`?

src/room/Room.ts
```diff
@@ -286,71 +286,16 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       }
 
       try {
-        const joinResponse = await this.engine.join(
+        const joinResponse = await this.connectSignal(
```
davidzhao, line 289: maybe the initial connect function can be wrapped in a helper as well? so that we'd try backup urls a few times (including if peerconnection fails)

src/room/track/RegionUrlProvider.ts
```diff
@@ -0,0 +1,86 @@
...
+function getCloudConfigUrl(serverUrl: URL) {
+  const urlParts = serverUrl.hostname.split('.');
+  const subdomain = urlParts[0];
+  const isStaging = urlParts[1] === 'staging';
+
+  // TODO REMOVE DEBUG
+  return `localhost:7880/region`;
+
+  if (isStaging) {
+    return `https://${subdomain}.config.staging.livekit.cloud/region`;
```
davidzhao, line 82: we can probably get away with not hardcoding any URLs here, instead of splitting we could find the first substring before the `.` and insert `.config`? in case other environments are added in the future

davidzhao (commented):
looking good!

src/room/Room.ts
```diff
@@ -258,155 +259,210 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
+        resultingError.reason = err.reason;
+        resultingError.status = err.status;
+      }
+      log.debug(`error trying to establish signal connection`, { error: err });
+      throw resultingError;
+    }
 
-    return this.connectFuture.promise;
+    try {
+      await this.engine.waitForPCInitialConnection(this.connOptions.peerConnectionTimeout);
```
davidzhao, line 450: should abortController be passed in here?

davidzhao (approved):
lg! the connect stack seems much cleaner. though it is a significant change. I just wanted to make sure our tests already cover this.

davidzhao (approved):

PR #643, opened 2023-04-07
Add VP9 SVC support

cnderrauber:


src/room/PCTransport.ts
```diff
@@ -142,6 +144,9 @@ export default class PCTransport extends EventEmitter {
             if (
               !media.fmtp.some((fmtp): boolean => {
                 if (fmtp.payload === codecPayload) {
+                  if (!fmtp.config.includes('x-google-start-bitrate')) {
```
cnderrauber, line 147: Use 0.7 target bitrates as start bitrate, in the test case with 720p AV1, the publisher will start from 540p, then go to 720p about 3~5 seconds if network is good

davidzhao: this seems like it'll impact other codecs too?

src/room/utils.ts
```diff
@@ -40,25 +42,33 @@ export function supportsDynacast() {
...
       }
     }
-    for (const ext of capabilities.headerExtensions) {
-      if (
-        ext.uri ===
-        'https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension'
-      ) {
-        hasDDExt = true;
+  }
+  return hasAV1;
```
davidzhao, line 53: is this true on Safari as well? is it able to publish AV1 in SVC mode? If not, it'd be good to add a browser type check here.

davidzhao (commented):
one remaining question:

how should we adjust default bitrate when it comes to VP9? AFAIK, it's more efficient than vp8.

src/room/PCTransport.ts
cnderrauber: No, it will only work for svc codecs.

src/room/utils.ts
cnderrauber: No, Safari doesn't have av1 in its codec capabilities. By the way, munging sdp to enable Dependency Descriptor works well for vp9 svc in Safari too.

cnderrauber:
> one remaining question:
> 
> how should we adjust default bitrate when it comes to VP9? AFAIK, it's more efficient than vp8.

For camera usage, have tested 0.7x vp8 bitrates and it looks well, so using same bitrates as av1 now.

src/room/participant/publishUtils.ts
```diff
@@ -248,6 +249,8 @@ export function determineAppropriateEncoding(
   if (codec) {
     switch (codec) {
       case 'av1':
+      case 'vp9':
+        encoding = { ...encoding };
         encoding.maxBitrate = encoding.maxBitrate * 0.7;
```
davidzhao, line 254: I think this is too aggressive for VP9. What I've read is av1 is about 25% more efficient compared to vp9.

Perhaps use 0.85 or 0.9 for vp9?

cnderrauber: Does the comparison result of what you have read is real-time case? In my test the bitrates would vary a lot with different encode parameters for different scenarios (real-time/file). I don't have a definite result too, maybe 0.85 would be better that 0.7.

lukasIO: side note: not sure how expected it is in general if users set a custom `videoEncoding` that we alter the `maxBitrate` value based on the codec here. 
If a user consciously chooses to publish av1 or vp9 with a desired `maxBitrate` why would we opt to adjust that beyond what the user has set?

cnderrauber:
@davidzhao @lukasIO Going to merge this PR so that we can enable vp9/av1 svc now, any concern about it?

src/room/participant/publishUtils.ts
lukasIO: ah, missed that, thank you!

lukasIO (approved):

anishmenon:
on client side

`publishDefaults: {
          videoEncoding: {
            maxBitrate: 0.9,
            maxFramerate: 60,
          },
          screenShareEncoding: {
            maxBitrate: 0.9,
            maxFramerate: 30,
          },
          audioBitrate: 20000,
          dtx: true,
          stopMicTrackOnMute: true,
         
          videoCodec:"av1",
          backupCodec: { codec: 'vp9' },          // only needed if overriding defaults
        },`


on server  config


`room:
      enabled_codecs:
        - mime: audio/opus
        - mime: video/vp9
        - mime: video/av1`

but getting error from all VP9 supported browsers


Uncaught (in promise) DOMException: Failed to execute 'setRemoteDescription' on 'RTCPeerConnection': Failed to set remote offer sdp: Failed to set remote video description send parameters for m-section with mid='2'.

cnderrauber:
Which version of lk-server are you using? You can try v1.4.2 to test it again, and post your offer & answer sdp if it still can't work.

Issue #647, opened 2023-04-11
Memory Leaks Caused in Subscription / Unsubscription Process


StasCosmos:
### Describe the bug

If we subscribe and then unsubscribe from RemoteTrack, we have listeners leak.

### What's Happening

During Livekit `this.subscriber.pc.ontrack` subscription, `RemoteAudioTrack` is created and the `Track` constructor is called with adding a `visibilitychange` listener. However, Livekit is not removing the listeners properly when `stream.onremovetrack` event occurs.

### The issue
The problem occurs when Livekit does not remove the document `visibilitychange` listener and (perhaps monitoring) on unsubscribe. This results in memory leaks every time on every resubscription to track.

### Reproduction

I'm subscribing to the track using the code below:
```javascript
room.participants.forEach((participant: RemoteParticipant) => {
  ...
  participant.audioTracks.forEach((remoteTrack: RemoteTrackPublication) => {
    ...
    remoteTrack.setSubscribed(true);
    remoteTrack.setEnabled(true);
```

On livekit it cause create a `RemoteTrack` with call Track constructor:
```javascript
class Track extends eventsExports.EventEmitter {
  constructor(mediaTrack, kind) {
    ...
    document.addEventListener('visibilitychange', this.appVisibilityChangedListener);
  }
}
```

And when I'm unsubscribing using the code below:
```javascript
room.participants.forEach((participant: RemoteParticipant) => {
  ...
  //if we needs to unsubscribe
  participant.tracks.forEach((remoteTrack: RemoteTrackPublication, key: string) => {
    remoteTrack.setSubscribed(false);
  });
```

After `stream.onremovetrack` Livekit is not clear the track listeners and perhaps timers (Monitor timer).

### System Info

```shell
System:
    OS: macOS 13.2.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 56.13 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - /usr/local/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
    npm: 8.19.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 112.0.5615.49
    Firefox: 111.0.1
    Safari: 16.3
  npmPackages:
    livekit-client: https://github.com/cosmosvideo/client-sdk-js.git => 1.7.1
```

### Severity

serious, but I can work around it

PR #655, opened 2023-04-14
Remove page visibility listener when last element gets detached

lukasIO:
fixes https://github.com/livekit/client-sdk-js/issues/647

on first call of `attach` the `appVisibilityListener` gets added. when `detach` is called and there are no more attached elements afterwards, the `appVisibilityListener` gets removed.


davidzhao (approved):
lgtm!

PR #668, opened 2023-04-23
Exposing numParticipants and numPublishers on Room

davidzhao:
Server will start returning an accurate count of participants and publishers.

This PR is preparing for the future where not all subscribers will have their metadata sent down.

### DIFF
--- a/src/proto/google/protobuf/timestamp.ts
+++ b/src/proto/google/protobuf/timestamp.ts
@@ -127,31 +127,22 @@ export const Timestamp = {
   },
 
   decode(input: _m0.Reader | Uint8Array, length?: number): Timestamp {
-    const reader = input instanceof _m0.Reader ? input : _m0.Reader.create(input);
+    const reader = input instanceof _m0.Reader ? input : new _m0.Reader(input);
     let end = length === undefined ? reader.len : reader.pos + length;
     const message = createBaseTimestamp();
     while (reader.pos < end) {
       const tag = reader.uint32();
       switch (tag >>> 3) {
         case 1:
-          if (tag != 8) {
-            break;
-          }
-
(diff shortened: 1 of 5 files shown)


src/room/Room.ts
```diff
@@ -99,23 +99,13 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
   localParticipant: LocalParticipant;
 
-  /** room metadata */
-  metadata: string | undefined = undefined;
-
   /** options of room */
   options: InternalRoomOptions;
 
-  private _isRecording: boolean = false;
+  private _room?: RoomModel;
```
lukasIO, line 108: nit: personally not a huge fan of `room._room`, maybe calling it `info` instead would be an option?

src/proto/google/protobuf/timestamp.ts
```diff
@@ -127,31 +127,22 @@ export const Timestamp = {
   },
 
   decode(input: _m0.Reader | Uint8Array, length?: number): Timestamp {
-    const reader = input instanceof _m0.Reader ? input : _m0.Reader.create(input);
+    const reader = input instanceof _m0.Reader ? input : new _m0.Reader(input);
```
lukasIO, line 130: we keep having diffs like this from time to time, wondering whether there's a difference in the ts-proto versions we're using...

Ocupe (approved):

src/proto/google/protobuf/timestamp.ts
davidzhao: It appears so. I just did a `yarn install` and it should be reverted back.

src/room/Room.ts
davidzhao: I'll update it to `roomInfo`

PR #680, opened 2023-05-01
Handle connection state mismatch with periodic reconciliation

davidzhao:
It's possible for Room's connection state to get out of sync with the underlying transport. This is rare but could happen when the computer is going to sleep. For example, when the computer sleeps, PeerConnection would go straight to "closed" without firing the callbacks.

We have also seen rare conditions where RTCEngine went to close without our knowledge.

This PR adds a last line of defense to ensure Room's state is always correct.

### DIFF
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -321,14 +321,7 @@ export class SignalClient {
       };
 
       this.ws.onclose = (ev: CloseEvent) => {
-        if (!this.isConnected) return;
-
-        log.debug(`websocket connection closed: ${ev.reason}`);
-        this.isConnected = false;
-        if (this.onClose) {
-          this.onClose(ev.reason);
-        }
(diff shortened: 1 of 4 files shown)


src/api/SignalClient.ts
```diff
@@ -321,14 +321,7 @@ export class SignalClient {
...
 
       this.ws.onclose = (ev: CloseEvent) => {
-        if (!this.isConnected) return;
-
-        log.debug(`websocket connection closed: ${ev.reason}`);
-        this.isConnected = false;
-        if (this.onClose) {
-          this.onClose(ev.reason);
-        }
-        this.ws = undefined;
```
davidzhao, line 331: I think we were leaking `clearPingInterval` here and when ping timeout fires. Since engine could be recreated; the old copy would keep running.

lukasIO (approved):

PR #687, opened 2023-05-03
Change TS target to es2015

lukasIO:
we had an `array.at` call in the codebase which tripped up a user when using react native. 
Even in browser environments `array.at` isn't well supported. 
I was (wrongly) under the impression that babel would transpile this away, but in order to do that we would have to add a core-js polyfill. 
So instead of adding explicit polyfills for this via core-js, I lowered the target for TS compilation to ES2015 (so that we don't think we can use more recent features that might actually need a polyfill) and removed the `.at` usage (TS compiler still didn't catch that one by itself).

### DIFF
--- a/.changeset/orange-bobcats-tap.md
+++ b/.changeset/orange-bobcats-tap.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Change TS target to es2015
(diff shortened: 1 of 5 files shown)


davidliu (approved):
Tested as working on RN.

davidzhao (approved):

PR #689, opened 2023-05-04
Fix supportsAV1/VP9 compatibility with FireFox

davidzhao:
Only populating supported codecs in sample.

### DIFF
--- a/src/room/utils.ts
+++ b/src/room/utils.ts
@@ -41,7 +41,10 @@ export function supportsDynacast() {
 }
 
 export function supportsAV1(): boolean {
-  const capabilities = RTCRtpReceiver.getCapabilities('video');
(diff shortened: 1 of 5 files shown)


src/room/utils.ts
```diff
@@ -41,7 +41,10 @@ export function supportsDynacast() {
 }
 
 export function supportsAV1(): boolean {
-  const capabilities = RTCRtpReceiver.getCapabilities('video');
+  if (!('getCapabilities' in RTCRtpSender)) {
+    return false;
+  }
+  const capabilities = RTCRtpSender.getCapabilities('video');
```
davidzhao, line 47: Using sender capabilities instead of receiver. this more accurately reflects encoding support

cnderrauber (approved):

lukasIO (approved):

PR #697, opened 2023-05-10
Use barebones browser parser, fix browser version comparison

lukasIO:
While testing the custom browser parser, I ran into issues with the `compareVersion` function, so there's also some fixes for it as part of this PR.

### DIFF
--- a/src/room/utils.ts
+++ b/src/room/utils.ts
@@ -1,5 +1,6 @@
-import UAParser from 'ua-parser-js';
 import { ClientInfo, ClientInfo_SDK } from '../proto/livekit_models';
+import { getBrowser } from '../utils/browserParser';
+import type { DetectableBrowser } from '../utils/browserParser';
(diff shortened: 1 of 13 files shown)


src/room/utils.ts
```diff
@@ -197,12 +192,18 @@ export function getDevicePixelRatio(): number {
...
-  const k = Math.min(v1.length, v2.length);
+  const k = Math.min(parts1.length, parts2.length);
   for (let i = 0; i < k; ++i) {
     const p1 = parseInt(parts1[i], 10);
     const p2 = parseInt(parts2[i], 10);
     if (p1 > p2) return 1;
     if (p1 < p2) return -1;
+    if (i === k - 1 && p1 === p2) return 0;
+  }
+  if (v1 === '' && v2 !== '') {
```
davidzhao, line 203: ah, good fix here for empty checks!

davidzhao (approved):

PR #700, opened 2023-05-11
Replace ts-proto with protobuf-es

lukasIO:
The only inconsistency that I could find was that we were using protobuf enum values of `Enum.UNRECOGNIZED` in two places. The generated types of protobuf-es do not expose that `UNRECOGNIZED` option, so not sure what the best way to work around this would be.


src/api/SignalClient.ts
```diff
@@ -63,7 +70,7 @@ export interface SignalOptions {
 
 type SignalMessage = SignalRequest['message'];
 
-type SignalKind = NonNullable<SignalMessage>['$case'];
+type SignalKind = NonNullable<SignalMessage>['case'];
```
davidzhao, line 73: interesting to see they kept this design

src/room/track/LocalVideoTrack.ts
```diff
@@ -404,7 +406,8 @@ export function videoQualityForRid(rid: string): VideoQuality {
     case 'q':
       return VideoQuality.LOW;
     default:
-      return VideoQuality.UNRECOGNIZED;
+      // FIXME should be unrecognized
+      return VideoQuality.OFF;
```
davidzhao, line 410: maybe default to something other than off? I know this is likely an invalid case.. but `OFF` is a reserved enum that only the server uses

davidzhao: maybe default to HIGH?

davidzhao (approved):
lgtm! This actually looks like a clean switch.

src/room/track/LocalVideoTrack.ts
lukasIO: makes sense! 
I think protobuf behaviour generally would be to default to the first enum value, but HIGH does make more sense in this context.

lukasIO:
the reason the test-workflow compat check is failing is due to https://github.com/robatwilliams/es-compat/issues/72. 
We are working around this issue already for `globalThis` by claiming to have a polyfill for it (which we don't).
But it's currently failing on this branch, because `protobuf-es` is making use of `BigInt` if it is available. It verifies that via runtime checks. 
The runtime check probably still needs some tweaking though, opened [an issue](https://github.com/bufbuild/protobuf-es/issues/486) to hopefully address this.

lukasIO:
> the test-workflow compat check is failing is due to https://github.com/robatwilliams/es-compat/issues/72.

temporarily set the log level of this check to `warn` instead of `error` to make the CI succeed. Added a comment to revert that change, once the plugin allows for additional customisation of rules.

PR #706, opened 2023-05-16
Preserve sample rate of audio track when pausing upstream

lukasIO:


lukasIO:
previous implementation was leaking memory, the latest commits address this.

lukasIO:
closing in favor of #712

PR #708, opened 2023-05-17
Make sure TrackUnsubscribed events are emitted before the publication…

lukasIO:
… gets deleted from maps


davidzhao (approved):
great catch!

PR #710, opened 2023-05-22
Ensure same framerates for iOS simulcast

davidliu:


src/room/participant/publishUtils.ts
```diff
@@ -321,6 +321,26 @@ function encodingsFromPresets(
     }
     encodings.push(encoding);
   });
+
+  // RN ios simulcast requires all same framerates.
```
lukasIO, line 325: do we want to log a warning/info when this happens so that devs are aware of this happening?

src/room/participant/publishUtils.ts
```diff
@@ -321,6 +321,26 @@ function encodingsFromPresets(
...
+      }
+    });
+
+    encodings.forEach((encoding) => {
+      encoding.maxFramerate = topFramerate;
+    });
+  }
+
+  encodings.forEach((encoding) => {
+    console.log(encoding);
```
lukasIO, line 342: this seems like debug leftovers

lukasIO (approved):

PR #712, opened 2023-05-24
Use replaceTrack(null) for pauseUpstream

davidzhao:
Previously we were creating placeholder tracks, that is not compatible with react-native.

### DIFF
--- a/.changeset/clean-parents-drum.md
+++ b/.changeset/clean-parents-drum.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Use replaceTrack(null) for pauseUpstream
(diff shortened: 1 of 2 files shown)


boks1971 (approved):

davidliu:
Tested as working on RN with both iOS and Android. On Android, seemed to add a small delay (increased to about half a second) when connected to a local server, but didn't seem to increase over repeated usages and the audio/video stayed in sync.

davidliu (approved):

lukasIO:
there's a Safari issue from 2018 that might be a problem https://bugs.webkit.org/show_bug.cgi?id=184911

Only looking at the date of the fix it seems like the bug is present on Safari versions prior to 12

davidzhao:
@lukasIO thanks for checking on this.. do you think we should keep the existing behavior for older browsers like Safari 12?

lukasIO:
the existing behaviour would then only apply to Safari 11.x as versions before that do not support canvas.captureStream that we use for creating the dummy video track. (https://caniuse.com/mdn-api_htmlcanvaselement_capturestream)

Alternatively we could throw an error instead when trying to use the API on Safari 11 (instead of ending the track) and declare the API to be only supported on Safari >= 12.

davidzhao:
I like the latter approach. will update the docstring and throw an exception < Safari 12

PR #716, opened 2023-05-26
Use replaceTrack(null) for pauseUpstream

davidzhao:
Previously we were creating placeholder tracks, that is not compatible
with react-native.

This is the second attempt. With the previous attempt in #712, paused tracks were not unpublished correctly.

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -248,10 +248,12 @@ export default abstract class LocalTrack extends Track {
     this.emit(TrackEvent.Ended, this);
   };
 
-  // pauses publishing to the server without disabling the local MediaStreamTrack
-  // this is used to display a user's own video locally while pausing publishing to
-  // the server.
-  // this API is unsupported on Safari < 12 due to a bug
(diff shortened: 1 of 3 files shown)


src/room/track/LocalTrack.ts
```diff
@@ -253,6 +248,10 @@ export default abstract class LocalTrack extends Track {
     this.emit(TrackEvent.Ended, this);
   };
 
+  // pauses publishing to the server without disabling the local MediaStreamTrack
+  // this is used to display a user's own video locally while pausing publishing to
+  // the server.
+  // this API is unsupported on Safari < 12 due to a bug
```
lukasIO, line 254: I think for the JSdocs to show up correctly, the comment needs to be wrapped in `/** */` instead of single line comments

lukasIO (approved):

src/room/track/LocalTrack.ts
davidzhao: ah.. I've been doing too much Go lately.. will update.

PR #720, opened 2023-05-29
Fix svc mode for chrome v113, change default svc mode to L3T3_KEY

cnderrauber:
Fix svc mode for chrome v113.
Change default svc mode to L3T3_KEY as same as google meet. It reduces ~30% CPU load when encoding 720p videos.
Add closable spatial layer for svc dynacast but disable it now for two blur/frozen issues
with current server/client:
1. chrome 113: when switching to up layer with scalability Mode change, it will generate a
low resolution frame and recover very quickly, but noticable
2. livekit sfu: additional pli request cause video frozen for a few frames, also noticable

### DIFF
--- a/src/room/participant/publishUtils.ts
+++ b/src/room/participant/publishUtils.ts
@@ -128,17 +128,15 @@ export function computeVideoEncodings(
     // svc use first encoding as the original, so we sort encoding from high to low
     switch (scalabilityMode) {
       case 'L3T3':
-        for (let i = 0; i < 3; i += 1) {
-          encodings.push({
-            rid: videoRids[2 - i],
-            scaleResolutionDownBy: 2 ** i,
-            maxBitrate: videoEncoding.maxBitrate / 3 ** i,
-            /* @ts-ignore */
-            maxFramerate: original.encoding.maxFramerate,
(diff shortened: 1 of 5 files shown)


src/room/participant/publishUtils.ts
```diff
@@ -128,17 +128,15 @@ export function computeVideoEncodings(
...
-            maxBitrate: videoEncoding.maxBitrate / 3 ** i,
-            /* @ts-ignore */
-            maxFramerate: original.encoding.maxFramerate,
-            /* @ts-ignore */
-            scalabilityMode: 'L3T3',
-          });
-        }
+      case 'L3T3_KEY':
+        encodings.push({
+          rid: videoRids[2],
```
davidzhao, line 133: what if there are only two layers specified ? and does it retain compatibility with Chrome < 113?

davidzhao (approved):

src/room/participant/publishUtils.ts
cnderrauber: We only provide 3 layers in the publish options now. And for Chrome version < 113, I had not tested it yet but am afraid that it is broken in the early version as the API change.

PR #726, opened 2023-06-01
Do not fire Disconnected event if engine had been replaced.

davidzhao:
If a new connection attempt comes in while a reconnect is in-progress, The previous websocket connection might not be fully closed. The server would close the old websocket connection with `reason: duplicate_identity`. Due to certain timing, this could cause us to incorrectly fire a `Disconnected` event.

The right behavior is to ignore the leave message, instead of processing it.


lukasIO (approved):
nice find!

Issue #721, opened 2023-05-29
LocalParticipant.reconnectFuture rejects with "Got disconnected during publishing attempt" that cannot be handled


HermanBilous:
### What I'm expecting
No unhandled rejections are thrown.

### What happens instead
We haven't discovered the exact way to reproduce this issue, but here are our logs:

```
2023-05-29T05:38:37.220Z: (info) disconnect from room [object Object]
2023-05-29T05:43:17.407Z: (warning) websocket closed [object Object]
2023-05-29T05:43:17.410Z: (warning) signal disconnected
2023-05-29T05:43:17.419Z: (info) resuming signal connection, attempt 0
2023-05-29T05:43:38.845Z: (warning) primary peerconnection disconnected
2023-05-29T05:49:21.310Z: (warning) websocket closed [object Object]
2023-05-29T05:49:21.913Z: (warning) reconnect disconnected
2023-05-29T05:49:22.502Z: (info) resuming signal connection, attempt 1
2023-05-29T05:49:56.776Z: (warning) websocket closed [object Object]
2023-05-29T05:50:02.583Z: (warning) reconnect disconnected
2023-05-29T05:50:03.953Z: (info) resuming signal connection, attempt 2
2023-05-29T05:50:07.221Z: (warning) leave-reconnect disconnected
2023-05-29T05:50:07.247Z: (warning) reconnect disconnected
2023-05-29T05:50:07.257Z: (warning) websocket closed [object Object]
2023-05-29T05:50:07.259Z: (warning) signal disconnected
2023-05-29T05:50:10.442Z: (info) reconnecting, attempt: 3
2023-05-29T05:50:27.584Z: (warning) reconnect disconnected
2023-05-29T05:50:33.099Z: (warning) ping timeout triggered. last pong received at: Mon, 29 May 2023 05:50:13 GMT
2023-05-29T05:50:33.115Z: (error) cannot send signal request before connected, type: ping
2023-05-29T05:50:33.117Z: (error) cannot send signal request before connected, type: pingReq
2023-05-29T05:50:34.088Z: (info) reconnecting, attempt: 4
2023-05-29T05:50:34.089Z: (warning) signal disconnected
2023-05-29T05:51:15.088Z: (warning) reconnect disconnected
2023-05-29T05:51:23.092Z: (info) reconnecting, attempt: 5
2023-05-29T05:51:56.091Z: (warning) reconnect disconnected
[...]

PR #727, opened 2023-06-01
Always catch reconnectFuture rejections

HermanBilous:
Fixes [#721](https://github.com/livekit/client-sdk-js/issues/721)

I'm not entirely sure that this is the right approach to handling the disconnect-during reconnect-during track publish. Maybe adding a method like RTCEnging.waitForPCConnected that returns promise that resolves when engine is connected? Anyway, I did as @lukasIO suggested, tested in my own project, and the fix seem to do the trick.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -143,8 +143,11 @@ export default class LocalParticipant extends Participant {
   };
 
   private handleDisconnected = () => {
-    this.reconnectFuture?.reject?.('Got disconnected during publishing attempt');
-    this.reconnectFuture = undefined;
+    if (this.reconnectFuture) {
+      this.reconnectFuture.promise.catch((e) => log.warn(e));
+      this.reconnectFuture?.reject?.('Got disconnected during reconnection attempt');
+      this.reconnectFuture = undefined;
+    }
   };
 
   /**


lukasIO:
> Maybe the code that calls publishTracks must ensure that the connection is established? Maybe adding a method like RTCEnging.waitForPCConnected that returns promise that resolves when engine is connected, and rejects if it disconnected?

those are fair questions. the `await reconnectFuture` is the current attempt at waiting for connection to be re-established. This will, as you suggest, still reject if connection does not establish. The only reason we have to do the additional `catch` is for cases where there is no "listener" to the reconnectFuture and thus also no handler. 
I was experimenting also with keeping track of "listeners" to the future and swallowing the rejection if there are no listeners, but that also didn't feel exactly right. 

We can think about it a bit more and see if we find an approach that feels less hacky.

HermanBilous:
Thanks for the review and feedback on suggestions. I'd appreciate if a fix for this issue was merged sooner than later. I've added suggestions because I didn't like my solution, but at the same time, I don't see myself tackling a more sophisticated one. I don't feel confident because I only have lib user perspective on it, and haven't had a chance to deep dive. Do you think you will be able to look for a better approach in near future, or know someone who could?

lukasIO:
Can't make any promises on a timeline for a more sophisticated solution right now, but don't mind merging this in for now, as it's definitely an improvement over the current status quo. 
cc @davidzhao for thoughts

lukasIO (approved):

PR #728, opened 2023-06-02
Always add codec info to AddTrackRequest

cnderrauber:
Add codec info to AddTrackRequest so the sfu can prefer codec for client which don't support setCodecPreferences. Previously, we only add codec info for the publication that have both videoCodec and backup Codec settled.

@lukasIO The subscriptionFailure e2e can't work for firefox as it actually publishes a vp8 codec.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -647,6 +647,16 @@ export default class LocalParticipant extends Participant {
               enableSimulcastLayers: true,
             },
           ];
+        } else if (opts.videoCodec) {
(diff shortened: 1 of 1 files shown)


lukasIO:
makes sense, so the server will leave _only_ the preferred codec in the server's sdp answer?

cnderrauber:
> makes sense, so the server will leave _only_ the preferred codec in the server's sdp answer?

No, it will reorder the codec list in the answer to put the preferred one in the head

lukasIO (approved):

PR #731, opened 2023-06-08
Svc compatibility for browsers

cnderrauber:
Fix svc encodings for chrome old version and safari. Firefox can't publish simulcast or svc of vp9, and the resolution is very low when publishing non-simulcast vp9, but it can works well as subscriber when decode & render chrome&safari's vp9 track.

### DIFF
--- a/src/room/participant/publishUtils.ts
+++ b/src/room/participant/publishUtils.ts
@@ -133,17 +133,13 @@ export function computeVideoEncodings(
     for (let i = 0; i < sm.spatial; i += 1) {
       encodings.push({
         rid: videoRids[2 - i],
-        active: false,
         maxBitrate: videoEncoding.maxBitrate / 3 ** i,
         /* @ts-ignore */
         maxFramerate: original.encoding.maxFramerate,
-        /* @ts-ignore */
-        // scalabilityMode: scalabilityMode,
       });
     }
     /* @ts-ignore */
     encodings[0].scalabilityMode = scalabilityMode;
-    encodings[0].active = true;
     log.debug('encodings', encodings);
     return encodings;
   }
--- a/src/room/participant/publishUtils.ts
+++ b/src/room/participant/publishUtils.ts
@@ -138,8 +138,8 @@ export function computeVideoEncodings(
         /* @ts-ignore */
         maxFramerate: original.encoding.maxFramerate,
         /* @ts-ignore */
-        // scalabilityMode: scalabilityMode, 
-      })
+        // scalabilityMode: scalabilityMode,
+      });
     }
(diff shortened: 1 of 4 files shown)


src/room/track/LocalVideoTrack.ts
```diff
@@ -351,8 +351,9 @@ async function setPublishingLayersForSender(
 
     let hasChanged = false;
 
+    const closableSpatial = false;
     /* @ts-ignore */
-    if (encodings.length === 1 && encodings[0].scalabilityMode) {
+    if (closableSpatial && encodings[0].scalabilityMode) {
```
lukasIO, line 356: looks like this would always evaluate to false, because `closableSpatial` is a const that's set to `false`?

src/room/participant/publishUtils.ts
```diff
@@ -123,27 +123,29 @@ export function computeVideoEncodings(
...
-      default:
-        // TODO : support other scalability modes
-        throw new Error(`unsupported scalabilityMode: ${scalabilityMode}`);
+    if (sm.spatial > 3) {
+      throw new Error(`unsupported scalabilityMode: ${scalabilityMode}`);
+    }
+    for (let i = 0; i < sm.spatial; i += 1) {
+      encodings.push({
+        rid: videoRids[2 - i],
+        active: false,
```
davidzhao, line 136: are you trying to disable all layers except for the lowest quality one? could you add a bit of comments here to indicate why we are doing this?

src/room/participant/publishUtils.ts
```diff
@@ -123,27 +123,29 @@ export function computeVideoEncodings(
...
+        maxBitrate: videoEncoding.maxBitrate / 3 ** i,
+        /* @ts-ignore */
+        maxFramerate: original.encoding.maxFramerate,
+        /* @ts-ignore */
+        // scalabilityMode: scalabilityMode,
```
davidzhao, line 141: nit: delete instead of comment out.

src/room/track/LocalVideoTrack.ts
cnderrauber: This code scope is closable spatial layers for svc, but chrome/svc has problem to support it now, the transition is not smooth, so disable it first, I'll add a comment to it

src/room/participant/publishUtils.ts
cnderrauber: For svc codec, the encodings are sorted from high to low, so only the highest layer is enabled. It does not affect the actual svc encodings whether lower encodings are closed or not as they are used to indicate a spatial layer.

davidzhao (approved):
lg!

PR #732, opened 2023-06-08
Improve simulcast docstring

davidzhao:


lukasIO (approved):
nice, this makes it a lot clearer!

PR #734, opened 2023-06-10
Handle device mute and document freeze events

davidzhao:
When a source is unable to produce, it'll emit `mute` events. We are currently not capturing these and it could lead to unstable/garbage data being published to the SFU.

This PR will generate a publisher mute event when it is unable to produce. It also consolidates the multiple places where we store the _mediaStreamTrack reference to ensure consistent handling.

We also handle `freeze` event emitted on the document. This takes place in Chrome when memory saver is activated. Google [recommends](https://developer.chrome.com/blog/page-lifecycle-api/#developer-recommendations-for-each-state) tearing down WebRTC connections.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -495,6 +495,9 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       // capturing both 'pagehide' and 'beforeunload' to capture broadest set of browser behaviors
       window.addEventListener('pagehide', this.onPageLeave);
       window.addEventListener('beforeunload', this.onPageLeave);
+    }
+    if (isWeb()) {
+      document.addEventListener('freeze', this.onPageLeave);
       navigator.mediaDevices?.addEventListener('devicechange', this.handleDeviceChange);
     }
     this.setAndEmitConnectionState(ConnectionState.Connected);
@@ -965,6 +968,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       if (isWeb()) {
         window.removeEventListener('beforeunload', this.onPageLeave);
         window.removeEventListener('pagehide', this.onPageLeave);
+        window.removeEventListener('freeze', this.onPageLeave);
         navigator.mediaDevices?.removeEventListener('devicechange', this.handleDeviceChange);
       }
     } finally {
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -118,7 +118,7 @@ export default abstract class LocalTrack extends Track {
       // note this is different from LocalTrack.mute because we do not want to
       // touch MediaStreamTrack.enabled
       newTrack.addEventListener('mute', () => {
-        console.log('pausing upstream due to device mute');
+        log.info('pausing upstream due to device mute');
         this.pauseUpstream();
       });
       newTrack.addEventListener('unmute', this.resumeUpstream);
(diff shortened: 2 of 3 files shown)


src/room/Room.ts
```diff
@@ -495,6 +495,9 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       // capturing both 'pagehide' and 'beforeunload' to capture broadest set of browser behaviors
       window.addEventListener('pagehide', this.onPageLeave);
       window.addEventListener('beforeunload', this.onPageLeave);
+    }
+    if (isWeb()) {
+      document.addEventListener('freeze', this.onPageLeave);
```
lukasIO, line 500: this will close the connection entirely. 
in situations where the document freezes for a short time only, would we want to handle the `resume` case more gracefully?

src/room/track/LocalTrack.ts
```diff
@@ -88,6 +92,53 @@ export default abstract class LocalTrack extends Track {
...
+    this.mediaStream = new MediaStream([newTrack]);
+    if (newTrack) {
+      newTrack.addEventListener('ended', this.handleEnded);
+      // when underlying track emits mute, it indicates that the device is unable
+      // to produce media. In this case we'll need to signal with remote that
+      // the track is "muted"
+      // note this is different from LocalTrack.mute because we do not want to
+      // touch MediaStreamTrack.enabled
+      newTrack.addEventListener('mute', () => {
+        console.log('pausing upstream due to device mute');
```
lukasIO, line 121: should probably be a log.warn or log.info

src/room/track/LocalTrack.ts
```diff
@@ -203,16 +230,7 @@ export default abstract class LocalTrack extends Track {
...
-    if (this.sender) {
-      // Track can be restarted after it's unpublished
-      await this.sender.replaceTrack(newTrack);
-    }
-
-    this._mediaStreamTrack = newTrack;
-
-    await this.resumeUpstream();
-
-    this.mediaStream = mediaStream;
```
lukasIO, line 215: did you remove the `mediaStream` assignment on purpose?

src/room/track/LocalTrack.ts
```diff
@@ -133,37 +184,12 @@ export default abstract class LocalTrack extends Track {
...
-    // sync muted state with the enabled state of the newly provided track
-    this._mediaStreamTrack.enabled = !this.isMuted;
-
-    await this.resumeUpstream();
-
-    this.attachedElements.forEach((el) => {
-      attachToElement(track, el);
-    });
-
-    this.mediaStream = new MediaStream([track]);
```
lukasIO, line 165: IIRC react native depends on this to work properly

src/room/track/LocalTrack.ts
davidzhao: good catch.. I'll update.

src/room/Room.ts
davidzhao: I think in this case Chrome moves the page from memory to storage. They are recommending to disconnect WebRTC connections so I think it should be safe to disconnect fully.

src/room/track/LocalTrack.ts
davidzhao: I moved this into `setMediaStreamTrack` [here](https://github.com/livekit/client-sdk-js/pull/734/files#diff-43403f51e47283dce4cb915e0803845266c22afe713a26ea682d1aafbc3a42d5R112)

src/room/track/LocalTrack.ts
davidzhao: I moved it into `setMediaStreamTrack`, where it creates a new mediastream whenever Track is assigned. @davidliu are there any requirements from RN to keep the mediaStream that was returned by gUM?

src/room/track/LocalTrack.ts
lukasIO: thanks, missed that!

lukasIO (approved):

PR #735, opened 2023-06-10
Increase default adaptiveStream pixelDensity on high-density screens

davidzhao:
Mobile devices typically use high density screens (2.5-3+). On these platforms, using default CSS dimensions tends to produce less than ideal streaming quality when using adaptiveStream defaults.

### DIFF
--- a/src/room/track/RemoteVideoTrack.ts
+++ b/src/room/track/RemoteVideoTrack.ts
@@ -1,11 +1,11 @@
 import { debounce } from 'ts-debounce';
 import log from '../../logger';
 import { TrackEvent } from '../events';
-import { computeBitrate } from '../stats';
 import type { VideoReceiverStats } from '../stats';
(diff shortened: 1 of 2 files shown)


davidliu:
Can we switch to `screen` pixel density by default? Or does that not work for mobile browser cases?

davidzhao:
> Can we switch to `screen` pixel density by default? Or does that not work for mobile browser cases?

I'm trying to avoid a change in behavior on desktop systems. If we defaulted to screen, it will increase the bandwidth consumption on most displays today. On mobile, it'll mean fetching 720p by default for a video tile that takes about 200 css pixels.

Maybe this is a change we make for 2.0? as more efficient codecs are used.

davidliu (approved):
LGTM

lukasIO (approved):
that makes sense! 
we should probably do this with a minor changeset, as it is changing current default behaviour on mobile

PR #738, opened 2023-06-12
Add camera facing mode detection

Ocupe:
The goal of this PR is to add a unified way to get the facing mode of local camera tracks. Because the `facingMode` property is not available on all browsers, we have to do some extra work to increase the changes to infer the facing mode. 
This is WIP and open for suggestions.

## Approach
Currently we use a 2-step approach to determine the `facingMode`. 
1. Whenever it is available, use the `facingMode` property to get and return the correct facing mode.
2. If the facingMode property is not available, try to parse the device label string for known names or fractions like "backward facing" or something like that.

### Confidence Level
Since this aproche has the possibility to fail, we should return a value indicating how confident we are about the returned result. The first thing that came to mind was `high`, `medium` and `low`. Results from step 1. have a `high` confidence level and should skip step 2. altogether. Results from step 2 could then return an appropriate confidence level.

> Note: If the label detection approach is feasible, we could think of a separate package where we could update and maintain a list of regex/device labels that can also be used by other projects. If anyone knows of an existing list of device names that we could use, that would be great.

### DIFF
--- a/src/room/track/utils.ts
+++ b/src/room/track/utils.ts
@@ -147,15 +147,16 @@ export function facingModeFromLocalTrack(
     log.debug('rawFacingMode', { rawFacingMode });
     if (rawFacingMode && typeof rawFacingMode === 'string' && isFacingModeValue(rawFacingMode)) {
       result = { facingMode: rawFacingMode, confidence: 'high' };
-    } else if (Array.isArray(rawFacingMode) && rawFacingMode.length > 0) {
-      result = { facingMode: rawFacingMode[0], confidence: 'medium' };
     }
   }
 
-  // 2. If don't have a high confidence we try to get the facing mode from the device label.
+  // 2. If we don't have a high confidence we try to get the facing mode from the device label.
   if (['low', 'medium'].includes(result.confidence)) {
     log.debug(`Try to get facing mode from device label: (${track.label})`);
-    facingModeFromDeviceLabel(track.label);
+    const labelFacingMode = facingModeFromDeviceLabel(track.label);
+    if (labelFacingMode !== undefined) {
+      result = { facingMode: labelFacingMode, confidence: 'medium' };
+    }
   }
 
   if (result.confidence === 'low') {
@@ -167,15 +168,21 @@ export function facingModeFromLocalTrack(
   return result;
 }
 
+const knownDeviceLabels = new Map<string, FacingMode>([['obs virtual camera', 'environment']]);
 /**
  * Attempt to analyze the device label to determine the facing mode.
  *
  * @alpha
  */
-function facingModeFromDeviceLabel(deviceLabel: string): VideoCaptureOptions['facingMode'] {
+function facingModeFromDeviceLabel(deviceLabel: string): FacingMode | undefined {
+  // Empty string is a valid device label but we can't infer anything from it.
   if (deviceLabel === '') {
     return undefined;
   }
+  // Can we match against a wieldy known device label.
+  if (knownDeviceLabels.has(deviceLabel)) {
+    return knownDeviceLabels.get(deviceLabel);
+  }
+
   // TODO Attempt to analyze the device label to determine the facing mode
-  return 'user';
 }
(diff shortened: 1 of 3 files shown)


src/room/track/utils.ts
```diff
@@ -112,3 +114,75 @@ export function getNewAudioContext(): AudioContext | void {
...
+  if (['low', 'medium'].includes(result.confidence)) {
+    log.debug(`Try to get facing mode from device label: (${track.label})`);
+    const labelFacingMode = facingModeFromDeviceLabel(track.label);
+    if (labelFacingMode !== undefined) {
+      result = { facingMode: labelFacingMode, confidence: 'medium' };
+    }
+  }
+
+  if (result.confidence === 'low') {
+    log.warn(`Inferring facing mode from track with low confidence, possibly wrong result.`, {
```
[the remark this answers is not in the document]
Ocupe: Yes, we should probably remove this 👍

src/room/track/utils.ts
```diff
@@ -112,3 +114,75 @@ export function getNewAudioContext(): AudioContext | void {
...
+/**
+ * Try to analyze the local MediaStreamTrack or device label to determine the facing mode of a device.
+ * There is no accurate and common property supported by all browsers to detect whether a video track is originated from a user- or environment-facing camera device.
+ * For this reason, we use the `facingMode` property when available, but will fall back on a string-based analysis of the device label to determine the facing mode.
+ *
+ * {@link https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints/facingMode | MDN docs on facingMode}
+ * @alpha
+ */
+export function facingModeFromLocalTrack(
+  localTrack: LocalTrack,
```
davidzhao, line 138: would be great for the helper to work with MediaStreamTrack as well.

src/room/track/utils.ts
```diff
@@ -112,3 +114,75 @@ export function getNewAudioContext(): AudioContext | void {
...
+  return result;
+}
+
+const knownDeviceLabels = new Map<string, FacingMode>([['obs virtual camera', 'environment']]);
+/**
+ * Attempt to analyze the device label to determine the facing mode.
+ *
+ * @alpha
+ */
+function facingModeFromDeviceLabel(deviceLabel: string): FacingMode | undefined {
```
davidzhao, line 177: do we want to do a lowerCase() treatment on this label?

src/room/track/utils.ts
```diff
@@ -112,3 +114,87 @@ export function getNewAudioContext(): AudioContext | void {
...
+      result = { facingMode: labelFacingMode, confidence: 'medium' };
+    }
+  }
+
+  return result;
+}
+
+const knownDeviceLabels = new Map<string, FacingMode>([['obs virtual camera', 'environment']]);
+const knownDeviceLabelEndings = new Map<string, FacingMode>([
+  ['s iphone camera', 'environment'], // `<name>'s iPhone Camera` | Rear-facing iPhone camera used as web cam.
```
theomonnom, line 169: Just thinking out loud, I think the camera name of the Iphone camera depends on the language of the phone

lukasIO: good point. Also: `Iphone's camera as a webcam` for me (personally when I set it up like that) means that it is in fact user facing. But of course that's not true for every one

Ocupe: @theomonnom for me the camera name is language specific when I'm on mobile but it seems not to be when I'm on desktop and then select the camera as webcam but I could be wrong here. Can you verify this case, on desktop with iPhone as mobile camera the iPhone name is frensh?

@lukasIO That's a good point. I guess there is no way to know how the user will use the camera. The whole project seems like a fight against windmills. Maybe we should add an option to the local user tile to override the mirroring explicitly?

lukasIO: yeah, I think an option to override mirroring would be the best! we can still do some elaborate guessing, but the user needs to be able to overrule our assumptions

lukasIO (approved):
lgtm, just needs a changeset!

src/room/track/utils.ts
theomonnom: The name of the phone is just different, it is `iPhone de Theo` instead of `Theo's iphone `

Ocupe: @theomonnom thanks for checking 👍

theomonnom (approved):
lgtm

davidzhao (approved):

PR #739, opened 2023-06-12
When restarting a local track only replace a track after track processor initializes

burzomir:
Hello @davidzhao @lukasIO 
I created this to handle this specific use case:
Switching from environment to user camera with background replacement applied. 

Currently the only way to achieve that is to call `setProcessor` and then `restart` but it causes multiple calls to `replaceTrack` which in turn produces flickering picture (previous track visible for a moment, then the new track, then the new track with processor applied).

It would be also nice to pass a track processor directly to `restart` call. Currently when I want to restart a track with the new processor I have to call `setProcessor` first which will initialize and run on the current track (I'm transmitting a track with a processor which purpose is to be run only on the new track not the current). Then I call `restart` which leads to the second initialization of the processor on the new track.

I though of simply adding it as a second paramter but it potentially could break the API. What do you think?

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -280,7 +280,7 @@ export default abstract class LocalTrack extends Track {
     }
   }
 
-  protected handleEnded = () => {
+  private handleEnded = () => {
     if (this.isInBackground) {
       this.reacquireTrack = true;
     }
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -37,6 +37,8 @@ export default abstract class LocalTrack extends Track {
 
   protected processorLock: Mutex;
 
+  protected resumeUpstreamWithProcessedTrack = false;
+
   /**
    *
    * @param mediaTrack
@@ -98,7 +100,7 @@ export default abstract class LocalTrack extends Track {
     return this.processor?.processedTrack ?? this._mediaStreamTrack;
   }
 
-  private async setMediaStreamTrack(newTrack: MediaStreamTrack, force?: boolean) {
+  private async setMediaStreamTrack(newTrack: MediaStreamTrack, force?: boolean, replace = true) {
     if (newTrack === this._mediaStreamTrack && !force) {
       return;
     }
@@ -127,18 +129,20 @@ export default abstract class LocalTrack extends Track {
       newTrack.addEventListener('unmute', this.resumeUpstream);
       this._constraints = newTrack.getConstraints();
     }
-    if (this.sender) {
+    if (this.sender && replace) {
       await this.sender.replaceTrack(newTrack);
     }
     this._mediaStreamTrack = newTrack;
     if (newTrack) {
       // sync muted state with the enabled state of the newly provided track
       this._mediaStreamTrack.enabled = !this.isMuted;
-      // when a valid track is replace, we'd want to start producing
-      await this.resumeUpstream();
-      this.attachedElements.forEach((el) => {
-        attachToElement(newTrack, el);
-      });
+      if (replace) {
+        // when a valid track is replace, we'd want to start producing
+        await this.resumeUpstream();
+        this.attachedElements.forEach((el) => {
+          attachToElement(newTrack, el);
+        });
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


lukasIO:
Thanks for the PR.

How about simply moving the current `await this.setProcessor` to _before_ `sender.replaceTrack` - that seems lower impact and would also solve your issue IIUC?

burzomir:
@lukasIO 

I tried to simply move `await this.setProcessor` before `sender.replaceTrack` but it doesn't work for me. Probably because `setProcessor` will also call `restartTrack` and use old constraints.

Currently what I do:
```
async function switchToUserFacingCamera(environmentFacingCamera){
  // participants see clear environment
  await environmentFacingCamera.setProcessor(backgroundBlur)
  // some time passes to apply background blur
  // participants see blurred environment
  await environmentFacingCamera.restart({ facingMode: 'user'})
  // participants see participant's face without background blur
  // some time passes to apply background blur
  // background blur is visible 
}
```

What I would like to do:
```
async function switchToUserFacingCamera(environmentFacingCamera){
  // participants see clear environment
  await environmentFacingCamera.restart({ facingMode: 'user'}, backgroundBlurProcessor)
  // some time passes to apply background blur
  // participants see participant's face with background blur
}
```

burzomir:
@lukasIO @davidzhao I have another idea: I'd implement a `switch` method for `LocalVideoTrack`. Would you be interested in this implementation?

```
async switch(videoCaptureOptions, trackProcessor){
  // - Pause upstream
  // - Detach the current track
  // - Stop listening to the current track's events (ended, muted, ...)
  // - Stop the current processor without calling sender.replaceTrack
  // - Re-acquire new track using new videoCaptureOptions
  // - Start listening to the new track's events (ended, muted, ...)
  // - Run the new trackProcessor without calling sender.replaceTrack
  // - sender.replace(newTrackWithProcessorApplied)
  // - Resume upstream
}

// Example

cameraTrack.switch({ facingMode: 'user' }, backgroundBlurProcessor)
cameraTrack.switch({ facingMode: 'environment' }, noEffectProcessor)

```

lukasIO:
sorry for the delay, I had another look at this and feel like the way to go might be to have an optional `restartTrack: boolean` as a parameter for `stopProcessor()`. 

that way we could say that `stopProcessor` call within `setProcessor` always does `stopProcessor(false)` in order to not do an unnecessary track restart.

that way flipping the facing mode would just be `track.restartTrack(newConstraints)`

burzomir:
@lukasIO I tried your fix, but it doesn't solve the issue. I guess I wasn't clear enough with my issue description, so here's a short video of what happens when I use the following code. The switch method I implemented in this pull request reattaches the track only once after the track processor is initialized.

**Notice that the video track is reattached 4 times.**

```
async function switchToUserFacingCamera(environmentFacingCamera){
  // participants see clear environment
  await environmentFacingCamera.setProcessor(backgroundBlur)
  // some time passes to apply background blur
  // participants see blurred environment
  await environmentFacingCamera.restart({ facingMode: 'user'})
  // participants see participant's face without background blur
  // some time passes to apply background blur
  // background blur is visible 
}
```

https://github.com/livekit/client-sdk-js/assets/13577810/5d942d17-e782-4542-af32-051d49906da6

burzomir:
@lukasIO 

"Your idea is to have backgroundBlur disabled on environment facing cameras and only apply it once you are switching back to user facing camera?" - That's correct.

“but switching device and setting a processor are separate things, right? it doesn't necessarily happen all the time that you do both at the same time.” - ideally, for my use case, it should happen at the same time.

“I'd even argue that's a rare occasion.” - I believe people may want it quite often when targeting mobile devices.

lukasIO:
> I believe people may want it quite often when targeting mobile devices

that's fair, I agree. 

Coming back to your original suggestion, I think in that case having an overload on `restartTrack` that also accepts a processor as a second argument (or rather an options object, that can hold a processor) would be the most straight forward way to make this work.

burzomir:
@lukasIO I added the processor parameter to the restart method, but changes had to be more widespread to handle the scenario correctly. I'm open to refactoring suggestions.

burzomir:
@lukasIO  Anything new?

burzomir:
Fixed in https://github.com/livekit/client-sdk-js/pull/842

PR #747, opened 2023-06-15
Fix replaceTrack sync issue

boris-graeff:
Hey, is there a reason why there's no await in front of the `this.setMediaStreamTrack(track);` ?

I was facing the following issue :

```
await videoTrack.track.restartTrack({ resolution })
console.log(videoTrack.track.dimensions) => undefined
```

which is solved with the await

### DIFF
--- a/.changeset/blue-poets-repair.md
+++ b/.changeset/blue-poets-repair.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Fix missing await for async setMediaStreamTrack calls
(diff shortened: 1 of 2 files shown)


davidzhao:
thank you for the fix! This was an oversight on my part. Could you also add it to where we call it in `restart` [here](https://github.com/livekit/client-sdk-js/pull/747/files#diff-43403f51e47283dce4cb915e0803845266c22afe713a26ea682d1aafbc3a42d5R230)?

and run `yarn changeset` to generate a changelog for this for the next release.

boris-graeff:
@davidzhao , I updated the PR with what you asked for :+1:

davidzhao (approved):

davidzhao:
thank you!

PR #755, opened 2023-06-27
More granularity to getDevices

toger5:
Add the capability to ask for specific devices during the permission request that is caused by `getUserMedia` in `getDevices`.
Additionally this returns the devices the user has selected in the browser permission prompt.

Signed-off-by: Timo K <toger5@hotmail.de>

### DIFF
--- a/src/room/DeviceManager.ts
+++ b/src/room/DeviceManager.ts
@@ -3,6 +3,10 @@ import { isSafari } from './utils';
 
 const defaultId = 'default';
 
+export interface PermittedDevices {
+  audioDeviceId?: string;
+  videoDeviceId?: string;
+}
 export default class DeviceManager {
   private static instance?: DeviceManager;
 
@@ -15,12 +19,13 @@ export default class DeviceManager {
     return this.instance;
   }
 
-  static userMediaPromiseMap: Map<MediaDeviceKind, Promise<MediaStream>> = new Map();
+  static userMediaPromiseMap: Map<string, Promise<MediaStream>> = new Map();
 
   async getDevices(
     kind?: MediaDeviceKind,
+    mediaConstraints?: MediaStreamConstraints,
     requestPermissions: boolean = true,
-  ): Promise<MediaDeviceInfo[]> {
+  ): Promise<[MediaDeviceInfo[], PermittedDevices]> {
     if (DeviceManager.userMediaPromiseMap?.size > 0) {
       log.debug('awaiting getUserMedia promise');
       try {
@@ -33,13 +38,25 @@ export default class DeviceManager {
         log.warn('error waiting for media permissons');
       }
     }
+    console.log('userMediaDevice promises resolved');
     let devices = await navigator.mediaDevices.enumerateDevices();
+    // On firefox the user is given the option to choose which device they want to grant usage permission
+    // selectedDevices stores this so the user choice is accessible
+    let permittedDevices: PermittedDevices = {};
+
+    // For backwards compatibility we set mediaConsraints based on requestPremissions
+    if (!mediaConstraints && requestPermissions) {
+      mediaConstraints = {
+        video: kind !== 'audioinput' && kind !== 'audiooutput',
+        audio: kind !== 'videoinput',
+      };
+    }
 
     if (
-      requestPermissions &&
-      kind &&
+      mediaConstraints &&
+      (mediaConstraints.audio || mediaConstraints.video) &&
       // for safari we need to skip this check, as otherwise it will re-acquire user media and fail on iOS https://bugs.webkit.org/show_bug.cgi?id=179363
-      (!DeviceManager.userMediaPromiseMap.get(kind) || !isSafari())
+      (!DeviceManager.userMediaPromiseMap.get(JSON.stringify(mediaConstraints)) || !isSafari())
     ) {
       const isDummyDeviceOrEmpty =
         devices.length === 0 ||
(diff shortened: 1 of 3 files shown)


src/room/DeviceManager.ts
```diff
@@ -3,6 +3,10 @@ import { isSafari } from './utils';
 
 const defaultId = 'default';
 
+export interface PermittedDevices {
```
lukasIO, line 6: any particular reason why this isn't mapped to media device kinds? I.e. audioinput, audiooutput, videoinput. Seems like audiooutput might be interesting as well, given that [selectAudioOutput](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/selectAudioOutput) is around the corner already

src/room/DeviceManager.ts
```diff
@@ -15,12 +19,13 @@ export default class DeviceManager {
     return this.instance;
   }
 
-  static userMediaPromiseMap: Map<MediaDeviceKind, Promise<MediaStream>> = new Map();
+  static userMediaPromiseMap: Map<string, Promise<MediaStream>> = new Map();
```
lukasIO, line 22: not a big fan of `string` here as a type. 
IIUC the main thing we'd want to add is support for acquiring device specific (i.e. deviceId) permissions. 
I guess we could keep the mediadevicekind here and check after the promise resolved whether the newly requested deviceId is part of the `permittedDevices`? And only in case it's not, we could request permissions specifically for that device?

src/room/Room.ts
```diff
@@ -281,15 +281,22 @@ class Room extends EventEmitter<RoomEventCallbacks> {
...
+   * In some browsers the user is given the option to select a device in the permission prompt)
    */
   static getLocalDevices(
     kind?: MediaDeviceKind,
-    requestPermissions: boolean = true,
-  ): Promise<MediaDeviceInfo[]> {
-    return DeviceManager.getInstance().getDevices(kind, requestPermissions);
+    mediaConstraints?: MediaStreamConstraints,
+    requestPermissions?: boolean,
+  ): Promise<[MediaDeviceInfo[], PermittedDevices]> {
```
lukasIO, line 298: both the parameters and the return type would be a breaking change to user facing APIs. We'd want to avoid that. 
Two options I see:
- overload the function in a way that it takes in `requestPermissions` as a boolean as second argument and in that case it returns the plain mediaDeviceInfo array as before, and only if `requestPermissions` is set and not of type boolean, return the new array that includes permitted devices. (also wondering if returning an object here instead of the array would be clearer to users)
- factor the new stuff out in a separate function, e.g. `getLocalDevicesWithConstraints` and not touch the existing one at all

src/room/DeviceManager.ts
```diff
@@ -15,12 +19,13 @@ export default class DeviceManager {
...
 
   async getDevices(
     kind?: MediaDeviceKind,
+    mediaConstraints?: MediaStreamConstraints,
```
lukasIO, line 26: would it be enough here to have a `ConstrainDOMString` type that would allow for stuff like `{deviceId: {exact: 'my-device-id'}}` ? 
IIRC the reason to have `MediaStreamConstraints` here, is to be able to have a single permissions request for both mic and camera, right? 

This does somehow make the `kind` parameter redundant, so wondering if this might actually be its own function (also mentioned this in another comment) like `getDevicesByConstraints(constraints)` because that's effectively all we need to know? (also I guess mediaConstraints won't work for audiooutput)

src/room/Room.ts
toger5: I am very much in favor of factoring things out into a new function at this point. Since this very much becomes a `getLocalDevicesWithPermissionPrompt` (maybe a better name would be nice) function.

src/room/DeviceManager.ts
toger5: yes indeed. This solves both backwards compatibilty and gives us the freedom to make this "new" function as nice as possible (no duplicate `kind` + `mediaConstraints` props.

src/room/DeviceManager.ts
toger5: Using a string in this map was motivated by the idea to use the json of the Constraint object as the key for the userMediaPromiseMap. This would guarantee, that each constraint only is called once. the `userMediaPromiseMap` currently makes not much sense if it is also possible to request permission for both (video and audio)

src/room/track/LocalTrack.ts
```diff
@@ -124,10 +124,11 @@ export default abstract class LocalTrack extends Track {
     if (this.sender) {
       await this.sender.replaceTrack(newTrack);
     }
+    const previousMuteState = this._mediaStreamTrack.muted;
     this._mediaStreamTrack = newTrack;
     if (newTrack) {
       // sync muted state with the enabled state of the newly provided track
-      this._mediaStreamTrack.enabled = !this.isMuted;
+      this._mediaStreamTrack.enabled = !previousMuteState;
       // when a valid track is replace, we'd want to start producing
```
toger5, line 132: I will keep this open for now to double check if there are no issues left with this part:
It should be like this:
```ts
    const previousMuteState = this._mediaStreamTrack.enabled;
    this._mediaStreamTrack = newTrack;
    if (newTrack) {
      // sync muted state with the enabled state of the newly provided track
      this._mediaStreamTrack.enabled = !this.isMuted;
      this._mediaStreamTrack.enabled = !previousMuteState;
      // when a valid track is replace, we'd want to start producing
```

lukasIO: the idea is to treat the state of the `LocalTrack` as source of truth for the `muted` state. 
we will sync the newly create mediaStreamTrack to reflect the state of the LocalTrack. 
Your changes would have the opposite effect.

toger5: okay. so than whenever they are out of sync it would be interesting how/where `_mediaStreamTrack.enabled` is changed outside of a reaction to `isMuted` changing

PR #757, opened 2023-06-28
Use active device when publishing a new track

lukasIO:


src/room/participant/LocalParticipant.ts
```diff
@@ -464,14 +468,30 @@ export default class LocalParticipant extends Participant {
...
+      switch (track.source) {
+        case Track.Source.Microphone:
+          deviceKind = 'audioinput';
+          break;
+        case Track.Source.Camera:
+          deviceKind = 'videoinput';
+        default:
+          break;
+      }
+      if (deviceKind) {
```
davidzhao, line 483: should this be set only if activeDeviceMap contains the device?

src/room/participant/LocalParticipant.ts
```diff
@@ -464,14 +468,30 @@ export default class LocalParticipant extends Participant {
...
+        defaultConstraints = { deviceId: this.activeDeviceMap.get(deviceKind) };
+      }
+    }
     // convert raw media track into audio or video track
     if (track instanceof MediaStreamTrack) {
       switch (track.kind) {
         case 'audio':
-          track = new LocalAudioTrack(track, undefined, true);
+          track = new LocalAudioTrack(track, defaultConstraints, true);
```
davidzhao, line 491: by passing in a deviceId constraint, I think we would lose the other constraints already set on the device (such as dimensions). should this be appending it to the existing constraints instead of creating an empty set?

here's the current logic in LocalTrack constructor. `setMediaStreamTrack` grabs it from the local track.

```
    this.setMediaStreamTrack(mediaTrack);
    if (constraints) {
      this.constraints = constraints;
    }
```

src/room/track/LocalAudioTrack.ts
```diff
@@ -63,7 +63,9 @@ export default class LocalAudioTrack extends LocalTrack {
     try {
       if (
         this.source === Track.Source.Microphone &&
-        (this.stopOnMute || this._mediaStreamTrack.readyState === 'ended') &&
+        (this.stopOnMute ||
+          this._mediaStreamTrack.readyState === 'ended' ||
+          this._mediaStreamTrack.getSettings().deviceId !== this.constraints.deviceId) &&
```
davidzhao, line 68: does this.constraints.deviceId need unwrapping?

src/room/participant/LocalParticipant.ts
lukasIO: that's a good point, yeah we should be appending it only...

src/room/participant/LocalParticipant.ts
lukasIO: makes sense, yes!

src/room/track/LocalAudioTrack.ts
lukasIO: great catch! totally missed that...

davidzhao (approved):

PR #760, opened 2023-07-03
Handle new format streamId to better sync a/v tracks

cnderrauber:
Handle new format`msid` to better sync a/v tracks, especially when playout-delay is enabled. From remote tracks belong to the same `MediaStream`, chrome/safari will sync a/v better than standalone tracks. But firefox can't get benifits from this, and it also generates a unique track id for remote track might cause mismatch with signal track info, so keep the old format `msid` for it.


src/room/Room.ts
```diff
@@ -926,8 +926,11 @@ class Room extends EventEmitter<RoomEventCallbacks> {
     }
     const parts = unpackStreamId(stream.id);
     const participantId = parts[0];
-    let trackId = parts[1];
-    if (!trackId || trackId === '') trackId = mediaTrack.id;
+    let streamId = parts[1];
+    let trackId = mediaTrack.id;
```
lukasIO, line 930: do we need to test this also with older browser versions to make sure that chrome/safari are actually using our sdp id on the mediaTrack?

cnderrauber: Definitely be better if we can have test with more versions, I had tested 2 chrome versions (113 & 110). Can we use BrowserStack to do it automatically?

lukasIO: yeah, that should be possible!

lukasIO (approved):

src/room/Room.ts
```diff
@@ -926,8 +926,11 @@ class Room extends EventEmitter<RoomEventCallbacks> {
...
+    // firefox will get streamId (pID|trackId) instead of (pID|streamId) as it doesn't support sync tracks by stream
+    // and generates its own track id instead of infer from sdp track id.
+    if (streamId && streamId.startsWith('TR') && !trackId.startsWith('TR')) trackId = streamId;
```
davidzhao, line 933: this is also backwards compatibility with older server versions, correct?

davidzhao (approved):
lg!

src/room/Room.ts
cnderrauber: yeah, it will keep the old behavior with the old server version.

PR #766, opened 2023-07-05
Update dimensions when the remote track is detached

burzomir:
@davidzhao @lukasIO 

Use case for this change:
There are two video elements in the UI. The first is small (480x320) and placed in a participant grid. The second is big (1280x720) and only displays the active speaker. When no one is speaking, then the second video element is removed. In this case, I want to use only the low-resolution simulcast layer. Currently, simulcast layers change only on resize and visibility change events or when a remote track is attached.

### DIFF
--- a/src/room/track/RemoteVideoTrack.ts
+++ b/src/room/track/RemoteVideoTrack.ts
@@ -198,7 +198,6 @@ export default class RemoteVideoTrack extends RemoteTrack {
     for (const info of stopElementInfos) {
       this.stopObservingElementInfo(info);
     }
-    this.elementInfos = this.elementInfos.filter((info) => info.element !== element);
   }
 
   protected async handleAppVisibilityChanged() {
--- a/src/room/track/RemoteVideoTrack.ts
+++ b/src/room/track/RemoteVideoTrack.ts
@@ -123,7 +123,7 @@ export default class RemoteVideoTrack extends RemoteTrack {
     }
     this.elementInfos = this.elementInfos.filter((info) => info !== elementInfo);
     this.updateVisibility();
-    this.updateDimensions()
+    this.updateDimensions();
   }
 
   detach(): HTMLMediaElement[];
--- a/src/room/track/RemoteVideoTrack.ts
+++ b/src/room/track/RemoteVideoTrack.ts
@@ -131,6 +131,7 @@ export default class RemoteVideoTrack extends RemoteTrack {
     let detachedElements: HTMLMediaElement[] = [];
     if (element) {
       this.stopObservingElement(element);
+      this.updateDimensions();
       return super.detach(element);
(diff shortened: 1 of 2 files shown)


lukasIO:
hi @burzomir ,
thanks for the PR, this sounds like a reasonable change. 

Instead of calling it in `detach` could you add it to `stopObservingElementInfo` instead? We have `updateVisibility` there already.

burzomir:
@lukasIO Do you want me to replace `this.updateDimensions();` with `this.stopObservingElementInfo(elementInfo)` in this code?

```
detach(): HTMLMediaElement[];
  detach(element: HTMLMediaElement): HTMLMediaElement;
  detach(element?: HTMLMediaElement): HTMLMediaElement | HTMLMediaElement[] {
    let detachedElements: HTMLMediaElement[] = [];
    if (element) {
      this.stopObservingElement(element);
      this.updateDimensions();
      return super.detach(element);
    }
    detachedElements = super.detach();

    for (const e of detachedElements) {
      this.stopObservingElement(e);
    }
    this.updateDimensions();

    return detachedElements;
  }
  ```

File: src/room/track/RemoteVideoTrack.ts [hunk 1]
```diff
@@ -123,6 +123,7 @@ export default class RemoteVideoTrack extends RemoteTrack {
     }
     this.elementInfos = this.elementInfos.filter((info) => info !== elementInfo);
     this.updateVisibility();
+    this.updateDimensions();
```
lukasIO, line 126: I guess especially because we're calling this in a loop, it would make sense to replace this with the debounced version
```suggestion
    this.debouncedHandleResize();
```

burzomir: @lukasIO Done.

davidliu (approved):
One small comment, but LGTM.

src/room/track/RemoteVideoTrack.ts
```diff
@@ -195,7 +196,7 @@ export default class RemoteVideoTrack extends RemoteTrack {
   private stopObservingElement(element: HTMLMediaElement) {
     const stopElementInfos = this.elementInfos.filter((info) => info.element === element);
     for (const info of stopElementInfos) {
-      info.stopObserving();
+      this.stopObservingElementInfo(info);
     }
     this.elementInfos = this.elementInfos.filter((info) => info.element !== element);
```
davidliu, line 201: I think this line becomes redundant and can be removed?

burzomir: @davidliu You're right. `elementInfos.filter` is already executed in `stopObservingElementInfo.` I pushed a fix.

lukasIO:
this is looking great, thanks @burzomir !

PR #767, opened 2023-07-05
Fix peer connections leak

lukasIO:
PR addresses PeerConnections not clearing after GC kicks in in chrome.


davidzhao (approved):
awesome work!!!

PR #768, opened 2023-07-06
Fixed missed event listeners on MediaStreamTrack

davidzhao:
My previous attempt to consolidate `setMediaStreamTrack` missed the constructor case, so event listeners were missed the first run around.


lukasIO (approved):
great find!

Issue #770, opened 2023-07-06
[1.11.0+] createLocalScreenTracks returns `ended` tracks


burzomir:
### Describe the bug

I first discovered this while working with createLocalScreenTracks.  LocalTrack calls super(), which leads to setting this_mediaStreamTrack. Then it calls setMediaStreamTrack, which stops this._mediaStreamTrack if it is set.

```
1. LocalTrack.constructor(mediaStreamTrack) -> 
Track.constructor(mediaStreamTrack) -> 
this._mediaStreamTrack = mediaStreamTrack

2. LocalTrack.constructor(mediaStreamTrack) -> 
LocalTrack.setMediaStreamTrack(mediaStreamTrack) -> 
if (this._mediaStreamTrack) { this._mediaStreamTrack.stop() }
```

### Reproduction

Use createLocalScreenTracks. Returned track `readyState` is already `ended.`

### System Info

```shell
Any os, any browser.
```

### Severity

blocking all usage of LiveKit

lukasIO:
thanks for the report, fixed with https://github.com/livekit/client-sdk-js/pull/769

PR #771, opened 2023-07-06
createLocalScreenTracks returns ended tracks

burzomir:
@davidzhao @lukasIO This is a workaround that I figured out quickly. I hope it makes it easier to understand issue #770.

Could you take care of it or suggest a better solution that I could implement?


burzomir:
Fixed in #769

PR #772, opened 2023-07-06
Forward signal events through engine

lukasIO:
for easier cleanup


src/room/Room.ts
```diff
@@ -956,6 +950,7 @@ class Room extends EventEmitter<RoomEventCallbacks> {
   }
 
   private handleRestarting = () => {
+    console.log('restarting');
```
davidzhao, line 953: here too

src/room/participant/LocalParticipant.ts
```diff
@@ -821,7 +821,7 @@ export default class LocalParticipant extends Participant {
       }
     }
 
-    this.engine.negotiate();
+    await this.engine.negotiate();
```
davidzhao, line 824: ah this is a good change, so we would not ignore publish failures.

davidzhao (approved):
lgtm!

PR #773, opened 2023-07-06
Update constraints on track creation

toger5:
The constraints can be set to the wrong value causing the device to change on `unmute()`. This happens inside the `createLocalTracks` function if the browser gives access to different devices than the one requested by the `getUserMedia` constraints.


lukasIO (approved):
this makes a lot of sense, thanks for the PR!

PR #774, opened 2023-07-07
Updated README with script tag embed instructions

davidzhao:


lukasIO (approved):
nice addition! 

FWIW, in our `package.json` we have the entrypoint for `unpkg.com` specified directly already `  "unpkg": "./dist/livekit-client.umd.js",
` which is available at https://unpkg.com/livekit-client. No preference either way however

davidzhao:
a

> nice addition!
> 
> FWIW, in our `package.json` we have the entrypoint for `unpkg.com` specified directly already ` "unpkg": "./dist/livekit-client.umd.js",` which is available at https://unpkg.com/livekit-client. No preference either way however

ah ok, that's great to know!

PR #783, opened 2023-07-12
prepareConnection to better support Cloud routing and region-pinning

davidzhao:
When using Cloud, establish connection to fetch region URL so we have better control of routing decisions.

For projects with region-pinning enabled, `/settings/regions` would only return regions that are allowed. if prepareConnections is not used, Cloud would reject the initial connection if DNS-based routing ends up connecting to a disallowed region.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -1,7 +1,7 @@
 import EventEmitter from 'eventemitter3';
 import type { MediaAttributes } from 'sdp-transform';
-import { SignalClient } from '../api/SignalClient';
 import type { SignalOptions } from '../api/SignalClient';
+import { SignalClient } from '../api/SignalClient';
 import log from '../logger';
 import type { InternalRoomOptions } from '../options';
 import {
@@ -31,7 +31,7 @@ import {
 } from '../proto/livekit_rtc';
 import PCTransport, { PCEvents } from './PCTransport';
 import type { ReconnectContext, ReconnectPolicy } from './ReconnectPolicy';
-import { RegionUrlProvider } from './RegionUrlProvider';
+import type { RegionUrlProvider } from './RegionUrlProvider';
 import { roomConnectOptionDefaults } from './defaults';
 import {
   ConnectionError,
@@ -49,7 +49,6 @@ import { Track } from './track/Track';
 import type { TrackPublishOptions, VideoCodec } from './track/options';
 import {
   Mutex,
-  isCloud,
   isVideoCodec,
   isWeb,
   sleep,
@@ -122,7 +121,7 @@ export default class RTCEngine extends EventEmitter<EngineEventCallbacks> {
   // this is helpful to know if we need to restart ICE on the publisher connection
   private hasPublished: boolean = false;
 
-  // keep join info around for reconnect
+  // keep join info around for reconnect, this could be a region url
   private url?: string;
 
   private token?: string;
@@ -356,6 +355,11 @@ export default class RTCEngine extends EventEmitter<EngineEventCallbacks> {
       return undefined;
     }
     return getConnectedAddress(this.primaryPC);
+  }
+
+  /* @internal */
+  setRegionUrlProvider(provider: RegionUrlProvider) {
+    this.regionUrlProvider = provider;
   }
 
   private configure(joinResponse: JoinResponse) {
@@ -843,8 +847,10 @@ export default class RTCEngine extends EventEmitter<EngineEventCallbacks> {
     log.debug(`reconnecting in ${delay}ms`);
 
     this.clearReconnectTimeout();
-    if (this.url && this.token && isCloud(new URL(this.url))) {
-      this.regionUrlProvider = new RegionUrlProvider(this.url, this.token);
+    if (this.token && this.regionUrlProvider) {
+      // token may have been refreshed, we do not want to recreate the regionUrlProvider
+      // since the current engine may have inherited a regional url
+      this.regionUrlProvider.updateToken(this.token);
     }
     this.reconnectTimeout = CriticalTimers.setTimeout(
       () => this.attemptReconnect(disconnectReason),
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -351,7 +351,7 @@ class Room extends EventEmitter<RoomEventCallbacks> {
    * the current client to connect to if a token is provided.
    */
   async prepareConnection(url: string, token?: string) {
-    if (this.state !== 'disconnected') {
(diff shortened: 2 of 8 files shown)


src/room/RTCEngine.ts
```diff
@@ -843,8 +847,10 @@ export default class RTCEngine extends EventEmitter<EngineEventCallbacks> {
     log.debug(`reconnecting in ${delay}ms`);
 
     this.clearReconnectTimeout();
-    if (this.url && this.token && isCloud(new URL(this.url))) {
-      this.regionUrlProvider = new RegionUrlProvider(this.url, this.token);
```
davidzhao, line 847: this was problematic previously because `this.url` could have been a regional URL. In the case the region is unavailable, `RegionUrlProvider` would not have retrieved any valid URLs.

I think keeping the regional url in Engine makes sense. but we'd need to ensure the provider is able to fetch the appropriate regional URLs

src/room/Room.ts
```diff
@@ -337,15 +342,32 @@ class Room extends EventEmitter<RoomEventCallbacks> {
...
    */
-  async prepareConnection(url: string) {
-    await fetch(`http${url.substring(2)}`, { method: 'HEAD' });
+  async prepareConnection(url: string, token: string) {
+    log.debug(`prepareConnection to ${url}`);
+    try {
+      const regionUrlProvider = new RegionUrlProvider(url, token);
+      if (regionUrlProvider.isCloud()) {
+        this.regionUrlProvider = regionUrlProvider;
+        const regionUrl = await this.regionUrlProvider.getNextBestRegionUrl();
```
boks1971, line 359: Can this become `getBestRegionUrls()` or something more generic? Guessing `NextBest` is very specific and is a bit confusing if not used in that context. Or is that module maintaining a cursor internally and returns the next item?

boks1971 (approved):
lg!

paulwe (approved):

src/room/Room.ts
```diff
@@ -337,15 +342,32 @@ class Room extends EventEmitter<RoomEventCallbacks> {
...
+   * to speed up the connection attempt. This function will
+   * - perform DNS resolution and pre-warm the DNS cache
+   * - establish TLS connection and cache TLS keys
+   *
+   * With LiveKit Cloud, it will also determine the best edge data center for
+   * the current client to connect to.
    */
-  async prepareConnection(url: string) {
-    await fetch(`http${url.substring(2)}`, { method: 'HEAD' });
+  async prepareConnection(url: string, token: string) {
```
lukasIO, line 353: making `token` non-optional would be a breaking change

src/room/Room.ts
```diff
@@ -337,15 +342,32 @@ class Room extends EventEmitter<RoomEventCallbacks> {
   }
 
   /**
-   * prepares the connection to the livekit server by sending a HEAD request in order to
-   * 1. speed up DNS resolution
-   * 2. speed up TLS setup
-   * on the actual connection request
-   * throws an error if server is not reachable after the request timeout
-   * @experimental
+   * prepareConnection should be called as soon as the page is loaded, in order
```
lukasIO, line 345: this makes sense, but I'm wondering how realistic this is. 
If a (valid) token is required to call `prepareConnection` then that needs to be acquired first. 
In scenarios where the room connects on page load (without prior user interaction) `prepareConnection` will rather slow things down.

src/room/Room.ts
lukasIO, line 359: just to make sure, the returned region URLs would also include the node you're getting the settings from in case that is actually the best one to make an initial connection to?

src/room/utils.ts
```diff
@@ -483,3 +483,17 @@ export function unwrapConstraint(constraint: ConstrainDOMString): string {
   }
   throw Error('could not unwrap constraint');
 }
+
+export function toWebsocketUrl(url: string): string {
```
lukasIO, line 487: nice!

src/room/Room.ts
davidzhao: yeah, it returns the closest region urls to the user

src/room/Room.ts
davidzhao: made token an optional param for backwards-compat. if it's not passed in, it will keep existing behavior of a HEAD request

src/room/Room.ts
davidzhao: yeah it does keep an internal cursor to ensure we loop through the N that are returned

src/room/Room.ts
```diff
@@ -337,15 +342,31 @@ class Room extends EventEmitter<RoomEventCallbacks> {
...
+   * - perform DNS resolution and pre-warm the DNS cache
+   * - establish TLS connection and cache TLS keys
+   *
+   * With LiveKit Cloud, it will also determine the best edge data center for
+   * the current client to connect to.
    */
-  async prepareConnection(url: string) {
-    await fetch(`http${url.substring(2)}`, { method: 'HEAD' });
+  async prepareConnection(url: string, token?: string) {
+    log.debug(`prepareConnection to ${url}`);
```
lukasIO, line 354: should we bail here, in case `room.state !== 'disconnected'` ?

src/room/Room.ts
```diff
@@ -365,8 +386,19 @@ class Room extends EventEmitter<RoomEventCallbacks> {
...
     this.setAndEmitConnectionState(ConnectionState.Connecting);
-
-    const urlProvider = new RegionUrlProvider(url, token);
+    if (isCloud(new URL(url))) {
+      if (this.regionUrlProvider === undefined) {
+        this.regionUrlProvider = new RegionUrlProvider(url, token);
+        // trigger the first fetch without waiting for a response
+        // if initial connection fails, this will speed up picking regional url
+        // on subsequent runs
+        this.regionUrlProvider.fetchRegionSettings().catch((e) => {
```
lukasIO, line 395: I think this line can go after the `else` clause so that it runs even if `regionUrlProvider` has been set by a previous `prepareConnection` call already? 
also on `disconnect` we currently don't clear the regionUrlProvider. so with the sequence connect -> disconnect -> connect we would skip fetching region settings the second time around currently

lukasIO (approved):
looks good, just some minor comments.

src/room/Room.ts
davidzhao: yeah makes sense. it should clear out all the previous urls + providers as the primary url may have changed. I'll make that change

src/room/Room.ts
davidzhao: good idea

PR #787, opened 2023-07-13
Only check for setSinkId support on AudioContext if webaudiomix is enabled

lukasIO:
`audioContext.setSinkId` is only supported in recent Chrome versions. We only want to raise an error in that case, if `expWebAudioMix` is actually enabled


Ocupe (approved):
👍

PR #788, opened 2023-07-13
Also reset hasValidKey on manual ratchet request

lukasIO:
In case the `ratchetWindowSize` is `0`, we need to make sure that manual ratchet requests also reset the `hasValidKey` property on the key handlers.


cloudwebrtc (approved):
lgtm

PR #789, opened 2023-07-14
Add ScreenShareCaptureOptions.contentHint option

zesun96:
This PR adds one **ScreenShareCaptureOptions** attribute.
When we use shared desktop streams, we may need to set the track's contentHint to accommodate the shared content

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -468,6 +468,9 @@ export default class LocalParticipant extends Participant {
     if (tracks.length === 0) {
       throw new TrackInvalidError('no video track found');
     }
+    if (options.contentHint !== undefined) {
+      tracks[0].contentHint = options.contentHint;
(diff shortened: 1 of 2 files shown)


lukasIO:
thanks for the PR!

`contentHint` is not exclusive to screen share tracks and also not a `capture` option, but rather a property to set whenever. see https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/contentHint

Given this and the fact that it's not supported in Firefox suggest that it might be more appropriate to set this by accessing `track.mediaStreamTrack.contentHint` directly. 

It might make sense to have a higher level API for this directly on the `Track` with a getter and setter, which - depending on the `kind` of Track - allows for different contentHint values to be set and ignores/warns when not supported.

zesun96:
yes, `contentHint` is a common option，but it always effective for screen，directly modify the track also sounds good.

PR #790, opened 2023-07-14
Add TS decorator support and replace class arrow function fields with bound methods

lukasIO:


lukasIO:
needs more work, currently breaks `yarn sample`, converting to draft

davidzhao (commented):
this looks good!

one question, would this have any impact re: compatibility during the dev lifecycle? For example, if a project is using typescript 4, will they still be able to include the client SDK without having to also upgrade to typescript 5?

davidzhao (approved):

PR #799, opened 2023-07-20
Stop tracks if publishing fails with `setTrackEnabled`

### LINKED ISSUE #797: LocalParticipant.setCameraEnabled() leaks stream if publish fails
### Describe the bug

When calling LocalParticipant.setCameraEnabled(), the stream is leaked (ie. the webcam light remains on) if the setTrackEnabled() method re-enables the track but the publishing fails.

In this case, this was because LiveKit disconnected the call after it timing out from having no Internet. The camera track was automatically disabled when it disconnected, but if we try to re-enable the camera, the webcam light stays on until the tab is refreshed. In this case it was obviously a bug that we were calling that method with the call disconnected, but I would still expect the stream to not leak.

I suspect setTrackEnabled should just catch errors from failing to publish the stream and close the track again if it re-opened it?

### Reproduction

Call `LocalParticipant.setCameraEnabled()` with the livekit call in a disconnected state (hopefully this is enough given the detail above?)

### Logs

_No response_

### System Info

```shell
Livekit client JS 1.12.0, Google Chrome 114.0.5735.198
```

### Severity

annoyance

### Additional Information

_No response_

lukasIO:
fixes #797

### DIFF
--- a/.changeset/wet-dingos-drum.md
+++ b/.changeset/wet-dingos-drum.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Stop tracks if publishing fails with `setTrackEnabled`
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
nice!

PR #809, opened 2023-07-26
Debounce mediastreamtrack mute events

lukasIO:
For OBS virtual cameras (and reportedly also other devices) `mediastreamtrack` `muted` events get emitted in a regular interval, followed by `unmute` events. 
In order to avoid this resulting in a constant `pauseUpstream` -> `resumeUpstream` loop, we debounce the `mute` event so that we only pick it up if the track has been muted consistently for > 5s.

### DIFF
--- a/.changeset/brave-shirts-pay.md
+++ b/.changeset/brave-shirts-pay.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Debounce reacting to mediastreamtrack mute events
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
lgtm!

I tested a few different scenarios. I think muted handling is still needed because sometimes when the computer is put to sleep (closing MacBook cover), Safari still keeps the session and device active. to the other party, it could look like a black tile.

This is a good change to avoid flopping though.

PR #810, opened 2023-07-26
Add failureTolerance to KeyProvider options

lukasIO:
Adding `failureTolerance` as an option allows us to 
- specify how sensitive key invalidation should be (also helps with multiple tracks of the same participant not all being marked as invalid)
- setting it to `-1` just means a key never gets marked as invalid

### DIFF
--- a/.changeset/long-doors-push.md
+++ b/.changeset/long-doors-push.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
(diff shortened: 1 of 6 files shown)


SimonBrandner (approved):

cloudwebrtc (approved):
lgtm

davidzhao (approved):
lgtm

PR #811, opened 2023-07-26
Fix getDevices permissions when no kind is supplied

Talljoe:
The check for active devices for the Safari workaround skips getting permission if no kind is supplied.

### DIFF
--- a/src/room/DeviceManager.ts
+++ b/src/room/DeviceManager.ts
@@ -38,7 +38,7 @@ export default class DeviceManager {
     if (
       requestPermissions &&
       // for safari we need to skip this check, as otherwise it will re-acquire user media and fail on iOS https://bugs.webkit.org/show_bug.cgi?id=179363
-      !(isSafari() && this.hasActiveDevice(kind))
+      !(isSafari() && this.hasDeviceInUse(kind))
     ) {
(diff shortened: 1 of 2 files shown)


src/room/DeviceManager.ts
```diff
@@ -85,4 +84,10 @@ export default class DeviceManager {
 
     return device?.deviceId;
   }
+
+  private hasActiveDevice(kind?: MediaDeviceKind): boolean {
```
lukasIO, line 88: Thanks for the PR! I think this makes sense, only the naming on this one might be confusing. The check in this function does not really say anything about an `activeDevice`, but only whether or not a promise is in the map.

Talljoe: Hrmm. My understanding is the map contains devices that have been requested and the fact that they are "in use" in some fashion makes Safari unhappy. Is the objection that the devices might have been requested but not active? I'm trying to capture a name that explains why a device in the map is significant (for Safari).

Talljoe: That makes sense. I'll come up with a new name.

lukasIO (approved):
Thanks, looks good to me now!

PR #812, opened 2023-07-27
Add support for server injected frame trailer being passed down

lukasIO:


src/e2ee/worker/FrameCryptor.ts
```diff
@@ -65,22 +65,21 @@ export class FrameCryptor extends BaseFrameCryptor {
...
   }) {
     super();
     this.sendCounts = new Map();
     this.keys = opts.keys;
     this.participantId = opts.participantId;
     this.rtpMap = new Map();
     this.keyProviderOptions = opts.keyProviderOptions;
-    this.unencryptedFrameByteTrailer =
-      opts.unencryptedFrameBytes ?? new TextEncoder().encode('LKROCKS');
+    this.sifTrailer = opts.sifTrailer ?? new TextEncoder().encode('LKROCKS');
```
boks1971, line 82: do we still need (or want) this default?

src/e2ee/worker/FrameCryptor.ts
```diff
@@ -605,8 +614,14 @@ export enum NALUType {
...
+  if (trailerBytes.byteLength === 0) {
+    return false;
+  }
   const frameTrailer = new Uint8Array(
     frameData.slice(frameData.byteLength - trailerBytes.byteLength),
   );
+  if (trailerBytes.every((value, index) => value === frameTrailer[index])) {
+    workerLogger.info('frame is server injected');
+  }
   return trailerBytes.every((value, index) => value === frameTrailer[index]);
```
boks1971: looks like this can be done done and set a local boolean and log if the boolean is true?

src/e2ee/worker/FrameCryptor.ts
```diff
@@ -262,8 +261,10 @@ export class FrameCryptor extends BaseFrameCryptor {
       // skip for decryption for empty dtx frames
       encodedFrame.data.byteLength === 0 ||
       // skip decryption if frame is server injected
-      isFrameServerInjected(encodedFrame.data, this.unencryptedFrameByteTrailer)
+      isFrameServerInjected(encodedFrame.data, this.sifTrailer)
     ) {
+      // TODO when a frame is detected as being server injected, it would be preferable to construct
+      // an empty frame client-side instead of just passing it to the controller
```
boks1971, line 267: I think we have to strip the trailer before passing it on. Not sure if that is what you meant by constructing empty frame on client side? If we do not strip off, the decoder might barf. For example, server injects an opus silence frame on mute. If it has extra bytes when it hits the decoder, not sure if decoder will do the right thing.

boks1971 (approved):
LGTM! Some minor comments.

davidzhao (approved):
lgtm!

src/e2ee/worker/FrameCryptor.ts
lukasIO: we don't, thanks for catching this. Will replace it with an empty trailer that gets ignored.

PR #813, opened 2023-07-27
Override server injected frame data with empty array buffer

lukasIO:
based on #812 

this makes sure that the data contents of server injected frames are discarded and replaced with an empty array buffer of the same length client side.

### DIFF
--- a/src/e2ee/worker/FrameCryptor.ts
+++ b/src/e2ee/worker/FrameCryptor.ts
@@ -261,8 +261,6 @@ export class FrameCryptor extends BaseFrameCryptor {
       // skip for decryption for empty dtx frames
       encodedFrame.data.byteLength === 0
     ) {
-      // TODO when a frame is detected as being server injected, it would be preferable to construct
-      // an empty frame client-side instead of just passing it to the controller
       return controller.enqueue(encodedFrame);
     }
     if (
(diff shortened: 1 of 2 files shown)


src/e2ee/worker/FrameCryptor.ts
```diff
@@ -259,12 +259,15 @@ export class FrameCryptor extends BaseFrameCryptor {
...
+    ) {
+      return controller.enqueue(encodedFrame);
+    }
+    if (
+      // replace frame data with empty data if frame is server injected
       isFrameServerInjected(encodedFrame.data, this.sifTrailer)
     ) {
-      // TODO when a frame is detected as being server injected, it would be preferable to construct
-      // an empty frame client-side instead of just passing it to the controller
+      encodedFrame.data = new ArrayBuffer(encodedFrame.data.byteLength);
```
boks1971, line 270: while it is good to not trust un-decryptable frames, an empty buffer will not satisfy the intended purpose of those injected frames which is clear out decoder to a quiescent state. Should we say those features won't work when doing e2ee? Also, probably should strip out the trailer bytes from length as server is adding on those trailer bytes and they are in addition to the original intended length.

lukasIO: good point about stripping the trailer length! 

can you elaborate on why an empty frame data wouldn't clear out the decoder?

boks1971: These are handling RTP packets right (sorry I keep forgetting that)? The injected frame is an encoded opus silence frame or VP8/H.264 black frame. The data is in encoded format. If we give zeros to the decoder, it probably will not be decodable.

lukasIO:
closing in favour of #825

PR #815, opened 2023-07-28
Avoid de-structuring null promise

davidzhao:
I saw an exception logs while running stress testing. It threw an exception during this stack


lukasIO (approved):
nice catch! 

what does this indicate? that we call `onLocalTrackPublished` twice for a given `cid` or that the track resolver was never set in the first place? both sound unexpected.

davidzhao:
> nice catch!
> 
> Did you happen to see how this happened? Did we call `onLocalTrackPublished` twice for a given `cid` or that the track resolver was never set in the first place? both sound unexpected.

I'm not quite sure.. this is during an e2e run of one of the stress tests. and I have not been able to reproduce.

PR #816, opened 2023-07-30
Fix missing ScalabilityMode import when using SVC

### LINKED ISSUE #814: ScalabilityMode is not defined
### Describe the bug

this show up when upgraded from 1.12.1 to 1.12.2

### Reproduction

ScalabilityMode is not defined

### Logs

_No response_

### System Info

```shell
livekit client js: 1.12.2
```

### Severity

annoyance

### Additional Information

_No response_

davidzhao:
also fixed protobuf type mismatch when a float is put in place of uint32

closes #814

### DIFF
--- a/src/room/track/LocalVideoTrack.ts
+++ b/src/room/track/LocalVideoTrack.ts
@@ -491,9 +491,9 @@ export function videoLayersFromEncodings(
       layers.push(
         new VideoLayer({
           quality: VideoQuality.HIGH - i,
-          width: width / 2 ** i,
-          height: height / 2 ** i,
-          bitrate: encodings[0].maxBitrate ? Math.round(encodings[0].maxBitrate / 3 ** i) : 0,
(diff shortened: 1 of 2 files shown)


src/room/track/LocalVideoTrack.ts
```diff
@@ -483,15 +484,16 @@ export function videoLayersFromEncodings(
   if (svc) {
     // svc layers
     /* @ts-ignore */
-    const sm = new ScalabilityMode(encodings[0].scalabilityMode);
+    const encodingSM = encodings[0].scalabilityMode as string;
+    const sm = new ScalabilityMode(encodingSM);
```
lukasIO, line 488: I wish `ts-ignore` or `ts-expect-error` would allow for specifying the exact error you want to surpress so that we could avoid these issues in the future...

lukasIO (approved):

PR #817, opened 2023-07-30
Adjust bitrates following WebRTC bitrate guide

davidzhao:
We were previously using incorrect bitrates for 360p and 540p layers

See https://livekit.io/webrtc/bitrate-guide


lukasIO (commented):
lgtm!

lukasIO (approved):

PR #825, opened 2023-08-03
Drop frames if more than MAX_SIF_COUNT consecutive SIFs come in

lukasIO:
based on #812

### DIFF
--- a/src/e2ee/constants.ts
+++ b/src/e2ee/constants.ts
@@ -44,3 +44,4 @@ export const KEY_PROVIDER_DEFAULTS: KeyProviderOptions = {
 } as const;
 
 export const MAX_SIF_COUNT = 100;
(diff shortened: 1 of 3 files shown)


boks1971 (approved):
A couple of things
- Should an event be emitted when the limit is reached?
- Should we also add a time limit for SIF in addition to frame count? To protect against slow injection. Can't think of why someone would slow inject, but just thinking of ways to bolster this more. Could be something like MAX_SIF_DURATION (like 2 seconds) and whichever happens early, would start dropping.

lukasIO:
good points!

I would hold off with an event until we can be sure that we're not triggering any false positives. I think at this stage a warning log and dropping the frames seem sufficient, but agree that an event would eventually be nice.

PR #826, opened 2023-08-03
Add getRTCStatsReport method to local and remote track classes

tab1293:


lukasIO:
hi @tab1293 , 
thanks for the PR! 

We expose `getSenderStats` and `getReceiverStats` already with a stripped down version of RTCStats. 
Are you primarily interested in some specific stats that these APIs currently do not expose or is the main motivation for this PR to have a way to retrieve _all_ stats ?

tab1293:
@lukasIO hey sorry for the late response here. But yes the primary motivation of the PR is to have a way to retrieve all stats.

lukasIO:
makes sense, I think that's a good addition. Could you sign the CLA, so that I can merge this in ?

tab1293:
@lukasIO sorry CLA is signed now!

lukasIO (approved):

PR #827, opened 2023-08-04
Fix setTrackMute ignored simulcast-codec

cnderrauber:


lukasIO (approved):
lgtm, we probably need to handle this for `pauseUpstream` and `resumeUpstream`, too?

cnderrauber:
> lgtm, we probably need to handle this for `pauseUpstream`, `resumeUpstream` and `setProcessor`, too?

If those methods have disable/enable the mediaStreamTrack, we need this for them too.

PR #833, opened 2023-08-09
Support AudioTrack.setVolume for react-native

davidliu:


lukasIO (approved):
lgtm! would the `getVolume` method need a similar condition if `elementVolume` hasn't been set previously?

davidliu:
Ah yeah, good point. No equivalent getVolume method, but I think it defaults to 1 anyways.

PR #834, opened 2023-08-10
Add websocketTimeout to RoomConnectOptions

lukasIO:


src/api/SignalClient.ts
```diff
@@ -224,9 +217,16 @@ export class SignalClient {
     return new Promise<JoinResponse | ReconnectResponse | void>(async (resolve, reject) => {
       const abortHandler = async () => {
         this.close();
+        clearTimeout(wsTimeout);
         reject(new ConnectionError('room connection has been cancelled (signal)'));
       };
 
+      console.log('timeout', opts.websocketTimeout);
```
davidzhao, line 224: console to remove

src/connectionHelper/checks/websocket.ts
```diff
@@ -17,6 +17,7 @@ export class WebSocketCheck extends Checker {
       autoSubscribe: true,
       maxRetries: 0,
       e2eeEnabled: false,
+      websocketTimeout: 15_000,
```
davidzhao, line 20: should it just use default?

davidzhao (approved):
lgtm!

src/connectionHelper/checks/websocket.ts
lukasIO: the reason for it being explicit here is that the default opts get applied on the room level, not on the client level. So the client API requires it to be present explicitly

src/api/SignalClient.ts
lukasIO: thanks!

PR #836, opened 2023-08-14
Add vp9 support for E2EE

lukasIO:
Unfortunately av1 fails with `TypeError: Destination stream closed` when trying to decode frames. This needs further investigation. 
Until then, VP9 addition is nice, too!

~~Caveats: this is currently only secure if `backupCodec: undefined` is specified. Backup codecs are currently sent unencrypted, which is definitely unexpected. ~~
^ backup codecs are currently disabled in general when e2ee is enabled


lukasIO:
closing for now due to the backupCodec complications

Stogas:
@lukasIO 

Could you elaborate a little bit?

I'd still consider this open, as VP9 support is still wanted and missing.

lukasIO:
hi @Stogas, 

encryption support for backup codecs is missing, which would be needed to ensure that also clients that don't support vp9 would be able to receive the media stream. 

I can reopen the issue, but there's currently no ETA for encryption support of backup codecs

davidzhao:
We should disable backupCodec when e2ee is used. since vp9 is supported everywhere, I think the impact of that should be fairly minimal

lukasIO:
> We should disable backupCodec when e2ee is used.

That's what we do already today. 
I'll update this PR in order to merge it in

davidzhao (approved):
awesome!

PR #838, opened 2023-08-16
Fix maxJoinAttempts in RTC Engine

mpnri:
Hi.
We have a `maxRetries` property in `RoomConnectOptions`,  but it seems we never use it to retry the connection. On the other hand, we have a `maxJoinAttempts` value in RTCEngine which is hard-coded and cannot be changed by room connect options.
This PR allows to set the `maxJoinAttempts` value through the `maxRetries` property in `RoomConnectOptions`.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -191,6 +191,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     this.url = url;
     this.token = token;
     this.signalOpts = opts;
+    this.maxJoinAttempts = opts.maxRetries;
     try {
       this.joinAttempts += 1;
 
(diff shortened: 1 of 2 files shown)


src/room/RTCEngine.ts
```diff
@@ -191,6 +191,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     this.url = url;
     this.token = token;
     this.signalOpts = opts;
+    this.maxJoinAttempts = opts.maxRetries;
```
lukasIO, line 194: thanks for spotting this! 
It seems like only changing this line would be sufficient though?

mpnri: Yeah. You are right.
Fixed

lukasIO (approved):

PR #840, opened 2023-08-17
Always apply participant update, also on existing participants

lukasIO:


davidzhao (approved):
change looks good. though it's strange this is happening.. there is code in place to ensure we first send metadata before the offer comes down.. maybe there's a regression there.

lukasIO:
do we still want to merge this, even though the issue is probably mitigated with https://github.com/livekit/livekit/pull/1974 ?

davidzhao:
> do we still want to merge this, even though the issue is probably mitigated with [livekit/livekit#1974](https://github.com/livekit/livekit/pull/1974) ?

hmm your call on this one.. on one hand, it makes the client more resilient.. on the other, it has a side effect of "hiding" real issues.

lukasIO:
My take on this is to keep the behaviour as is in order to surface/notice the issues. 

But I'm thinking we should then remove the path for tracks to arrive first altogether, right?

davidzhao:
Yeah I agree.. I don't think we should have that path anymore.

lukasIO:
superseded by #854

PR #841, opened 2023-08-18
Ensure play requests are invoked synchronously in startAudio

lukasIO:
The idea is to aquire audio context in parallel with the play requests. This is done in order to not invoking an "awaited" function before the the play requests are invoked. Some browsers (e.g. iOS Safari) raise an exception if the play request isn't invoked directly from the user gesture. An `await` in the callback handler seems to mess with that. 

Tested this across Desktop browsers and iOS Safari with both `expWebAudioMix` and without.


davidzhao (approved):
lgtm!

PR #842, opened 2023-08-18
Fix track processor blips when restarting tracks

lukasIO:
Only works with processors with this PR https://github.com/livekit/track-processors-js/pull/11

supersedes #739 

@burzomir thank you for all the investigations you did around this, could you test that combination (this branch and the processor main branch) to see if it solves your use case?

### DIFF
--- a/.changeset/serious-boxes-train.md
+++ b/.changeset/serious-boxes-train.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Fix track processor blips when restarting tracks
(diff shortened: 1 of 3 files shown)


davidzhao (approved):

burzomir:
Thanks, @lukasIO, but it doesn't work at all now. LocalVideoTrack is displayed as a black video element; other participants no longer receive pictures. What I do is:
```
await camera.setProcessor(blurProcessor, true)
await camera.restartTrack({ facingMode: 'user' })
```

I'll try to create an example app later.

lukasIO:
@burzomir are you using track processors with the PR that I linked or the npm version?

burzomir:
@lukasIO  I rechecked, and it seems to do the trick. I tested it in a simple example app I created (setProcessor and then restart), and it worked without issues using BackgroundBlur.

PR #843, opened 2023-08-22
Fix track is not ended correctly with new stream id

cnderrauber:


src/room/track/RemoteTrack.ts
```diff
@@ -29,14 +29,16 @@ export default abstract class RemoteTrack extends Track {
...
-    // we send each track down in its own MediaStream, so we can assume the
-    // current track is the only one that can be removed.
     this.mediaStream = stream;
-    stream.onremovetrack = () => {
-      this.receiver = undefined;
-      this._currentBitrate = 0;
-      this.emit(TrackEvent.Ended, this);
+    const onRemoveTrack = (event: MediaStreamTrackEvent) => {
+      if (event.track === this._mediaStreamTrack) {
+        stream.removeEventListener('removetrack', onRemoveTrack);
```
boks1971, line 35: Will this `removeEventListener` affect some other track attached to this stream?

cnderrauber: By the document and my test, it only removes the current `event listener` so it will not remove other handles. CC @lukasIO

lukasIO: this looks good to me!

lukasIO (approved):

PR #844, opened 2023-08-22
Allow ArrayBuffers to be used for ExternalKeyProvider

lukasIO:
This adds a `createKeyMaterialFromBuffer` helper method and allows the `keyProvider.setKey` method to accept a `ArrayBuffer` in addition to a string. 

The string is used for pass phrases and uses `PBKDF2`. (not recommended, but easy to use)
The buffer is used for [cryptographically strong random numbers](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) and uses `HKDF`. (recommended, but needs crypto rand values instead of a easily type-able pass phrase)

### DIFF
--- a/.changeset/hip-brooms-jam.md
+++ b/.changeset/hip-brooms-jam.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
lgtm!

PR #846, opened 2023-08-22
Add ceil of width and height when emitting UpdateTrackSettings

### LINKED ISSUE #845: When adaptiveStream.pixelDensity = screen, and devicePixelRatio is with fractions, "SignalClient.ts:551 error sending signal message" is thrown
### Describe the bug

I run my livekit-client with `adaptiveStream.pixelDensity = 'screen'`. However, with latest release of livekit we noticed that on some devices, the stream did not always start. When debugging, we discovered that lk was throwing "error sending signal message" into console, with following message: `invalid uint 32: 363.0000078678131`.

Downgrading to version 1.12.1 fixes the issue. I think it's related to changes in #700, but ultimately I think it's RemoteTrackPublication.emitTrackUpdate() that should ceil both width and height before passing them to UpdateTrackSettings.

### Reproduction

1. Clone this repo, install dependencies
2. Replace line example/sample.ts:74 with `const adaptiveStream = { pixelDensity: 'screen' as const };`
3. run "sample" script from package.json
4. Open 2 tabs, connect to same room on both tabs
5. change devicePixelRatio to be a number with fractions. For example, zoom the page.
6. Observe browser console

### Logs

_No response_

### System Info

```shell
Not relevant
```

### Severity

blocking an upgrade

### Additional Information

_No response_

HermanBilous:
Fixes #845

### DIFF
--- a/.changeset/tidy-meals-double.md
+++ b/.changeset/tidy-meals-double.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Add ceil of width and height when emitting UpdateTrackSettings
--- a/src/room/track/RemoteTrackPublication.ts
+++ b/src/room/track/RemoteTrackPublication.ts
@@ -297,8 +297,8 @@ export default class RemoteTrackPublication extends TrackPublication {
       fps: this.fps,
     });
     if (this.videoDimensions) {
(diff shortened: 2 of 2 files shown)


lukasIO (approved):
awesome, thanks!

PR #850, opened 2023-08-23
Make KeyProvider and ParticipantKeyHandler work consistently for shared key and sender key scenarios (e2ee)

lukasIO:
main motivation for these fixes was an inconsistency in setting the local participant's keys, which expected the participantId to be `undefined`. This caused a bug when trying to set individual sender keys instead of using a shared-key provider.
Initially this was done because when `setKey` on a KeyProvider was called, we had no way of figuring out whether or not that participant is actually the local participant, until we're connected to the server and have the participant identity info. 

This PR
- replaces strings with dedicated event enums in e2ee event emitters
- introduces a `enabledEncryptionMap` that tracks whether or not encryption is enabled for a specific participant (previously this was handled as part of the `ParticipantKeyHandler`)
- updates local participant encryption keys on `SignalConnected` (the earliest we know the identity without manually parsing the token)
- introduces a dedicated `initAck` worker message in order not to have to mis-use the existing `enable` message for that purpose
- uses a dedicated `sharedKeyHandler` that handles all key interactions when `sharedKey` is set to true. Previously an instance was created for each participant, but in the shared key scenario that makes less sense. the only downside is that ratcheting doesn't work independently now, but I don't really see a reason in the shared key scenario for ratcheting anyways.

### DIFF
--- a/src/e2ee/E2eeManager.ts
+++ b/src/e2ee/E2eeManager.ts
@@ -81,9 +81,9 @@ export class E2EEManager extends (EventEmitter as new () => TypedEventEmitter<E2
   /**
    * @internal
    */
-  setParticipantCryptorEnabled(enabled: boolean, participantId: string) {
-    log.debug(`set e2ee to ${enabled} for participant ${participantId}`);
-    this.postEnable(enabled, participantId);
+  setParticipantCryptorEnabled(enabled: boolean, participantIdentity: string) {
+    log.debug(`set e2ee to ${enabled} for participant ${participantIdentity}`);
+    this.postEnable(enabled, participantIdentity);
   }
 
   /**
@@ -115,7 +115,7 @@ export class E2EEManager extends (EventEmitter as new () => TypedEventEmitter<E2
       case 'enable':
         if (
           this.encryptionEnabled !== data.enabled &&
-          data.participantId === this.room?.localParticipant.identity
+          data.participantIdentity === this.room?.localParticipant.identity
         ) {
           this.emit(
             EncryptionEvent.ParticipantEncryptionStatusChanged,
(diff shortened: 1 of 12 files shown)


src/e2ee/E2eeManager.ts
```diff
@@ -115,19 +101,35 @@ export class E2EEManager extends (EventEmitter as new () => TypedEventEmitter<E2
...
+            this.postKey(keyInfo);
+          });
+        }
         break;
+
       case 'enable':
-        if (this.encryptionEnabled !== data.enabled && !data.participantId) {
+        if (
+          this.encryptionEnabled !== data.enabled &&
+          data.participantId === this.room?.localParticipant.identity
```
davidzhao, line 118: is data.participantId identity? should we rename it to be clear?

src/room/Room.ts
```diff
@@ -209,10 +209,10 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
    */
   async setE2EEEnabled(enabled: boolean) {
     if (this.e2eeManager) {
-      await Promise.all([
-        this.localParticipant.setE2EEEnabled(enabled),
-        this.e2eeManager.setParticipantCryptorEnabled(enabled),
-      ]);
+      await Promise.all([this.localParticipant.setE2EEEnabled(enabled)]);
+      if (this.localParticipant.identity !== '') {
```
davidzhao, line 213: good to be safe here. could this this be true?

davidzhao (approved):
lg!

src/room/Room.ts
lukasIO: `localParticipant.identity` gets initialized with an empty string on room creation. As long as we don't know the actual identity, it doesn't make sense to enable the cryptor. We do however use the most recent setting for `localParticipant.isE2EEEnabled` on `SignalConnected` now so that we enable the cryptor once we know the local participant's identity.

PR #867, opened 2023-09-08
Update analyser cleanup to async function

kand193:
Because `AudioContext.close` function is asynchronous, I apply async/await to `cleanup` in `AudioAnalyser`.


lukasIO (approved):
thanks for spotting (and fixing) this!

PR #869, opened 2023-09-12
Add publish option to force a specific resolution on the sender

lukasIO:


src/room/participant/publishUtils.ts
```diff
@@ -134,14 +149,15 @@ export function computeVideoEncodings(
...
+        maxBitrate: Math.ceil(videoEncoding.maxBitrate / 3 ** i),
         /* @ts-ignore */
         maxFramerate: original.encoding.maxFramerate,
+        scaleResolutionDownBy: 2 ** i * originalLayerScale, // FIXME this currently freezes the downstream track on the receiver side
       });
     }
     /* @ts-ignore */
     encodings[0].scalabilityMode = scalabilityMode;
-    log.debug('encodings', encodings);
+    log.info('encodings', encodings);
```
davidzhao, line 160: intentional?

src/room/participant/publishUtils.ts
```diff
@@ -134,14 +149,15 @@ export function computeVideoEncodings(
     for (let i = 0; i < sm.spatial; i += 1) {
       encodings.push({
         rid: videoRids[2 - i],
-        maxBitrate: videoEncoding.maxBitrate / 3 ** i,
+        maxBitrate: Math.ceil(videoEncoding.maxBitrate / 3 ** i),
         /* @ts-ignore */
         maxFramerate: original.encoding.maxFramerate,
+        scaleResolutionDownBy: 2 ** i * originalLayerScale, // FIXME this currently freezes the downstream track on the receiver side
```
davidzhao, line 155: I think with SVC this would not work.. we'd have to keep the stream's original resolution

src/room/participant/publishUtils.ts
```diff
@@ -100,22 +100,37 @@ export function computeVideoEncodings(
...
+    return [{ scaleResolutionDownBy: originalLayerScale }];
   }
 
+  const adjustedWidth = width / originalLayerScale;
+  const adjustedHeight = height / originalLayerScale;
+
   if (!videoEncoding) {
     // find the right encoding based on width/height
-    videoEncoding = determineAppropriateEncoding(isScreenShare, width, height, videoCodec);
+    videoEncoding = determineAppropriateEncoding(
```
davidzhao, line 121: since we can't change this behavior for SVC, I think we'd have to use original width/height if it was SVC.

src/room/track/LocalVideoTrack.ts
```diff
@@ -340,6 +347,22 @@ export default class LocalVideoTrack extends LocalTrack {
     await setPublishingLayersForSender(this.sender, this.encodings, qualities, this.senderLock);
   }
 
+  protected async setMediaStreamTrack(newTrack: MediaStreamTrack, force?: boolean) {
+    if (this.sender && this.forcedResolution) {
+      const params = this.sender.getParameters();
+      const dims = await this.waitForDimensions(undefined, newTrack);
+      const scaleFactor = Math.max(1, dims.height / this.forcedResolution.height);
+      const originalEncoding = params.encodings[params.encodings.length - 1];
```
davidzhao, line 355: would we have to change all of them instead of just the last one?

lukasIO: that would override the settings that users set on the simulcast layers, I think it would be preferable to only force the resolution on the original one.

lukasIO:
converting to a draft, until we know what to do about SVC codecs

PR #871, opened 2023-09-13
Fix callback loop with metadata updates

davidzhao:
When canUpdateOwnMetadata grant is given to the participant, and metadata is updated remotely, it would throw the client into an invite loop in updating metadata with previous and current value.

This was really tricky to root cause. The way it happened was:
* RoomService.UpdateParticipant is called
* Participant.updateInfo called with new metadata
* LocalParticipant.setName called with the same name
  * setName triggers a `sendUpdateLocalMetadata` with previous metadata
* LocalParticipant.setMetadata called with new metadata
  * setMetadata triggers a `sendUpdateLocalMetadata` with new metadata
* Participant.updateInfo called with old metadata....

### DIFF
--- a/.changeset/violet-ants-help.md
+++ b/.changeset/violet-ants-help.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Fix infinite metadata loop when canUpdateOwnMetadata is granted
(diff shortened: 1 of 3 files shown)


lukasIO (approved):
great catch!

PR #874, opened 2023-09-14
Ensure publisher connection is reconnected if needed

lukasIO:


src/room/RTCEngine.ts
```diff
@@ -200,13 +200,14 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
 
+      this.pcState = PCState.New;
       this.subscriberPrimary = joinResponse.subscriberPrimary;
       if (!this.publisher) {
         this.configure(joinResponse);
       }
 
       // create offer
-      if (!this.subscriberPrimary) {
+      if (!this.subscriberPrimary || this.hasPublished) {
```
lukasIO, line 210: this is needed to re-establish publisher connection on full reconnects

src/room/RTCEngine.ts
```diff
@@ -1092,11 +1105,10 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
         // this means we'd have to check its status manually and update address
         // manually
         now - startTime > minReconnectWait &&
-        this.primaryPC?.connectionState === 'connected'
+        this.primaryPC?.connectionState === 'connected' &&
+        (!this.hasPublished || this.publisher?.pc.connectionState === 'connected')
```
lukasIO, line 1109: we only set it to connected if - in case we need it - the publisher connection is also connected

src/room/RTCEngine.ts
```diff
@@ -415,16 +418,17 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     primaryPC.onconnectionstatechange = async () => {
       log.debug(`primary PC state changed ${primaryPC.connectionState}`);
       if (primaryPC.connectionState === 'connected') {
-        const shouldEmit = this.pcState === PCState.New;
-        this.pcState = PCState.Connected;
-        if (shouldEmit) {
+        const initialFullConnection =
+          (this.pcState === PCState.New && !this.hasPublished) ||
+          this.publisher?.pc.connectionState === 'connected';
```
davidzhao, line 423: I'm not sure if this is right.. if publisher is connected before the subscriber in the reconnect case.. it should not emit this event.

I think the idea is that this would only emit event in case it's the initial connection, and reconnect would be checking its state/events in either the resume or full reconnect paths.

davidzhao: what is this change attempting to fix? it seems to increase complexity as we try to coordinate both peer connection states.

src/room/Room.ts
```diff
@@ -1119,9 +1119,12 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       // reconnection failed, handleDisconnect is being invoked already, just return here
       return;
     }
-    this.setAndEmitConnectionState(ConnectionState.Connected);
-    this.emit(RoomEvent.Reconnected);
-    this.registerConnectionReconcile();
+    // only set as connected if _only_ the signal connection had been severed
```
davidzhao, line 1122: not understanding the comment. it seems to be flipping it back only when PC transport have been restored.

src/room/RTCEngine.ts
lukasIO: you're right, we'd have to make sure not to emit the reconnect event.

the idea is to have a single understanding of what a connected state is. If `subscriberPrimary` but we also publish something, the connected state only happens if _both_ peer connections are connected

src/room/Room.ts
lukasIO: maybe bad wording. 
At this point we would only want to emit `Reconnected` if the transports still work. In case the transport don't we would wait for them to re-establish until a `Reconnected` event is fired.

PR #877, opened 2023-09-18
fix: fix h264 freeze when cipherText contains more than 2 consecutive zeros.

cloudwebrtc:


src/e2ee/worker/FrameCryptor.ts
```diff
@@ -243,17 +243,22 @@ export class FrameCryptor extends BaseFrameCryptor {
           new Uint8Array(encodedFrame.data, this.getUnencryptedBytes(encodedFrame)),
         );
 
-        const newData = new ArrayBuffer(
-          frameHeader.byteLength + cipherText.byteLength + iv.byteLength + frameTrailer.byteLength,
+        var newDataWithoutHeader = new Uint8Array(
```
lukasIO, line 246: nit: `let` instead of `var`

src/e2ee/worker/FrameCryptor.ts
```diff
@@ -553,6 +592,58 @@ export class FrameCryptor extends BaseFrameCryptor {
...
 }
 
+export function needsRbspUnescaping(frameData: Uint8Array) {
+  for (var i = 0; i < frameData.length - 3; i++) {
+    if (frameData[i] == 0 && frameData[i + 1] == 0 && frameData[i + 2] == 3) return true;
+  }
+  return false;
+}
+
+export function ParseRbsp(stream: Uint8Array): Uint8Array {
```
lukasIO, line 602: let's maybe move the `parse` and `write` functions (and the constants) to `e2ee/utils.ts`?

src/e2ee/worker/FrameCryptor.ts
```diff
@@ -493,6 +515,23 @@ export class FrameCryptor extends BaseFrameCryptor {
     return iv;
   }
 
+  private frameIsH264(frame: RTCEncodedVideoFrame): boolean {
+    let detectedCodec = this.getVideoCodec(frame) ?? this.videoCodec;
+    const data = new Uint8Array(frame.data);
+    try {
+      const naluIndices = findNALUIndices(data);
```
lukasIO, line 522: wondering if we can re-use the `findNALUIndices` that we do as part of `getUnencryptedBytes` already so we don't have to compute them twice for each frame...

lukasIO (commented):
Great find, just some minor comments.

lukasIO (approved):
nice!

PR #882, opened 2023-09-28
Fix vp9 svc encode failed for screenshare

cnderrauber:
VP9 SVC for screenshare has problem to encode with error `Failed to initialize the encoder associated with codec type: VP9 (2)`, use L1T3 mode for screenshare video track to fix it.

### DIFF
--- a/.changeset/silver-swans-build.md
+++ b/.changeset/silver-swans-build.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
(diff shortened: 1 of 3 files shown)


davidzhao:
Does AV1 have the same issue?

davidzhao (approved):

cnderrauber:
@davidzhao AV1 works well for screen share with L3T3 mode.

PR #889, opened 2023-10-12
Do not constrain screenshare resolution by default

davidzhao:
This will allow us to capture at the original resolution and avoid resizing. It also helps to workaround a Safari 17 bug where setting a resolution constraint leads to a 480p capture.

https://bugs.webkit.org/show_bug.cgi?id=263015

### DIFF
--- a/.changeset/fast-cows-melt.md
+++ b/.changeset/fast-cows-melt.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Do not constrain screenshare resolution by default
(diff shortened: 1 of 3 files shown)


.changeset/fast-cows-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+'livekit-client': patch
```
lukasIO, line 2:
```suggestion
'livekit-client': minor
```

lukasIO (approved):
lgtm!

PR #894, opened 2023-10-18
Do not allow VP9 publishing on FF

davidzhao:
Since FF does not support SVC, and we require SVC for VP9, it will not be able to support VP9 publishing.

Previously FF didn't support `getCapabilities`, so supportsVP9 check was passing for it. As much as I hate browser checks, there isn't another way to check for SVC publishing.


lukasIO (approved):
lgtm!

PR #898, opened 2023-10-23
Fix Safari reporting wrong initial track resolution

lukasIO:


src/room/track/LocalTrack.ts
```diff
@@ -163,6 +163,12 @@ export default abstract class LocalTrack extends Track {
       throw new Error('cannot get dimensions for audio tracks');
     }
 
+    if (getBrowser()?.os === 'iOS') {
```
davidzhao, line 166: wow... just.... 🤯

davidzhao (approved):

PR #900, opened 2023-10-25
Fix wrong parameter type in the TypeDoc of TrackSubscriptionPermissionChanged event

pabloFuente:


lukasIO (approved):
thanks for fixing this!

PR #903, opened 2023-10-25
Make peerconnection private on PCTransport

lukasIO:
First preparation step for connection handling refactor. 
Aim of this PR is to make access to the underlying PC of PCTransport explicit by hiding the PC object and exposing dedicated methods for access.


davidzhao (approved):
lg!

PR #904, opened 2023-10-25
Add key index to log lines

dbkr:
There were a couple of places that were phrased as if they intended to log the relevant key index but didnt, and one where it's also just useful.


lukasIO (approved):
thanks, this is great!

Issue #897, opened 2023-10-22
createLocalTracks and mergeDefaultOptions doesn't clone deeply


vas11yev1work:
### Describe the bug

`mergeDefaultOptions` inside `createLocalTracks` gets `options` as a parameter and change it. It happens because `options` can have nested objects. But in `mergeDefaultOptions` you do the following:

```typescript
const opts: CreateLocalTracksOptions = {
  ...options,
};
``` 

Is it intended or does it need to be fixed?

### Reproduction

You can see it here https://codesandbox.io/s/cocky-ritchie-ry68d9

In App.vue file

I created a `previewOptions` computed and pass it in `createLocalTracks`

### System Info

```shell
Windows 11 22621
x64 12th Gen Intel(R) Core(TM) i5-12500H
16 GB RAM
Node: 18.18.2
Yarn: 1.22.19
npm: 9.8.1
Chrome: 118.0.5993.89
"livekit-client": "^1.14.1"
```

### Severity

serious, but I can work around it

lukasIO:
Thanks for the report! I think we should fix this. 
Do you want to open a PR that serializes and de-serializes the `CreateLocalTrackOptions`?

vas11yev1work:
@lukasIO yes, I do
But what do you think is the best way to do it? I think to create a deepClone function (or this function has already been created). Or just use JSON namespace. But I don't sure about correctness and performance of this solution. Maybe use lodash or something similar?
I haven't delved into the specifics of the project, so I don't know how to do it better

lukasIO:
I think for this case, the best way would be to use `structuredClone` if it's available in the browser and fallback to `JSON.parse(JSON.stringify())` if it's not available. 
If you do this, it would be great if you could add a unit test for it as well! 
https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

lukasIO:
closed via #906, thanks @vas11yev1work !

PR #906, opened 2023-10-25
Add clone deep

vas11yev1work:
Fixes issue #897

### DIFF
--- a/src/utils/cloneDeep.test.ts
+++ b/src/utils/cloneDeep.test.ts
@@ -1,0 +1,54 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { cloneDeep } from './cloneDeep';
+
+describe('cloneDeep', () => {
+  beforeEach(() => {
+    global.structuredClone = vi.fn((val) => {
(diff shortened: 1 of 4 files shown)


src/utils/cloneDeep.test.ts
```diff
@@ -0,0 +1,54 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { cloneDeep } from './cloneDeep';
+
+describe('cloneDeep', () => {
+  beforeEach(() => {
+    global.structuredClone = vi.fn((val) => {
+      return JSON.parse(JSON.stringify(val));
+    });
+  });
```
vas11yev1work, line 9: It looks strange, but apparently this is the only workaround
https://github.com/jsdom/jsdom/issues/3363

lukasIO: huh, well, then the unit test is exclusively testing for JSON serialization 😅 
Thanks for looking into it!

lukasIO (approved):
This looks good! Thanks for the PR

PR #908, opened 2023-10-26
Fix a race in setKeyFromMaterial

dbkr:
this.currentKeyIndex was set prior to await-ing, so if several keys were set in quick succession, they would all be set at the index of the last one.

This makes the function a little more complex but maintains the behaviour of setting the key at `this.currentKeyIndex` if the function is called with a negative number, although with this version it will use the value of `this.currentKeyIndex` *after* the await. An alternative fix would be to just move the setting of `this.currentKeyIndex` to after the `await`: I figured perhaps this was a little more explicit, but I don't mind.

Also expanded the log line to include the index being set.

### DIFF
--- a/.changeset/selfish-eagles-learn.md
+++ b/.changeset/selfish-eagles-learn.md
@@ -2,4 +2,4 @@
 'livekit-client': patch
 ---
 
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
nice! thanks for spotting (and fixing) this :)

PR #909, opened 2023-10-26
PCTransportManager

lukasIO:
Second phase for connection logic refactor. 
This PR introduces a PCManager. The primary idea behind it is to
- encapsulate PC related logic 
- and thus simplify logic handled in RTCEngine
- surface connection state changes, where only changes of "required" PCs are taken into account (support for pub only / sub only)


src/room/RTCEngine.ts
```diff
@@ -73,9 +73,9 @@ enum PCState {
 
 /** @internal */
 export default class RTCEngine extends (EventEmitter as new () => TypedEventEmitter<EngineEventCallbacks>) {
-  publisher?: PCTransport;
+  // publisher?: PCTransport;
 
-  subscriber?: PCTransport;
+  // subscriber?: PCTransport;
```
boks1971, line 78: would be good to delete unused stuff.

Just my personal habit: when I have long running projects (actually for most projects, but definitely for long running projects) where I know I will be thinking/churning/re-doing a ton of stuff, I tend to comment out stuff so that I know where things were before. But, I also add a marker (something like `// RAJA-REMOVE`) which allows me to clean up at the end of the project when I set up a PR (of course, there have been instances of me forgetting, but it at least helps me 99% of the times :-) )

src/room/RTCEngine.ts
```diff
@@ -247,28 +245,10 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-    if (this.subscriber) {
-      this.subscriber.close();
-      this.subscriber = undefined;
-    }
-    this.hasPublished = false;
-    this.primaryTransport = undefined;
+    await this.pcManager?.close();
+    this.pcManager = undefined;
+
+    // this.primaryTransport = undefined;
```
boks1971, line 251: unused?

src/room/RTCEngine.ts
```diff
@@ -448,42 +396,30 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-      if (!this.publisher) {
+      if (!this.pcManager) {
         return;
       }
       log.debug('received server answer', {
         RTCSdpType: sd.type,
-        signalingState: this.publisher.getSignallingState().toString(),
       });
-      await this.publisher.setRemoteDescription(sd);
+      await this.pcManager.setAnswer(sd);
```
boks1971, line 405: Is this always SDP answer?

src/room/RTCEngine.ts
```diff
@@ -1038,72 +974,29 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-      this.once(EngineEvent.Connected, onConnected);
-    });
+    await this.pcManager.ensurePCTransportConnection(abortController, timeout);
   }
 
   private async waitForPCReconnected() {
-    const startTime = Date.now();
-    let now = startTime;
+    // const startTime = Date.now();
+    // let now = startTime;
```
boks1971, line 985: unused?

src/room/RTCEngine.ts
```diff
@@ -108,7 +110,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private subscriberPrimary: boolean = false;
 
-  private primaryTransport?: PCTransport;
+  // private primaryTransport?: PCTransport;
```
boks1971, line 113: unused?

src/room/PCTransportManager.ts
```diff
@@ -0,0 +1,328 @@
...
+    subState: RTCPeerConnectionState,
+  ) => void;
+
+  public onIceCandidate?: (ev: RTCIceCandidate, target: SignalTarget) => void;
+
+  public onDataChannel?: (ev: RTCDataChannelEvent) => void;
+
+  public onTrack?: (ev: RTCTrackEvent) => void;
+
+  public onLocalOffer?: (offer: RTCSessionDescriptionInit) => void;
```
boks1971, line 49: maybe call this `onPublisherOffer` as `publisher` PC should be only one offering. Feels like that can be more explanatory.

src/room/PCTransportManager.ts
```diff
@@ -0,0 +1,328 @@
...
+  requireSubscriber(require = true) {
+    this.isSubscriberConnectionRequired = require;
+    this.updateState();
+  }
+
+  createAndSendOffer(options?: RTCOfferOptions) {
+    return this.publisher.createAndSendOffer(options);
+  }
+
+  setAnswer(sd: RTCSessionDescriptionInit) {
```
boks1971, line 108: Made a comment at the caller location. Maybe, call this `setPublisherAnswer` to make it explicit?

src/room/PCTransportManager.ts
```diff
@@ -0,0 +1,328 @@
...
+      signalingState: this.subscriber.getSignallingState().toString(),
+    });
+    await this.subscriber.setRemoteDescription(sd);
+
+    // answer the offer
+    const answer = await this.subscriber.createAndSetAnswer();
+    return answer;
+  }
+
+  updateConfiguration(config: RTCConfiguration, iceRestart?: boolean) {
```
boks1971, line 162: could there be separate configs for publisher, subscriber?

src/room/PCTransportManager.ts
```diff
@@ -0,0 +1,328 @@
...
+  addTrack(track: MediaStreamTrack) {
+    return this.publisher.addTrack(track);
+  }
+
+  createDataChannel(label: string, dataChannelDict: RTCDataChannelInit) {
+    return this.publisher.createDataChannel(label, dataChannelDict);
+  }
+
+  getConnectedAddress() {
+    return this.requiredTransports[0].getConnectedAddress();
```
boks1971, line 233: May need to have methods to get publisher and subscriber connected address too, maybe useful in testing.

src/room/PCTransportManager.ts
```diff
@@ -0,0 +1,328 @@
...
+          );
+          return;
+        }
+      }
+      CriticalTimers.clearTimeout(connectTimeout);
+      abortController?.signal.removeEventListener('abort', abortHandler);
+      resolve();
+    });
+  }
+}
```
boks1971, line 328: Maybe add `Publisher` or `Subscriber` to method names. Ended up doing that in transport manager on server side as I found it easier to understand from the caller ([example](https://github.com/livekit/livekit/blob/5ec692bb03bb2f1d6b980db7c5ae4fd52efe2017/pkg/rtc/transportmanager.go#L260)). I agree that it is implicit for several functions, but adding to the method name reduced cognitive load for me :-) . For example, in this file, I believe `addTransceiver`, `addTrack`, `removeTrack` are all on `publisher. Making them something like `addPublisherTransceiver`, `addPublisherTrack` etc. makes it very clear (at least to me :-) )

boks1971 (commented):
First look, looks good! Much easier to follow. Left some minor comments.

I will read this again. One thing I noticed is a bunch more `await`. Example
```
 -     if (!this.publisher) {
 -       this.configure(joinResponse);
 +     if (!this.pcManager) {
 +       await this.configure(joinResponse);
```
Were those places not handling things properly or the new code needs that behaviour?

src/room/PCTransport.ts
```diff
@@ -33,10 +33,16 @@ export default class PCTransport extends EventEmitter {
...
-    throw new UnexpectedConnectionState('Expected peer connection to be available');
+    if (!this._pc) {
+      this._pc = this.createPC();
+    }
+    return this._pc;
   }
 
+  private config?: RTCConfiguration;
+
+  private mediaConstraints: Record<string, unknown>;
```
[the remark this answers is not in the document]
lukasIO: `unknown` is the preferred way for handling `any` types in TS as it will force you to handle the unknown type, whereas `any` just gets accepted without the compiler complaining ([reference](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type))

src/room/RTCEngine.ts
lukasIO: This basically mirrors the way the signal callbacks are set up, so that the flow of `client.onAnswer` -> `pcManager.setAnswer` uses the same terminology. 
Do you think keeping it at `setRemoteDescription` makes more sense?

src/room/PCTransportManager.ts
lukasIO: makes sense to me. Initial thought was that which PC is being used could be an implementation detail of `pcManager`, but happy to change that if it feels more explanatory with `publisher` in there

src/room/PCTransportManager.ts
lukasIO: Previously we never had a way to set different configs either, so I assumed we want to keep the behaviour. 
Can you think of a scenario where separate configs would be useful?

src/room/PCTransportManager.ts
lukasIO: good call

src/room/RTCEngine.ts
boks1971: `setAnswer` is fine. I suggested elsewhere, doing something like `setPublisherAnswer`. That makes it clearer to me  by just reading caller code that this is on `publisher PC` and it is an `SDP answer`.

src/room/PCTransportManager.ts
boks1971: Agreed that it is implementation detail. But, we have two PCs (or in publisher only case only publisher PC and in subscriber only case only subscriber PC). We never use a single object called just `peerConnection`. So, thought about adding the extra `publisher` word in the method name to make it easier to follow from the call site itself. I am okay leaving it as `onLocalOffer`, but suggested the change given that we have two distinct things in code always.

src/room/PCTransportManager.ts
lukasIO: should have read all your comments before writing mine. Yeah, that makes sense to me naming wise. As mentioned above the thought was to hide that fact within the pcManager. With your initial feeling that it's easier to reason about with publisher/subscriber in the method names, I'm happy to do that change. (the whole purpose is to make the whole thing easier to reason about, so thanks for that feedback)

src/room/PCTransportManager.ts
boks1971: Can't think of any, but just wanted to check. We do have simulation scenarios where we change transport type. Wondering if we will have something down the line which will set different transport for different peer connection. So, was thinking about keeping it flexible. Can change if it is required down the line.

src/room/Room.ts
```diff
@@ -1551,8 +1551,11 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
   }
 
   private sendSyncState() {
-    const previousAnswer = this.engine.subscriber?.getLocalDescription();
-    const previousOffer = this.engine.subscriber?.getRemoteDescription();
+    if (!this.engine.pcManager) {
+      return;
+    }
+    const previousAnswer = this.engine.pcManager.subscriber.getLocalDescription();
+    const previousOffer = this.engine.pcManager.subscriber.getRemoteDescription();
```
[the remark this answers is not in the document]
lukasIO: True, that would be undefined for publisher only. What do you think is the right way to handle this? Fall back to publisher if subscriber is undefined?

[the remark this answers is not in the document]
boks1971: Yeah, I think it should include the publisher parts. `previousAnswer` and `previousOffer` are subscriber specific thing needed for migration. The code below returns if there is no `previousAnswer`. I think it should proceed to send the message, but just not fill out the fields that are nil/undefined. Server will see `nil` in those fields and ignore them, but will use `publish_tracks` and `data_channels` (publisher only data channels) from the message.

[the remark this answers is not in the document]
boks1971: `LocalParticipant.dataChannelsInfo()` I believe should just populate with publisher side data channels as subscriber is not there. That should be fine.

src/room/RTCEngine.ts
```diff
@@ -1038,72 +960,28 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+      await sleep(minReconnectWait); // FIXME setTimeout again not ideal for a connection critical path
+      if (!this.pcManager) {
+        throw new UnexpectedConnectionState('PC manager is closed');
       }
-      await sleep(100);
-      now = Date.now();
+      await this.pcManager.ensurePCTransportConnection(undefined, this.peerConnectionTimeout);
+      this.pcState = PCState.Connected;
+    } catch (e: any) {
+      // TODO do we need a `failed` state here for the PC?
```
davidzhao, line 981: I'm guessing we don't.

src/room/PCTransportManager.ts
```diff
@@ -0,0 +1,336 @@
...
+      transports.push(this.subscriber);
+    }
+    return transports;
+  }
+
+  private updateState = () => {
+    const previousState = this.state;
+
+    const connectionStates = this.requiredTransports.map((tr) => tr.getConnectionState());
+    if (connectionStates.every((st) => st === 'connected')) {
```
davidzhao, line 259: do we care about disconnected state at all? I know for Safari, it's not particularly reliable.

is it intentional we ignore disconnected as a state?

src/room/Room.ts
```diff
@@ -1551,12 +1551,12 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
   }
 
   private sendSyncState() {
```
davidzhao, line 1553: I'm wondering if this could move into RTCEngine or TransportManager. I understand it'll need a list of Tracks, but perhaps that could be a function that Room provides to RTCEngine?

davidzhao (approved):
Much cleaner! Nice work

src/room/PCTransportManager.ts
lukasIO: Good question, previously we only reacted to `failed`. Might make sense though to include it

src/room/Room.ts
lukasIO: addressed this in [f57d7c1](https://github.com/livekit/client-sdk-js/pull/909/commits/f57d7c1bf68496d60afbf8184a49fc8f8258162b)

src/room/PCTransportManager.ts
davidzhao: I think ignoring it as you do is probably the right behavior, it'll always turn from disconnected to failed. so it's better to not introduce a new behavior as part of an already-large change.

lukasIO:
> Were those places not handling things properly or the new code needs that behaviour?

Hardly any new behaviour needed. The changes are just surfacing the fact that some of the underlying methods had been async before already. I don't think this was actually causing any issues, but even just for error handling, it felt better to await things were possible

PR #911, opened 2023-10-26
Improve handling of incompatible published codecs

davidzhao:
When the client publishes a codec that the server doesn't support (or if server determines the client isn't compatible with it), we would reject those track publications.

However, the better thing to do from a user experience standpoint is to publish with another compatible codec. This compatibility layer works with server v1.5.1+

This change will use the preferred codec as returned in TrackPublishedResponse. Client will update its state accordingly.

Tested: disabling VP8 and publishing VP8, H.264, and AV1, verified backup is chosen and publishing at expected bitrates.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -749,7 +749,7 @@ export default class LocalParticipant extends Participant {
       }
     });
     if (primaryCodecMime && track.kind === Track.Kind.Video) {
-      const updatedCodec = primaryCodecMime.replace(/video\//y, '');
+      const updatedCodec = primaryCodecMime.replace(/video\//y, '').toLowerCase();
       if (updatedCodec !== videoCodec) {
         log.debug('falling back to server selected codec', { codec: updatedCodec });
         /* @ts-ignore */
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -18,6 +18,7 @@ import {
   TrackUnpublishedResponse,
 } from '../../proto/livekit_rtc_pb';
 import type RTCEngine from '../RTCEngine';
+import { defaultVideoCodec } from '../defaults';
 import { DeviceUnsupportedError, TrackInvalidError, UnexpectedConnectionState } from '../errors';
 import { EngineEvent, ParticipantEvent, TrackEvent } from '../events';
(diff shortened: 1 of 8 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -746,30 +740,19 @@ export default class LocalParticipant extends Participant {
...
-          `primary codec ${opts.videoCodec} not supported, fallback to ${backupCodec.codec}`,
-        );
-        opts.videoCodec = backupCodec.codec;
-        opts.videoEncoding = backupCodec.encoding;
-        encodings = simEncodings;
+    if (primaryCodecMime) {
+      const updatedCodec = primaryCodecMime.replace(/audio\/|video\//y, '');
+      if (updatedCodec !== videoCodec) {
+        /* @ts-ignore */
+        opts.videoCodec = updatedCodec;
```
cnderrauber, line 755: Do we need to compute encodings again when codec is updated? The bitrates for each resolution is determined by codec, especially for svc codecs.

davidzhao: got it. yeah let me add recompute here

cnderrauber (approved):
looks good!

src/room/participant/LocalParticipant.ts
```diff
@@ -746,30 +740,28 @@ export default class LocalParticipant extends Participant {
...
-        throw Error('cannot publish track, codec not supported by server');
-      }
-
-      if (!primaryCodecSupported && opts.backupCodec) {
-        const backupCodec = opts.backupCodec;
-        opts = { ...opts };
-        log.debug(
-          `primary codec ${opts.videoCodec} not supported, fallback to ${backupCodec.codec}`,
+    if (primaryCodecMime && track.kind === Track.Kind.Video) {
+      const updatedCodec = primaryCodecMime.replace(/video\//y, '').toLowerCase();
```
lukasIO, line 752: Just as a suggestion, we have a `mimeTypeToVideoCodecString` util function as part of e2ee, maybe makes sense to re-use here?

lukasIO (approved):
looks good to me!

src/room/participant/LocalParticipant.ts
davidzhao: awesome, this function is perfect. I've moved it to track/utils since it's now used outside of e2ee

PR #918, opened 2023-11-01
Don't set the autoplay attribute on video elements in Safari

lukasIO:
When in low power mode (applies to both macOS and iOS), Safari will show a play/pause overlay when a video starts that has the `autoplay` attribute set. We work around this by _not_ setting the autoplay attribute on safari and instead call `setTimeout(() => el.play(),0)` further down


davidzhao (approved):
lg!

PR #922, opened 2023-11-03
Correctly apply elementVolume on attach for webaudioMix

lukasIO:
Previously when attaching an element we did not honor the webaudioMix setting and would only set the volume on the element. 
This PR fixes this by calling `setVolume` directly to make sure we set the volume correctly for both modes html audio playback and webaudio mix


davidzhao (approved):
good catch!

PR #923, opened 2023-11-03
Simplify Multi-Codec Simulcast usage by setting backupCodec to true

davidzhao:
Currently it requires too many parameters to turn on the feature. The defaults were also dependent on input resolution. Making this optional and picking sane defaults.

### DIFF
--- a/example/sample.ts
+++ b/example/sample.ts
@@ -78,7 +78,7 @@ const appActions = {
     const autoSubscribe = (<HTMLInputElement>$('auto-subscribe')).checked;
     const e2eeEnabled = (<HTMLInputElement>$('e2ee')).checked;
 
-    setLogLevel(LogLevel.info);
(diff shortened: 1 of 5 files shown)


example/sample.ts
```diff
@@ -78,7 +78,7 @@ const appActions = {
     const autoSubscribe = (<HTMLInputElement>$('auto-subscribe')).checked;
     const e2eeEnabled = (<HTMLInputElement>$('e2ee')).checked;
 
-    setLogLevel(LogLevel.info);
+    setLogLevel(LogLevel.debug);
```
lukasIO, line 81: thanks 🙏

lukasIO (approved):
looks good to me!

example/sample.ts
davidzhao: hah, I'm finding myself having to change this each time I use it. it was getting annoying.

PR #926, opened 2023-11-07
Replace yarn with pnpm

lukasIO:
CI is failing due to a previously swallowed type error on main branch. 
This error only surfaced with the pnpm switch. It's fixed in this branch already, but the size-limit action also tries to build main with the same config as in this branch, which then fails.


davidzhao (approved):
good bye 🧶

Ocupe (approved):
🚀

PR #927, opened 2023-11-08
Add getStats() to PCTransport

rnakano:
We used to manually collect and analyze PeerConnection stats like this:

```
room.engine.publisher.pc.getStats();
```

However, this became unavailable with the changes in #903, So, we wanted to add a way to call getStats() as an alternative.

### DIFF
--- a/.changeset/strange-mails-smile.md
+++ b/.changeset/strange-mails-smile.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
Thanks for the PR. Understand the requirement, just be aware that we're planning some more changes to internal modules in the coming weeks, so there might be additional changes you'll have to make in the near term future.

rnakano:
Got it, and thanks for the review!

PR #933, opened 2023-11-13
Treat all signal messages as ping response

davidzhao:
In cases of rooms that are highly active, it's possible to have a lot of signal messages queued that it blocks the ping messages for awhile.

Since the purpose of ping/pong is to ensure signal channel is active, it makes sense to use any type of signal traffic as that indicator of liveliness.

### DIFF
--- a/.changeset/healthy-pillows-own.md
+++ b/.changeset/healthy-pillows-own.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Treat all signal messages as ping response
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -558,6 +558,8 @@ export class SignalClient {
       log.debug('received unsupported message');
       return;
     }
+
+    let pingHandled = false;
     if (msg.case === 'answer') {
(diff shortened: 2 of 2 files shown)


lukasIO:
I think what we're missing with this approach is detecting when only _sending_ of signal messages is broken.
IIRC the initial motivation to implement ping/pong as is, was due to the fact that we had seen some odd cases where it looked like server updates were still coming through but signal messages sent from the client to the server seemingly weren't.

davidzhao:
> I think what we're missing with this approach is detecting when only _sending_ of signal messages is broken. IIRC the initial motivation to implement ping/pong as is, was due to the fact that we had seen some odd cases where it looked like server updates were still coming through but signal messages sent from the client to the server seemingly weren't.

interesting.. so websocket was broken only in one direction? that's certainly a strange behavior.. but I guess we've seen stranger!

The motivation for this PR is due to a customer report that after paging very quickly, they were disconnected from the room due to idle timer. The reason for that failure is ping timer unable to deliver its messages to the server quickly enough.. likely due to the number of subscription changes to send before then.

lukasIO (approved):

PR #939, opened 2023-11-16
Add `startVideo` method and `RoomEvent.VideoPlaybackStatusChanged`

lukasIO:
This is useful to show a play button when auto playing of videos has been blocked by the browser. 
Currently this is the case on Safari in low power mode.


davidzhao (approved):
LG! gotta love these Safari quirks /s

PR #945, opened 2023-11-28
Rename getTrack and room.participants

lukasIO:
- https://github.com/livekit/client-sdk-js/pull/945/commits/f9aa02c681b00aa0c3380dafe3232bc66b7bd5b0 only deprecates APIs and introduces new ones
- https://github.com/livekit/client-sdk-js/pull/945/commits/fde6dd4195825aa50d7dc8ff92f6d9686f5619f8 actually removes the deprecated ones

splitting it up like that would in theory allow us to cherry pick the deprecation PR into main, release the deprecation changes as a minor version, and then actually remove the APIs in v2, which is the recommended semver way of doing things.


src/room/Room.ts
```diff
@@ -1497,14 +1503,14 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
-  private getOrCreateParticipant(id: string, info: ParticipantInfo): RemoteParticipant {
-    if (this.participants.has(id)) {
-      return this.participants.get(id) as RemoteParticipant;
+  private getOrCreateParticipant(identity: string, info: ParticipantInfo): RemoteParticipant {
+    if (this.remoteParticipants.has(identity)) {
+      return this.remoteParticipants.get(identity) as RemoteParticipant;
     }
-    const participant = this.createParticipant(id, info);
-    this.participants.set(id, participant);
+    const participant = this.createParticipant(identity, info);
```
davidliu, line 1510: Did `createParticipant` also get changed? As it is, `createParticipant` expects the participant sid in the first argument.

lukasIO: That's a good call. I don't think there are still code paths that would instantiate a participant without a accompanying participantInfo object, but added a commit to instantiate the participant with identity if that's the case.

PR #946, opened 2023-11-28
Change publishData signature

lukasIO:
I think for this one we'll have to live with a hard change as we simultaneously want to switch from using `sids` to `identities` which I don't think we could easily do in a backwards compatible way in a deprecation step. (We'd need to figure out if the user intends to set sids or identities)

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -1044,7 +1044,7 @@ export default class LocalParticipant extends Participant {
    */
   async publishData(data: Uint8Array, options: DataPublishOptions = {}): Promise<void> {
     const kind = options.reliable ? DataPacket_Kind.RELIABLE : DataPacket_Kind.LOSSY;
-    const destination = options.destination;
+    const destination = options.destinations;
     const topic = options.topic;
 
     const destinationIdentities: Array<string> = [];
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -44,7 +44,6 @@ import { Future, isFireFox, isSVCCodec, isSafari, isWeb, supportsAV1, supportsVP
 import Participant from './Participant';
 import type { ParticipantTrackPermission } from './ParticipantTrackPermission';
 import { trackPermissionToProto } from './ParticipantTrackPermission';
-import RemoteParticipant from './RemoteParticipant';
 import {
   computeTrackBackupEncodings,
   computeVideoEncodings,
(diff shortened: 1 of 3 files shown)


davidzhao (commented):
I think it's still kind of tricky for someone who's using the `options` interface to realize it's changed for them. Either of the following could overcome that:
* separate identity spec from `destination` field so that it's `destinationIdentities`
* create a type for Participant.Identity so that it would be impossible to mix up with a standard string.

davidzhao:
> > separate identity spec from destination field so that it's destinationIdentities
> 
> this is not ideal as the accepted type is `string | RemoteParticipant`

what I meant is separating them so it's not bundled together

```
type DataPublishOptions = {
  destinationParticipants?: []RemoteParticipant
  destinationIdentities?: []string
}
```
My main hope is that users will realize it no longer compiles and specifically set identities field. If it's simply renaming a field, they would not bother reading the details of the change.

lukasIO:
what happens when both are set? Does it result in the message being receivend by both added together or does one field take precedence over the other?

I think as a user I would feel confused whether I have to set both of them or not.

> My main hope is that users will realize it no longer compiles and specifically set identities field. If it's simply renaming a field, they would not bother reading the details of the change.

Yeah that makes sense. But it's a major change after all. Our migration guide should make this clear.

davidzhao:
> what happens when both are set? Does it result in the message being received by both added together or does one field take precedence over the other?

What do you think about throwing an exception? Just an idea tho. It's your decision on which path we ended up taking.

davidzhao (approved):

src/room/participant/LocalParticipant.ts
```diff
@@ -1040,64 +1039,21 @@ export default class LocalParticipant extends Participant {
...
-      destination.forEach((val: any) => {
-        if (val instanceof RemoteParticipant) {
-          destinationSids.push(val.sid);
-        } else {
-          destinationSids.push(val);
-        }
-      });
-    }
+  async publishData(data: Uint8Array, options: DataPublishOptions = {}): Promise<void> {
+    const kind = options.reliable ? DataPacket_Kind.RELIABLE : DataPacket_Kind.LOSSY;
```
hiroshihorie, line 1045: Not sure if I'm reading the code correctly but `DataPublishOptions.reliable` is null by default so this will default to `LOSSY` ?

lukasIO: yeah, `LOSSY` by default is what is intended!

davidzhao (approved):
lgtm!

PR #949, opened 2023-11-29
Use single state var to track connection state of signal client

lukasIO:
previously we had `isConnecting` and `isReconnecting` as two boolean values, with the possibility to create ambigious states. 
This PR instead uses a singe `SignalConnectionState` to track the current signal client connection status. 

(hide whitespace recommended for review)

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -1,3 +1,4 @@
+import { SignalConnectionState } from '../../api/SignalClient';
 import log from '../../logger';
 import type { InternalRoomOptions } from '../../options';
 import {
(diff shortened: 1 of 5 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -1129,7 +1130,7 @@ export default class LocalParticipant extends Participant {
   ) {
     this.participantTrackPermissions = participantTrackPermissions;
     this.allParticipantsAllowedToSubscribe = allParticipantsAllowed;
-    if (this.engine.client.isConnected) {
+    if (this.engine.client.currentState < SignalConnectionState.DISCONNECTING) {
```
davidzhao, line 1133: it'd be better to abstract the enum math under something like a `get isDisconnected():boolean`

davidzhao (approved):
nice! much cleaner

PR #965, opened 2023-12-15
Fix dynacast can't work after full reconnect

cnderrauber:


src/api/SignalClient.ts
```diff
@@ -368,11 +368,8 @@ export class SignalClient {
     this.onAnswer = undefined;
     this.onLeave = undefined;
     this.onLocalTrackPublished = undefined;
-    this.onLocalTrackUnpublished = undefined;
```
davidzhao, line 371: seems like we'd want to keep these for when it's really cleaning up. So I guess that's the challenge here.

I think resetting them might be ok, but we just need to hook them up to the new callbacks during reconnect? wdyt @lukasIO ?

lukasIO: agree that we'd want to unset them at one point. Will take a look at how to hook them up during reconnect already

PR #966, opened 2023-12-15
Make sure all client callbacks are set up after reconnect

lukasIO:
closes #965, attempts to fix the same issue without compromising on the cleanup of the callbacks.


cnderrauber (approved):
Looks good!

PR #972, opened 2023-12-22
Improve default screen sharing FPS, limiting capturing surface

davidzhao:
Previously when we didn't constrain screen share resolution, screen share tracks weren't resized to a reasonable size prior to publishing. Having to encode a huge frame caused FPS to be very low (often just 1-5 fps).

Because that's rather confusing behavior, we are now defaulting to 1080p when capturing


lukasIO:
When removing the default resolution other than fixing the Safari bug we also did so in order to enable users to capture at native screen sizes. When applying a default like this, that won't be possible anymore.

I think both makes sense: having a reasonable default, but also allowing users to capture uncapped resolution if they wish to do so. Maybe a `ScreenSharePreset.Uncapped` or something similar would allow for that? Not sure

davidzhao:
@lukasIO I've added `original`. lmk what you think.

lukasIO:
nice! was thinking whether explicitly passing `null` would be an option, too, but I somehow prefer the original preset!

lukasIO (approved):

package.json
```diff
@@ -40,7 +40,7 @@
     "build:watch": "rollup --watch --config rollup.config.js",
     "build-docs": "typedoc",
     "proto": "protoc --es_out src/proto --es_opt target=ts -I./protocol ./protocol/livekit_rtc.proto ./protocol/livekit_models.proto",
-    "sample": "vite example -c vite.config.js",
+    "sample": "vite example -c vite.config.mjs",
```
lukasIO, line 43: thanks 🙏

lukasIO (approved):

Issue #971, opened 2023-12-21
`mediaStreamTrack` doesn't stop after executing `setMediaStreamTrack` multiple times


mpnri:
### Describe the bug

The `restartTrack` function creates a new `mediaStreamTrack` and stops the previous one. Then, it assigns the new `mediaStreamTrack` to the `_mediaStreamTrack` attribute of the `LocalTrack` class (in the `setMediaStreamTrack` function).

However, if the `restartTrack` function is called multiple times in a short time interval, it can cause a bug. For example, imagine that `restartTrack` is executed twice within a short time interval. Then the code `this._mediaStreamTrack.stop()` in the (2) second execution may run before the code `this._mediaStreamTrack = newTrack` in the (1) first execution. As a result, the `mediaStreamTrack` from the (1) first execution will never be stopped and there will be no way to access it again.
![image](https://github.com/livekit/client-sdk-js/assets/47795908/d5e3ad20-d09b-4d79-969a-4535d7cf311a)

This problem causes the browser to say `This tab is using the camera or microphone` **even after the end of the meet**, which causes the user to feel **insecure**.
![image](https://github.com/livekit/client-sdk-js/assets/47795908/88c9f54c-f0d6-478f-a85c-9e6bdc300eb7)

### Suggested solutions
+ #### 1- Stop older `mediaStreamTrack` just before placing it in `this._mediaStreamTrack`(easier solution):
This change ensures that the older `mediaStreamTrack` is being stopped each time before the new `mediaStreamTrack` is being saved.(#980)

+ #### 2- Add a Mutex lock for `LocalTrack.restart()` function(safer solution):
This change ensures that `restart` method calls are not executed in parallel:

![image](https://github.com/livekit/client-sdk-js/assets/47795908/711b491f-deb8-4fcd-b449-6c8db4e9d9a8)

### Reproduction

A simple way for reproducing this problem is clicking multiple times on `Flip(back) Camera` button  in a short time interval in the [example project](https://github.com/livekit/client-sdk-js/tree/main/example).

### System Info

```shell
-
```

### Severity

serious, but I can work around it

### Additional Information

Besides the case mentioned above, there are some other scenarios that may happen very rarely, but they are caused by the uncertainty in the sequential execution of some async functions.
For instance, the `localParticipant.setCameraEnabled(false)` function may be executed while the `restartTrack` function is still running, which results in the new `mediaStreamTrack` being unmuted despite `localParticipant.isCameraEnabled` being false.

[...]

PR #980, opened 2024-01-01
Fix stopping old track in `setMediaStreamTrack`

mpnri:
Fixes [#971](https://github.com/livekit/client-sdk-js/issues/971)'s main problem using first solution


lukasIO (approved):
thanks, this makes sense!

PR #985, opened 2024-01-03
Remove `OFF` option from VideoQuality enum

lukasIO:


src/room/track/Track.ts
```diff
@@ -13,6 +17,27 @@ const BACKGROUND_REACTION_DELAY = 5000;
...
+}
+
+/**
+ * @internal
+ */
+export function videoQualityFromProto(q: ProtoQuality): VideoQuality {
+  switch (q) {
+    case ProtoQuality.HIGH:
+    case ProtoQuality.OFF:
+      return VideoQuality.HIGH;
```
boks1971, line 33: Would this cause trouble? Should the last video quality be cached and used when OFF? This is tricky.

boks1971 (approved):
LGTM! One comment about the mapping.

src/room/track/Track.ts
lukasIO: that's a good point. Currently this mapping function isn't even used anywhere in the code base. I added it because we have a similar helper to convert `ConnectionQuality` to a string based enum. 

We can also simply remove this for now.

davidzhao (approved):
lgtm!

PR #988, opened 2024-01-04
Add class level configurable logger

lukasIO:
- By default each (base)class gets its own logger instance that starts with `livekit-` (e.g. `livekit-participant`, `livekit-track`
- This can be overriden by providing a `loggerName` in the roomOpts. If set, this loggerName will be propagated across all objects/classes instantiated within the room and only a single logger instance will be used per room
- this allows for two usage scenarios
   - class based logging and the possibility to set log levels independently for each (base)class
   - room based logging and the possibility to set different log levels and/or extensions for each room

### DIFF
--- a/src/room/PCTransport.ts
+++ b/src/room/PCTransport.ts
@@ -2,7 +2,7 @@ import { EventEmitter } from 'events';
 import type { MediaDescription } from 'sdp-transform';
 import { parse, write } from 'sdp-transform';
 import { debounce } from 'ts-debounce';
-import log from '../logger';
+import log, { getLogger } from '../logger';
 import { NegotiationError, UnexpectedConnectionState } from './errors';
 import { ddExtensionURI, isChromiumBased, isSVCCodec } from './utils';
 
@@ -43,6 +43,8 @@ export default class PCTransport extends EventEmitter {
 
   private mediaConstraints: Record<string, unknown>;
 
+  private log = log;
+
   pendingCandidates: RTCIceCandidateInit[] = [];
 
   restartingIce: boolean = false;
@@ -71,8 +73,13 @@ export default class PCTransport extends EventEmitter {
 
   onTrack?: (ev: RTCTrackEvent) => void;
 
-  constructor(config?: RTCConfiguration, mediaConstraints: Record<string, unknown> = {}) {
+  constructor(
+    config?: RTCConfiguration,
+    mediaConstraints: Record<string, unknown> = {},
+    logger = 'livekit-pc-transport',
+  ) {
     super();
+    this.log = getLogger(logger);
     this.config = config;
     this.mediaConstraints = mediaConstraints;
     this._pc = this.createPC();
@@ -229,7 +236,7 @@ export default class PCTransport extends EventEmitter {
     }
 
     if (options?.iceRestart) {
-      log.debug('restarting ICE');
+      this.log.debug('restarting ICE');
       this.restartingIce = true;
     }
 
@@ -246,12 +253,12 @@ export default class PCTransport extends EventEmitter {
         return;
       }
     } else if (!this._pc || this._pc.signalingState === 'closed') {
-      log.warn('could not createOffer with closed peer connection');
+      this.log.warn('could not createOffer with closed peer connection');
       return;
     }
 
     // actually negotiate
-    log.debug('starting to negotiate');
+    this.log.debug('starting to negotiate');
     const offer = await this.pc.createOffer(options);
 
     const sdpParsed = parse(offer.sdp ?? '');
@@ -452,7 +459,7 @@ export default class PCTransport extends EventEmitter {
       const originalSdp = sd.sdp;
       sd.sdp = munged;
       try {
-        log.debug(`setting munged ${remote ? 'remote' : 'local'} description`);
+        this.log.debug(`setting munged ${remote ? 'remote' : 'local'} description`);
         if (remote) {
           await this.pc.setRemoteDescription(sd);
         } else {
@@ -460,7 +467,7 @@ export default class PCTransport extends EventEmitter {
         }
         return;
       } catch (e) {
-        log.warn(`not able to set ${sd.type}, falling back to unmodified sdp`, {
+        this.log.warn(`not able to set ${sd.type}, falling back to unmodified sdp`, {
           error: e,
           sdp: munged,
         });
@@ -491,7 +498,7 @@ export default class PCTransport extends EventEmitter {
       if (!remote && this.pc.remoteDescription) {
         fields.remoteSdp = this.pc.remoteDescription;
       }
-      log.error(`unable to set ${sd.type}`, fields);
+      this.log.error(`unable to set ${sd.type}`, fields);
       throw new NegotiationError(msg);
     }
   }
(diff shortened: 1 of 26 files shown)


src/room/participant/RemoteParticipant.ts
```diff
@@ -159,7 +160,7 @@ export default class RemoteParticipant extends Participant {
     // yet arrived. Wait a bit longer for it to arrive, or fire an error
     if (!publication) {
       if (triesLeft === 0) {
-        log.error('could not find published track', { participant: this.sid, trackSid: sid });
+        this.log.error('could not find published track', { participant: this.sid, trackSid: sid });
```
boks1971, line 163: Have not looked at logger implementation. But, wondering if things like  `sid` can be set in the logger context. And log it with all the log messages. Basically, set it once and then you have pertinent information with all log messages.

Then make logger hierarchical. Basically, this participant logger will have the participant identity, sid set. When creating a track of this participant, inherit this participant's logger and add things like track sid, kind, mime, etc to the track logger. So, when the track logger logs, it has the participant information + track information all baked in with all log messages.

We have such loggers on server side. Basically, it is all key/value pairs set in the logger context and whatever key/value pairs are in the context are logged with every message. key will be string. value could be any type.

src/room/track/LocalTrack.ts
```diff
@@ -50,8 +49,9 @@ export default abstract class LocalTrack extends Track {
     kind: Track.Kind,
     constraints?: MediaTrackConstraints,
     userProvidedTrack = false,
+    logger?: string,
```
boks1971, line 52: `loggerName`

src/room/track/LocalTrackPublication.ts
```diff
@@ -16,8 +16,8 @@ export default class LocalTrackPublication extends TrackPublication {
     return this.track?.isUpstreamPaused;
   }
 
-  constructor(kind: Track.Kind, ti: TrackInfo, track?: LocalTrack) {
-    super(kind, ti.sid, ti.name);
+  constructor(kind: Track.Kind, ti: TrackInfo, track?: LocalTrack, logger?: string) {
```
boks1971, line 19: `loggerName`

src/room/track/RemoteTrack.ts
```diff
@@ -11,8 +11,10 @@ export default abstract class RemoteTrack extends Track {
     sid: string,
     kind: Track.Kind,
     receiver?: RTCRtpReceiver,
+    logger?: string,
```
boks1971, line 14: `loggerName`

src/room/track/Track.ts
```diff
@@ -50,8 +50,15 @@ export abstract class Track extends (EventEmitter as new () => TypedEventEmitter
 
   protected monitorInterval?: ReturnType<typeof setInterval>;
 
-  protected constructor(mediaTrack: MediaStreamTrack, kind: Track.Kind) {
+  protected log: StructuredLogger = log;
+
+  protected constructor(
+    mediaTrack: MediaStreamTrack,
+    kind: Track.Kind,
+    logger: string = LoggerNames.Track,
```
boks1971, line 58: `loggerName`

src/room/track/TrackPublication.ts
```diff
@@ -39,8 +39,16 @@ export class TrackPublication extends (EventEmitter as new () => TypedEventEmitt
...
   protected encryption: Encryption_Type = Encryption_Type.NONE;
 
-  constructor(kind: Track.Kind, id: string, name: string) {
+  protected log = log;
+
+  constructor(
+    kind: Track.Kind,
+    id: string,
+    name: string,
+    logger: string = LoggerNames.Publication,
```
boks1971, line 48: `loggerName`

boks1971 (approved):
LGTM! Some minor comments. Mainly interested if setting key/value pairs in context is possible.

davidzhao (commented):
lg. agreed with the naming nit Raja already mentioned.

src/room/participant/RemoteParticipant.ts
lukasIO: Agreed, that would be an ideal solution. 

With that setup we'd need a logger instance for each participant+publication+track whereas right now there's only one for each class. Don't think the impact should be big, but when thinking about rooms with 1000s of participants we should verify that's not a bottleneck at some point. 

Generally the logger context we currently provide is a per-message context. We'd have to wrap the logging library we currently use to include that information. As far as I'm aware a big downside of doing that would be that we lose the browser built-in tracing information of where the log originated from. 
I'll do some more tests to see if we could potentially work around this though.

And a thought: Hierarchical logging might be trickier for `Tracks` as those can be created outside of a room instance.

src/room/PCTransport.ts
```diff
@@ -71,8 +73,13 @@ export default class PCTransport extends EventEmitter {
 
   onTrack?: (ev: RTCTrackEvent) => void;
 
-  constructor(config?: RTCConfiguration, mediaConstraints: Record<string, unknown> = {}) {
+  constructor(
+    config?: RTCConfiguration,
+    mediaConstraints: Record<string, unknown> = {},
+    logger: string = LoggerNames.PCTransport,
```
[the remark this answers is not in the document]
lukasIO: yeah, not sure why I switched to `logger` for half of those 🤦  
`loggerName` makes a lot more sense.

src/room/participant/RemoteParticipant.ts
boks1971: Got it. Yeah thousands of tracks is definitely a concern. I am guessing doing some sort of helper to augment logging context would work as well. Basically something like 'this.log(message, this.logFields()...)' and each instance can populate it.

And for tracks without room, things can be set in logging context as and when they become available.

May not be worth all this though.

lukasIO: yeah, the helper method was on my mind, too! 
Agree that that would definitely be an improvement

lukasIO: Added log context to all log calls. log context gets inherited where possible so that we get as much info as possible

src/room/track/utils.ts
```diff
@@ -209,3 +209,23 @@ export function getTrackPublicationInfo<T extends TrackPublication>(
...
+      kind: track.kind,
+    };
+  } else {
+    return {
+      sid: track.trackSid,
+      name: track.trackName,
+      track: track.track ? getLogContextFromTrack(track.track) : undefined,
+      enabled: track.isEnabled,
+      encrypted: track.isEncrypted,
+    };
```
boks1971, line 229: maybe add codec mime type if available, will be useful with the various combinations we have :-)

boks1971 (approved):
LGTM! Great work @lukasIO !!!

src/room/Room.ts
```diff
@@ -1870,12 +1917,29 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
   ): boolean {
     // active speaker updates are too spammy
     if (event !== RoomEvent.ActiveSpeakersChanged) {
-      log.debug(`room event ${event}`, { event, args });
+      // only extract logContext from arguments in order to avoid logging the whole object tree
+      const minimizedArgs = mapArgs(args).filter((arg: unknown) => arg !== undefined);
+      this.log.debug(`room event ${event}`, { ...this.logContext, event, args: minimizedArgs });
```
lukasIO, line 1922: this is new and might be of interest: 
previously the `args` of each roomevent got posted in their entirety to the console, including whole participant + track classes. 
This change extracts only the `logContext` from the args if available or in case they are primitive values leaves them as is.

boks1971: This is great!!!

PR #990, opened 2024-01-10
Await unpublish before re-publishing on signal-reconnect

lukasIO:


boks1971 (approved):

src/room/Room.ts
```diff
@@ -1131,40 +1131,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
-      this.localParticipant.trackPublications.forEach((pub) => {
-        if (pub.track) {
-          localPubs.push(pub);
-        }
-      });
-
-      await Promise.all(
-        localPubs.map(async (pub) => {
-          const track = pub.track!;
-          this.localParticipant.unpublishTrack(track, false);
```
lukasIO, line 1144: this was missing an `await`. FWIW the `republishAllTracks` method had the await already, so opted to just use that method instead.

davidzhao: I think previously we didn't automatically republish screenshare/audio tracks.. I'm not too sure why.. I think it makes sense not to make that differentiation though.

davidzhao (approved):
nice fix!

src/room/Room.ts
lukasIO: AFAIK we tried to re-publish them but we didn't try to `restart` them as that's not really possible in the browser

lukasIO: we should probably add that check to the participant method, too. good call :)

PR #994, opened 2024-01-12
Fix unexpected closing of signal connection during reconnect

lukasIO:


src/api/SignalClient.ts
```diff
@@ -350,6 +360,8 @@ export class SignalClient {
                 resolve();
                 shouldProcessMessage = true;
               }
+            } else if (this.isEstablishingConnection && resp.message.case === 'leave') {
+              reject(new ConnectionError('Received leave request while trying to (re)connect'));
```
lukasIO, line 364: not sure if this is the intended behaviour: 

This means if a user is reconnecting and the very first message they receive is not a `ReconnectResponse` but rather a leave request, we mark the reconnection attempt as failed, which will result in a new reconnection attempt happening.

boks1971: The intended behaviour there is for the client to do a full reconnect. Server sends `LeaveRequest` when it cannot resume. So, a full reconnect is needed. Is the new reconnection attempt a full reconnect?

boks1971 (approved):
LGTM, but please get @davidzhao 's 👀 as I am not aware of all the subtleties here.

davidzhao (approved):
lgtm!

PR #1005, opened 2024-01-24
Enable layers after resume

lukasIO:


boks1971 (approved):

davidzhao (approved):
oh wow! great catch.

should we send an explicit signal to resume all layers on the server side too?

lukasIO:
> should we send an explicit signal to resume all layers on the server side too?

great point. closing this, as it makes more sense to trigger this from the server side

PR #1006, opened 2024-01-25
Send appropriate event if participant metadata changed during disconnect/resume

jure:
We've encountered an issue where:
- if we're going through a signal disconnect/resume and
- a remote participant's metadata changes in the mean time
- when we come back, there are no `ParticipantMetadataChanged` events, yet the metadata is potentially different

I'm not sure this is the right way to do it (though it does match in approach and semantics to what you're doing with caching participant IDs during reconnects), but I thought I would open up a discussion with a solution in tow - if nothing else I hope this PR clarifies what exactly the issue is.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -142,6 +142,8 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
 
   private cachedParticipantIdentities: Array<string>;
 
+  private cachedParticipantMetadata: Map<string, string | undefined>;
+
   private connectionReconcileInterval?: ReturnType<typeof setInterval>;
 
   private regionUrlProvider?: RegionUrlProvider;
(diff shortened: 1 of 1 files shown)


lukasIO:
Thanks for the PR! 
This makes sense, but makes me think that we actually would want to potentially synthesise all relevant ParticipantEvents, not only the metadatachanged one. cc @davidzhao

davidzhao:
> makes me think that we actually would want to potentially synthesise all relevant ParticipantEvents

That makes sense. I think this stems from our use of `emitWhenConnected`. The events that would have been sent would have been swallowed during that time.. It seems that during resume, the right behavior would be to buffer them instead of dropping them entirely.

lukasIO:
buffering them would only be relevant to events that actually arrive though, right? 
In cases where signal updates would be sent during a signal-reconnect, we might still miss them

davidzhao:
> buffering them would only be relevant to events that actually arrive though, right? In cases where signal updates would be sent during a signal-reconnect, we might still miss them

after signal connection is established, it would receive participant updates for everyone in the room. So the delta logic that fires events should work ok

lukasIO:
> after signal connection is established, it would receive participant updates for everyone in the room. 

that makes sense, I started an effort to buffer events in the timespan between signal reconnected and connection fully re-established. 

@jure could you give this PR a try?
https://github.com/livekit/client-sdk-js/pull/1007

PR #1009, opened 2024-01-26
Update participant info when getting participant

lukasIO:


src/room/Room.ts
```diff
@@ -1558,7 +1558,11 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
 
   private getOrCreateParticipant(identity: string, info: ParticipantInfo): RemoteParticipant {
     if (this.remoteParticipants.has(identity)) {
-      return this.remoteParticipants.get(identity) as RemoteParticipant;
+      const existingParticipant = this.remoteParticipants.get(identity)!;
+      if (info) {
```
davidzhao, line 1562: I think we already update this [here](https://github.com/livekit/client-sdk-js/blob/main/src/room/Room.ts#L1293). Should we remove that update then?

we'll probably want to move over the setter for sidToIdentity logic as well.

davidzhao (approved):

src/room/Room.ts
lukasIO: oh, you're right. Thanks for pointing that out.
Given my confusion as an example, it might make sense to unify the update path. Will revert as a draft for now, this doesn't improve anything as is.

src/room/Room.ts
```diff
@@ -1283,21 +1283,17 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
 
       let remoteParticipant = this.remoteParticipants.get(info.identity);
-      const isNewParticipant = !remoteParticipant;
 
       // when it's disconnected, send updates
-      if (info.state === ParticipantInfo_State.DISCONNECTED) {
+      if (
+        info.state === ParticipantInfo_State.DISCONNECTED &&
+        // make sure we emit the disconnected event only if the info is newer
+        (!remoteParticipant?.infoVersion || info.version > remoteParticipant.infoVersion)
```
lukasIO, line 1291: not sure if this check is strictly necessary. The intention is to make sure that we only disconnect a participant if we don't have more up-to-date info about them already

davidzhao: I don't think we need this check. the server is already serializing the events so it would not send a disconnected out of order

davidzhao (approved):

PR #1012, opened 2024-01-30
Don't emit RoomEvent.Reconnecting for resumes

lukasIO:


src/room/Room.ts
```diff
@@ -587,7 +587,11 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     opts: RoomConnectOptions | undefined,
     abortController: AbortController,
   ) => {
-    if (this.state === ConnectionState.Reconnecting) {
+    if (
+      this.state === ConnectionState.Reconnecting ||
+      this.engine.client.currentState === SignalConnectionState.RECONNECTING ||
+      this.engine.pcManager?.currentState === PCTransportState.CONNECTING
```
lukasIO, line 593: this isn't nice. thinking about it, it might make more sense to keep an `isResuming` flag internally for these kind of checks.

src/room/Room.ts
```diff
@@ -689,7 +693,8 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       });
       if (
         this.state === ConnectionState.Connecting ||
-        this.state === ConnectionState.Reconnecting
+        this.state === ConnectionState.Reconnecting ||
+        this.connectFuture
```
lukasIO, line 697: same as above, would be more consistent with an `isResuming` flag

davidzhao (approved):
lgtm

PR #1018, opened 2024-01-31
Point to migration guide in README

lukasIO:


davidzhao (approved):
lg! we can merge after promoting docs this week.

PR #1022, opened 2024-02-05
Remove googConstraints from RTCPeerConnection constructor

lukasIO:
These constraints have been deprecated in Chromium since M103 and should be simply ignored by Chrome since then. 
Users reported errors however (coincidentally with Chrome 103) with the error message `Failed to construct 'RTCPeerConnection': Malformed constraints object.` Was not able to repro this error using Chrome 103.

Given that this constraint was only ever used for enabling setting local network priority and has been ignored in Chrome for a long time, it seems like it should be ok to remove it entirely and therefore reducing error potential. Removing it will mean manually setting `networkPriority` will not have any effect on Chrome prior to version 103.

Alternative would be to try/catch RTCPeerConnection instantiation.

### DIFF
--- a/.changeset/great-impalas-try.md
+++ b/.changeset/great-impalas-try.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Remove googConstraints from RTCPeerConnection constructor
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
lgtm. agreed this is safer from a compatibility standpoint

lukasIO:
that's a good catch! that might have been the issue. Thanks for pointing that out. 
Don't see a need to revert the removal of the googConstraints though. Do you have a usecase that depends on them?

PR #1023, opened 2024-02-05
feat: add resuming state to room events

holzgeist:
I want to give users feedback about network events ASAP.
Currently, if network access is temporarily lost, the first event that gets emitted is a Reconnected event.
Having access to the internal Resuming event, we can show the users that something is up. This should help to avoid repeated "Hello? Can you hear me?" that nobody will ever hear ;)

WDYT?

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -347,6 +347,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
         this.registerConnectionReconcile();
         this.isResuming = false;
         this.log.info('Resumed signal connection', this.logContext);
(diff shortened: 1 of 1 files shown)


holzgeist:
I didn't add a changeset because I'm not sure about the severity of this change. Is adding an enum variant a major/breaking change?

lukasIO:
hi @holzgeist,

thanks for the PR! We're still discussing internally when and how to exactly emit such an event. We'll come back to this once a decision has been made.

holzgeist:
closing, because same functionality has been implemented in  #1158

lukasIO:
thank you for your efforts here, @holzgeist !

PR #1033, opened 2024-02-16
Stronger kind type for Tracks to improve processor support

lukasIO:
this makes `kind` of `Track`s readonly and makes it easier to deal with different options needed for different processor types (e.g. video vs. audio). 

This should not break any existing types.


davidzhao (approved):
lgtm

PR #1036, opened 2024-02-19
Emit event when track processor changes

lukasIO:
some more glue code to better support advanced track processors.


davidzhao (approved):
lg

PR #1040, opened 2024-02-21
Add `stopProcessor`  param to replaceTrack function

lukasIO:


src/room/track/LocalTrack.ts
```diff
@@ -221,7 +221,7 @@ export default abstract class LocalTrack<
     return this;
   }
 
-  async replaceTrack(track: MediaStreamTrack, userProvidedTrack = true) {
+  async replaceTrack(track: MediaStreamTrack, userProvidedTrack = true, stopProcessor = true) {
```
davidzhao, line 224: not sure if we want to do this now.. but might be cleaner to support a ReplaceOptions object instead of multiple args? I really wish JS supported keyword args.

davidzhao (approved):

src/room/track/LocalTrack.ts
lukasIO: you're right, that would be cleaner!

davidzhao (approved):
awesome!

PR #1044, opened 2024-02-22
Don't compute video aspectRatio by default on presets

lukasIO:
- adds a new overload for VideoPreset that takes a videoPresetOptions object with optional `aspectRatio` parameter.
- does not change the old behaviour of auto computing an aspect ratio if the old constructor is used


davidzhao (approved):
looks good!

PR #1048, opened 2024-02-27
Only perform mute/unmute actions if necessary

lukasIO:
restarting tracks seemingly takes a lot longer on new Safari versions (>=17.3.1). 
This tries to minimise that impact by avoiding unnecessary restarts to tracks. 
Previously calling `unmute` repeatedly would have led to multiple track restarts even if the track is unmuted already.

### DIFF
--- a/src/room/track/LocalAudioTrack.ts
+++ b/src/room/track/LocalAudioTrack.ts
@@ -50,7 +50,6 @@ export default class LocalAudioTrack extends LocalTrack<Track.Kind.Audio> {
 
   async mute(): Promise<typeof this> {
     const unlock = await this.muteLock.lock();
-
     try {
       if (this.isMuted) {
         this.log.debug('Track already muted', this.logContext);
@@ -72,7 +71,6 @@ export default class LocalAudioTrack extends LocalTrack<Track.Kind.Audio> {
 
   async unmute(): Promise<typeof this> {
     const unlock = await this.muteLock.lock();
-
     try {
       if (!this.isMuted) {
         this.log.debug('Track already unmuted', this.logContext);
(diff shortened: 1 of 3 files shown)


src/room/track/LocalAudioTrack.ts
```diff
@@ -50,6 +50,10 @@ export default class LocalAudioTrack extends LocalTrack<Track.Kind.Audio> {
 
   async mute(): Promise<typeof this> {
     const unlock = await this.muteLock.lock();
+    if (this.isMuted) {
+      this.log.debug('Track already muted', this.logContext);
+      return this;
```
boks1971, line 55: Unlock here and in unmute?

boks1971 (commented):
Sorry, just drive by review

src/room/track/LocalAudioTrack.ts
lukasIO: uff. thanks for catching this in a drive-by! 
Had it within the `try` clause earlier which would have made the unlock unnecessary. Will move it back in there

src/room/track/LocalVideoTrack.ts
```diff
@@ -133,6 +138,11 @@ export default class LocalVideoTrack extends LocalTrack<Track.Kind.Video> {
   async unmute(): Promise<typeof this> {
     const unlock = await this.muteLock.lock();
     try {
+      if (!this.isMuted) {
```
lukasIO, line 141: we could add a check here for `mediaStreamTrack.readyState === 'live'` just to make sure?

boks1971: I think it is fine to leave it out. If it not live, it is ended and the following action is probably not user perceptible any way. Want to avoid checks which probably increases the surface area of browser weirdness affecting functionality in unexpected ways.

PR #1052, opened 2024-02-29
Don't treat PC config error as SignalReconnectError

### LINKED ISSUE #1050: Failed to execute 'setRemoteDescription' on 'RTCPeerConnection': The RTCPeerConnection's signalingState is 'closed'.
### Describe the bug

I have this issue that I cannot reproduce at all, but get a lot of errors in sentry for. Could you please take a look and point in a direction to how to fix it. Here's the log I can extract from sentry. 

### Reproduction

N/A

### Logs

```shell
[
  {
    "level": "warning",
    "data": {
      "arguments": [
        "websocket closed",
        {
          "code": 1006,
          "identity": "<user_id>",
          "reason": "",
          "room": "room_b",
          "roomSid": "room_sid_b",
          "state": 1,
          "wasClean": false
        }
      ]
    }
  },
  {
    "level": "warning",
    "data": {
      "arguments": [
        "signal disconnected",
        {
          "identity": "<user_id>",
          "room": "room_b",
          "roomSid": "room_sid_b"
        }
      ]
    }
  },
  {
    "level": "info",
    "data": {
      "arguments": [
        "resuming signal connection, attempt 0",
        {
          "identity": "<user_id>",
          "room": "room_b",
          "roomSid": "room_sid_b"
        }
      ]
    }
  },
  {
    "level": "info",
    "data": {
      "arguments": [
        "Resuming signal connection",
        {
          "identity": "<user_id>",
          "room": "room_b",
          "roomSid": "room_sid_b"
        }
      ]
    }
  },
  {
    "level": "error",
    "data": {
      "arguments": [
        "Failed to execute 'setConfiguration' on 'RTCPeerConnection': The RTCPeerConnection's signalingState is 'closed'.",
        {
          "error": "[DOMException]",
          "identity": "<user_id>",
          "room": "room_b",
          "roomSid": "room_sid_b"
        }
      ]
    }
  },
  {
    "level": "warning",
    "data": {
      "arguments": [
        "reconnect disconnected",
        {
          "identity": "<user_id>",
          "room": "room_b",
          "roomSid": "room_sid_b"
        }
      ]
    }
  },
  {
    "level": "error",
    "data": {
      "arguments": [
        "unable to set offer",
        {
          "fields": "[Object]",
          "identity": "<user_id>",
          "room": "room_b",
          "roomSid": "room_sid_b"
        }
      ]
    }
  }
]
```
```


### System Info

```shell
Browser: any, but mostly Chrome
OS: Windows / MacOS
```

### Severity

annoyance

### Additional Information

_No response_

lukasIO:
closes #1050 

Previously an exception thrown in `pc.setConfiguration` was treated as a `SignalReconnectError` which in turn meant the engine kept trying to resume, while a full reconnect would be the appropriate measure.

### DIFF
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -140,8 +140,6 @@ export class SignalClient {
 
   onLeave?: (leave: LeaveRequest) => void;
 
-  onReconnectResponse?: (response: ReconnectResponse) => void;
-
   connectOptions?: ConnectOpts;
 
   ws?: WebSocket;
@@ -359,8 +357,11 @@ export class SignalClient {
               this.startPingInterval();
               if (resp.message?.case === 'reconnect') {
                 resolve(resp.message.value);
-                this.onReconnectResponse?.(resp.message.value);
               } else {
+                this.log.debug(
+                  'declaring signal reconnected without reconnect response received',
+                  this.logContext,
+                );
                 resolve(undefined);
                 shouldProcessMessage = true;
               }
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -216,7 +216,7 @@ export class SignalClient {
     token: string,
     sid?: string,
     reason?: ReconnectReason,
-  ): Promise<ReconnectResponse | void> {
+  ): Promise<ReconnectResponse | undefined> {
     if (!this.options) {
       this.log.warn(
         'attempted to reconnect without signal options being set, ignoring',
@@ -242,7 +242,7 @@ export class SignalClient {
     token: string,
     opts: ConnectOpts,
     abortSignal?: AbortSignal,
-  ): Promise<JoinResponse | ReconnectResponse | void> {
+  ): Promise<JoinResponse | ReconnectResponse | undefined> {
     this.connectOptions = opts;
     url = toWebsocketUrl(url);
     // strip trailing slash
@@ -252,7 +252,7 @@ export class SignalClient {
     const clientInfo = getClientInfo();
     const params = createConnectionParams(token, clientInfo, opts);
 
-    return new Promise<JoinResponse | ReconnectResponse | void>(async (resolve, reject) => {
+    return new Promise<JoinResponse | ReconnectResponse | undefined>(async (resolve, reject) => {
       const unlock = await this.connectionLock.lock();
       try {
         const abortHandler = async () => {
@@ -358,7 +358,7 @@ export class SignalClient {
               if (resp.message?.case === 'reconnect') {
                 resolve(resp.message?.value);
               } else {
-                resolve();
+                resolve(undefined);
                 shouldProcessMessage = true;
               }
             } else if (this.isEstablishingConnection && resp.message.case === 'leave') {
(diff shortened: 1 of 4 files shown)


src/room/RTCEngine.ts
```diff
@@ -1029,6 +1025,11 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
     this.emit(EngineEvent.SignalResumed);
 
+    if (res) {
```
lukasIO, line 1028: I think we should additionally make sure somehow that we actually have a reconnect response at this point.

boks1971 (approved):

lukasIO:
Added a maximum of 2s waiting period for ReconnectResponse to arrive after signal reconnection has re-established.

src/api/SignalClient.ts
```diff
@@ -356,9 +358,10 @@ export class SignalClient {
               abortSignal?.removeEventListener('abort', abortHandler);
               this.startPingInterval();
               if (resp.message?.case === 'reconnect') {
-                resolve(resp.message?.value);
+                resolve(resp.message.value);
```
davidzhao, line 361: I'm not following this piece. why do we need to fire off a callback vs just resolving the promise? It seems that the promise is resolved either way with the same value?

lukasIO: it can happen that the promise gets resolved a couple of lines down with `undefined` if the message that comes in is not of type `reconnect`. So we'd miss the reconnect response when just looking at the return value of that resolving promise.

src/api/SignalClient.ts
```diff
@@ -356,9 +358,10 @@ export class SignalClient {
...
+                this.onReconnectResponse?.(resp.message.value);
               } else {
-                resolve();
+                resolve(undefined);
```
lukasIO, line 364: for context to the above comment:

this is line is the culprit. in a reconnect scenario we would resolve a signal reconnect promise with `undefined` instead of the reconnect response, as we consider the first incoming message as a successful signal reconnect.

davidzhao: I see, this is handling the case that reconnect is not the most immediate message.

When does that happen? this seems like a rather unexpected behavior.. I think the protocol indicates that Reconnect would need to be the first thing that's sent.. just like JoinResponse.

lukasIO: definitely mostly a safe guard. 
if it's unexpected behaviour, we can just throw an error instead of another signal message arrives? 
We'll have to figure out server version support for reconnect response though.

davidzhao: yeah, so I'm thinking if we can throw an error if ReconnectResponse comes later than expected. That should trigger a full reconnect sequence.

if the server is sending ReconnectResponse after the initial message, it would be breaking spec. We wouldn't want other clients to implement that behavior.

lukasIO: sounds good. Though we cannot reject the promise/throw an error per se after we've resolved the promise already. That will just result in an unhandled exception. 
We could introduce a new error event/callback to handle that. 

Will remove the late response handling parts from the PR for now.

davidzhao: cool, sounds good! we can add that separately

PR #1061, opened 2024-03-05
Set logExtension on all livekit loggers if not specified

lukasIO:
We shifted to more granular (class based) loggers.
This PR sets the `LogExtension` on all those class based loggers by default (instead of just on one), unless a specific logger gets passed to `setLogLevel` as a second argument.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -36,17 +36,11 @@ export type StructuredLogger = log.Logger & {
 };
 
 let livekitLogger = log.getLogger('livekit');
+const livekitLoggers = Object.values(LoggerNames).map((name) => log.getLogger(name));
 
 livekitLogger.setDefaultLevel(LogLevel.info);
 
 export default livekitLogger as StructuredLogger;
-
-function getLiveKitLoggers() {
-  const loggers = Object.entries(log.getLoggers()).filter(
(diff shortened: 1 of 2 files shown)


src/logger.ts
```diff
@@ -67,24 +72,28 @@ export type LogExtension = (level: LogLevel, msg: string, context?: object) => v
  * use this to hook into the logging function to allow sending internal livekit logs to third party services
  * if set, the browser logs will lose their stacktrace information (see https://github.com/pimterry/loglevel#writing-plugins)
  */
-export function setLogExtension(extension: LogExtension, logger = livekitLogger) {
-  const originalFactory = logger.methodFactory;
+export function setLogExtension(extension: LogExtension, logger?: StructuredLogger) {
+  const loggers = logger ? [logger] : getLiveKitLoggers();
+
+  loggers.forEach((logR) => {
+    const originalFactory = logR.methodFactory;
```
boks1971, line 79: Where is this used? From the looks of it, feels like this should restore the method factory, but did not see it.

boks1971 (approved):

src/logger.ts
lukasIO: the approach is taken directly from https://github.com/pimterry/loglevel?tab=readme-ov-file#writing-plugins

PR #1063, opened 2024-03-05
Handle errors thrown in signal client callbacks

HermanBilous:


lukasIO (commented):
catching those is a good call, but they should trigger an action to try to fix things. The right call would likely be a full reconnect with 

```ts
this.fullReconnectOnNext = true;
this.handleDisconnect(errorDescription);
```

PR #1064, opened 2024-03-06
Throw error if trying to connect with a non-compatible browser

davidliu:


src/room/utils.ts
```diff
@@ -133,6 +133,11 @@ export function supportsSetCodecPreferences(transceiver: RTCRtpTransceiver): boo
 }
 
 export function isBrowserSupported() {
+  // global defined only for ReactNative.
+  // @ts-ignore
+  if (isReactNative() && global && global.RTCPeerConnection === undefined) {
```
lukasIO, line 138:
```suggestion
  if (isReactNative() && typeof global !== 'undefined' && global.RTCPeerConnection === undefined) {
```

src/room/utils.ts
```diff
@@ -133,6 +133,11 @@ export function supportsSetCodecPreferences(transceiver: RTCRtpTransceiver): boo
...
+    return false;
+  }
   return supportsTransceiver() || supportsAddTrack();
```
lukasIO, line 141: we whould probably add a `if(typeof RTCPeerConnection === 'undefined') return false` here too to catch cases where browser extensions disabled/removed it for security reasons

src/room/Room.ts
```diff
@@ -161,6 +163,14 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
    */
   constructor(options?: RoomOptions) {
     super();
+
+    if (!isBrowserSupported()) {
+      if (isReactNative()) {
+        throw Error("react-native-webrtc isn't detected, have you called registerGlobals?");
+      } else {
+        throw Error("LiveKit isn't supported on this browser.");
```
lukasIO, line 171: as mentioned in the other comment, this might also be due to some privacy browser extension so we could give a hint about that to users here.

lukasIO: not sure, maybe something like `LiveKit doesn't seem to be supported on this browser. Try to update your browser and make sure no browser extensions are disabling webRTC.` ?

lukasIO (approved):
nice!

PR #1069, opened 2024-03-13
Fix FPS and latency bugs with VP9 screenshare.

davidzhao:
Also restored L3T3_KEY publishing for screen share. It was disabled previously.

Chrome does not allow more than 5 fps with L1T3, and it has encoding bugs with L3T3 It has a different path for screenshare handling and it seems to be untested/buggy As a workaround, we are setting contentHint to force it to go through the same path as regular camera video. While this is not optimal, it delivers the performance that we need.

The problem is [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/video_coding/codecs/vp9/svc_config.cc;drc=b8a4daa31c86ef61f7a78ea73711b86ab732db32;l=31). where layer 0 always gets capped at 5fps, and it doesn't get reset to 30 even if there's [only a single layer](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/video_coding/codecs/vp9/svc_config.cc;l=207;drc=b8a4daa31c86ef61f7a78ea73711b86ab732db32) published.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -736,7 +736,16 @@ export default class LocalParticipant extends Participant {
         if (isSVCCodec(videoCodec)) {
           // vp9 svc with screenshare has problem to encode, always use L1T3 here
           if (track.source === Track.Source.ScreenShare && videoCodec === 'vp9') {
-            opts.scalabilityMode = 'L1T3';
+            // Chrome does not allow more than 5 fps with L1T3, and it has encoding bugs with L3T3
+            // It has a different path for screenshare handling and it seems to be untested/buggy
+            // As a workaround, we are setting contentHint to force it to go through the same
+            // path as regular camera video. While this is not optimal, it delivers the performance
+            // that we need
+            if ('contentHint' in track.mediaStreamTrack) {
+              track.mediaStreamTrack.contentHint = 'motion';
+            } else {
+              opts.scalabilityMode = 'L1T3';
+            }
           }
           // set scalabilityMode to 'L3T3_KEY' by default
           opts.scalabilityMode = opts.scalabilityMode ?? 'L3T3_KEY';
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -743,6 +743,10 @@ export default class LocalParticipant extends Participant {
             // that we need
             if ('contentHint' in track.mediaStreamTrack) {
               track.mediaStreamTrack.contentHint = 'motion';
+              this.log.info('forcing contentHint to motion for screenshare with VP9', {
+                ...this.logContext,
+                ...getLogContextFromTrack(track),
+              });
             } else {
               opts.scalabilityMode = 'L1T3';
             }
--- a/.changeset/smooth-glasses-jam.md
+++ b/.changeset/smooth-glasses-jam.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Fix FPS and latency issues with VP9 screenshare
--- a/src/room/track/RemoteAudioTrack.ts
+++ b/src/room/track/RemoteAudioTrack.ts
@@ -1,6 +1,6 @@
 import { TrackEvent } from '../events';
-import { computeBitrate } from '../stats';
 import type { AudioReceiverStats } from '../stats';
+import { computeBitrate } from '../stats';
 import type { LoggerOptions } from '../types';
 import { isReactNative, supportsSetSinkId } from '../utils';
 import RemoteTrack from './RemoteTrack';


src/room/participant/LocalParticipant.ts
```diff
@@ -736,7 +736,16 @@ export default class LocalParticipant extends Participant {
         if (isSVCCodec(videoCodec)) {
           // vp9 svc with screenshare has problem to encode, always use L1T3 here
           if (track.source === Track.Source.ScreenShare && videoCodec === 'vp9') {
-            opts.scalabilityMode = 'L1T3';
+            // Chrome does not allow more than 5 fps with L1T3, and it has encoding bugs with L3T3
```
boks1971, line 739: I think it allows 30 fps. The array in Chrome code is for three temporal layers. If we do `T3`, it should do 30 fps if I am not mistaken (that layer array is for temporal layers as it is FPS). But, my interpretation could be incorrect.

boks1971 (approved):

src/room/participant/LocalParticipant.ts
davidzhao: yes, it would if L3T3 functioned correctly. I think both you and @cnderrauber confirmed that it does not?

davidzhao: From what I can see, they are using spatial layer as the index into the array: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/video_coding/codecs/vp9/svc_config.cc;l=63;drc=b8a4daa31c86ef61f7a78ea73711b86ab732db32

So I think the T3 is indicating 3 temporal bounded by selected maxFramerate

boks1971: I could not get either `L3T3` or `L3T3_KEY` to work. I must have done something incorrect with `L3T3_KEY` as it is working for you.

I meant that the array is setting FPS for temporal layers. With `L1T3`, it should still be one spatial and three temporal layers. So, it should do 30 fps and allow temporal scale down to lower frame rate. I am confused as to how it is limiting to only at 5 fps (which looks like temporal layer 0 in that array). But, if it is actually doing three temporal layers with one spatial, it should still do 30 fps. That's how I was reading that array.

boks1971: > From what I can see, they are using spatial layer as the index into the array: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/video_coding/codecs/vp9/svc_config.cc;l=63;drc=b8a4daa31c86ef61f7a78ea73711b86ab732db32
> 
> So I think the T3 is indicating 3 temporal bounded by selected maxFramerate

oh okay, that is interesting.

davidzhao: does this branch work for you? the main difference that's needed is setting `contentHint` to motion.. which bypasses the screenshare code path in libwebrtc

boks1971: Trying now.

boks1971: > > From what I can see, they are using spatial layer as the index into the array: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/video_coding/codecs/vp9/svc_config.cc;l=63;drc=b8a4daa31c86ef61f7a78ea73711b86ab732db32
> > So I think the T3 is indicating 3 temporal bounded by selected maxFramerate
> 
> oh okay, that is interesting.

Did not realise that the fps is scoped to the spatial layer. Sorry, my bad.

boks1971: > does this branch work for you? the main difference that's needed is setting `contentHint` to motion.. which bypasses the screenshare code path in libwebrtc

Definitely much better. Seemed to start off with low fps/high latency, but got much smoother after a few seconds.

boks1971: The initial choppiness could be due to some server side changes I am trying to test.

src/room/participant/LocalParticipant.ts
```diff
@@ -736,7 +736,16 @@ export default class LocalParticipant extends Participant {
...
+            // It has a different path for screenshare handling and it seems to be untested/buggy
+            // As a workaround, we are setting contentHint to force it to go through the same
+            // path as regular camera video. While this is not optimal, it delivers the performance
+            // that we need
+            if ('contentHint' in track.mediaStreamTrack) {
+              track.mediaStreamTrack.contentHint = 'motion';
+            } else {
```
lukasIO, line 746: not sure I follow this condition:
we only set `L1T3` if the track does _not_ support contentHint ? 
My assumption (might be wrong) was that we would want L1T3 regardless for every screen share?

src/room/participant/LocalParticipant.ts
```diff
@@ -736,7 +736,16 @@ export default class LocalParticipant extends Participant {
         if (isSVCCodec(videoCodec)) {
           // vp9 svc with screenshare has problem to encode, always use L1T3 here
           if (track.source === Track.Source.ScreenShare && videoCodec === 'vp9') {
-            opts.scalabilityMode = 'L1T3';
+            // Chrome does not allow more than 5 fps with L1T3, and it has encoding bugs with L3T3
+            // It has a different path for screenshare handling and it seems to be untested/buggy
+            // As a workaround, we are setting contentHint to force it to go through the same
```
lukasIO, line 741: Should we scope that to `if(isChrome())`?

src/room/participant/LocalParticipant.ts
```diff
@@ -736,7 +736,16 @@ export default class LocalParticipant extends Participant {
...
+            // path as regular camera video. While this is not optimal, it delivers the performance
+            // that we need
+            if ('contentHint' in track.mediaStreamTrack) {
+              track.mediaStreamTrack.contentHint = 'motion';
```
lukasIO, line 745: If users have set an explicit contentHint and we end up overriding their setting, we should probably log an info (or warning) log to let them know this is happening.

lukasIO (commented):
a couple of nits and questions

src/room/participant/LocalParticipant.ts
lukasIO: ah, re-read the PR description and understood the problem. Ignore above

src/room/participant/LocalParticipant.ts
boks1971: We did see poor screen share quality from Safari in a meet. Not sure if the underlying reason is the same.

lukasIO: true. I did some quick testing of sharing youtube content on Safari with current livekit-client and it seemed to perform fine. Latency was maybe a bit high, but FPS looked fine

lukasIO (approved):

PR #1077, opened 2024-03-15
Fix quality issues with SVC screenshares.

davidzhao:
When we munged SDP to set bitrate, it was set on the codec's fmtp line.
The codec and its fmtp is shared across media tracks. So if a user is sharing
a camera and screenshare tracks, both tracks will be published at the
same bitrate (typically the user's camera track). This leads to blurry
screenshares.

Additionally, screenshares cannot be encoded in VP9 with L3T3_KEY. Chrome
degrades quality aggressively whenever L1T3 isn't used.

### DIFF
--- a/.changeset/wet-kids-unite.md
+++ b/.changeset/wet-kids-unite.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Fix quality issues with SVC screenshares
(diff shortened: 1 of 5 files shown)


lukasIO:
nice findings!

lukasIO (approved):

PR #1084, opened 2024-03-20
Fix share desktop picture using av1 codec made browser page crash

zesun96:
details in https://github.com/livekit/livekit/issues/2569


davidzhao:
I don't think this is the right fix.. this will prevent all browsers from being able to use AV1.

which version of the client SDK did you see this error with?

zesun96:
Mabye it is a webrtc bugs? I can't be sure.
The [example](https://meet.livekit.io/rooms/kwlt-5iuf?codec=av1) can be triggered.

PR #1087, opened 2024-03-21
Signal local audio track feature updates

lukasIO:
afaik there's no way to listen to events that relate to media stream track constraint/settings changes. 

The two options I see are either proxy the call to `applyConstraints` and force an update on track restarts or set up an interval to regularly check for changes.

### DIFF
--- a/src/room/track/LocalAudioTrack.ts
+++ b/src/room/track/LocalAudioTrack.ts
@@ -1,3 +1,4 @@
+import { AudioTrackFeature } from '@livekit/protocol';
 import { TrackEvent } from '../events';
 import { computeBitrate, monitorFrequency } from '../stats';
 import type { AudioSenderStats } from '../stats';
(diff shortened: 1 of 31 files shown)


src/room/track/LocalAudioTrack.ts
```diff
@@ -175,6 +207,14 @@ export default class LocalAudioTrack extends LocalTrack<Track.Kind.Audio> {
       this.processor = processor;
       if (this.processor.processedTrack) {
         await this.sender?.replaceTrack(this.processor.processedTrack);
+        this.processor.processedTrack.addEventListener(
```
davidzhao, line 210: do we need to remove these from the old processor instance? incase it was replaced?

davidzhao (approved):
lg!

src/room/track/LocalAudioTrack.ts
lukasIO: the `processedTrack` will be a newly generated MediaStreamTrack whenever the processor changes. 
This track will also be teared down once the processor stops, so I think it should be fine without explicitly removing the listeners beforehand

PR #1091, opened 2024-03-23
Create processorElement before processor init

xdef:
The processorElement should be created before calling the processor initialization function.
Otherwise we always get the error: "Currently only video transformers are supported" because the element is undefined


lukasIO (approved):
thanks!

PR #1093, opened 2024-03-27
Use legacy SVC encoding specification for React-Native

davidzhao:
Even tho RN runs M114, it does not produce SVC layers when a single encoding is provided.
So we'll use the legacy SVC specification for now.


lukasIO (approved):
lgtm

davidliu:
Whoops, missed this PR. Will run tests tomorrow.

davidliu (approved):
Tested and creates the extra layers with the patch.

PR #1103, opened 2024-04-09
Add client connection time metrics

lukasIO:
starting with a minimal set of timings and errors. 
This is not considered completed, would value some feedback WRT what else we'll want to send.


toger5:
To understand this right, the connection times/durations will be sent through the websocket connection to the SFU. The SFU probably uses it to build the dashboard?
Is there a way to locally emit this data with js events from the livekit client?
So that one could use different ways to analyze/visualize this data? (for example posthog)

` publisher?: number;`, ` subscriber?: number;` refers to the sending and receiving peer connection durations right?

lebaudantoine:
This work seems super interesting to me, is it stale ?
(I am also using Posthog in my product)

PR #1106, opened 2024-04-15
Force playback after app visibility changes back to visible

lukasIO:
On Safari sometimes the video element failed to auto play again after the browser came back from being in the background. 
In order to work around this, we force `play` to be called on all attached elements once the app comes back into the foreground.


davidzhao (approved):
lgtm!

PR #1107, opened 2024-04-16
Await data publisher connection without multiple negotiations

lukasIO:
We had user reports of `maxListeners exceeded` when they were using e.g. user input events to trigger data publishing. 

This was due to the fact that in those handlers users would typically do something like

```ts
document.addEventListener('mousemove', async (ev) => { await room.localParticipant.publishData(...); }
```

This could trigger multiple negotiation requests while the publisher data connection is establishing as the calls to publishData could be happening in parallel. 

To make this easier for users to handle we keep an internal reference to the publisher connection promise and await that.


davidzhao (approved):
lgtm!

PR #1114, opened 2024-04-23
Remove internal calls to setCodecPreferences on sender transceivers

lukasIO:
As recent issues with Chrome 124 arose, we remove all calls to `setCodecPreferences` on publish transceivers and let the server dictate the selected codec by order hierarchy.
This is in accordance with the spec, as `setCodecPreferences` is meant to be a receiver only setting, see [here](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setcodecpreferences)

### DIFF
--- a/.changeset/slimy-flowers-travel.md
+++ b/.changeset/slimy-flowers-travel.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Remove internal calls to setCodecPreferences on senders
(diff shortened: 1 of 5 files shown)


lukasIO:
>  Is there anything to remove from sample app?

not that I'm aware of, selecting a preferred codec client side will still work, but merely by signalling the preferred codec to the server instead of trying to set it on the transceiver.

davidzhao (approved):
nice! love it when deleting code makes it work better!

PR #1117, opened 2024-04-25
Disable simulcast for all iOS browsers when E2EE is enabled

lukasIO:
Additionally to this fix, it makes sense to re-evaluate whether the restriction (simulcast + E2EE) is still not working in recent Safari versions. 
For iOS it will be tricky to differentiate that based on browser version though as we'll need to get insight on the actual Safari version rather than the Firefox/Chrome version.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -11,6 +11,7 @@ import {
   UserPacket,
 } from '@livekit/protocol';
 import type { InternalRoomOptions } from '../../options';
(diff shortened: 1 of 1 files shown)


lukasIO:
found the related webkit bug https://bugs.webkit.org/show_bug.cgi?id=257803. 
The fix got released as part of Safari v17.2, so everything beyond that should be fine. 
Added iOS version parsing in order to determine the underlying Safari version on chrome and firefox iOS.

davidzhao (approved):
lgtm! this is a good find

PR #1118, opened 2024-04-26
Don't create data channel of publisher until sending data message

cnderrauber:
~~Will not merge this change until sfu server with https://github.com/livekit/livekit/pull/2686 updated as the old server version always requires data channel~~
Enable this optimization if server supports optional publisher datachannel (protocol > 13)

### DIFF
--- a/.changeset/clean-gifts-applaud.md
+++ b/.changeset/clean-gifts-applaud.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Don't create data channel of publisher until sending data message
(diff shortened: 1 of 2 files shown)


cnderrauber:
> But, I think this requires a protocol version bump to ensure backwards compatibility.

This is a client change so protocol bump doesn't work for it, the new server can work with old/new clients.

> Also, all clients need to do this to take advantage of this optimisation right?

Right

lukasIO:
> as the old server version always requires data channel

what does that mean in terms of backwards compatibility of a client with this change trying to connect to an older server verison?

cnderrauber:
> > as the old server version always requires data channel
> 
> what does that mean in terms of backwards compatibility of a client with this change trying to connect to an older server verison?

The publisher peerconnection will be considered as failed by sfu.

lukasIO (approved):

PR #1119, opened 2024-04-26
Add support for transcription handling

lukasIO:


davidzhao (approved):
lgtm!

src/room/Room.ts
```diff
@@ -2071,6 +2084,11 @@ export type RoomEventCallbacks = {
     kind?: DataPacket_Kind,
     topic?: string,
   ) => void;
+  transcriptionReceived: (
```
davidzhao, line 2087: thinking about this a bit more.. wondering if we should expose each segment separately here. perhaps with a native type instead of the protobuf obj? 

Maybe something like:

```
interface TranscriptionSegment {
  id
  text
  start_time
  end_time
  language
  is_final
}
````

Comment (lukasIO) [reply src/room/Room.ts:2087]: makes sense to me

Comment (davidzhao) [reply src/room/Room.ts:2095]: right.. agreed it's a bit awkward.. but it depends on the sender then? i.e. why is it sending segments that have not changed?

Comment (davidzhao) [reply src/room/Room.ts:2095]: or maybe it should fire with a list of segments? instead of multiple callbacks?

File: src/room/types.ts [hunk 2]
```diff
@@ -55,3 +55,12 @@ export type LoggerOptions = {
   loggerName?: string;
   loggerContextCb?: () => Record<string, unknown>;
 };
+
+export interface TranscriptionSegment {
```
davidzhao, line 59: we'd need to export this type in index.ts

src/room/Room.ts
lukasIO: > or maybe it should fire with a list of segments? instead of multiple callbacks?

I think that's a neat idea!

src/room/track/RemoteTrack.ts
```diff
@@ -77,7 +77,20 @@ export default abstract class RemoteTrack<
...
       this.monitorInterval = setInterval(() => this.monitorReceiver(), monitorFrequency);
     }
+    this.registerTimeSyncUpdate();
   }
 
   protected abstract monitorReceiver(): void;
+
+  registerTimeSyncUpdate() {
+    const loop = () => {
+      this.timeSyncHandle = requestAnimationFrame(() => loop());
```
lukasIO, line 87: this registers a loop in `startMonitor` to continuously update the current `rtpTimestamp` of the track. If the value changes it also emits a `TrackEvent.TimeSyncUpdate` with the latest value

davidzhao (approved):
lgtm

davidzhao:
awesome!

PR #1130, opened 2024-05-07
Support SIP DTMF data messages

dennwc:
Added support for SIP DTMF data messages on rooms and participants.

For consistency with other SDKs, the `RTCEngine` data packet handler signature was changed to receive all message types, which is then switched by type in the Room.

Also compatibility code was added to properly select participant from a shared `participant_identity` on the `DataPacket` (see https://github.com/livekit/protocol/pull/706).

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -26,6 +26,7 @@ import {
   TrackUnpublishedResponse,
   Transcription,
   UpdateSubscription,
+  type UserPacket,
 } from '@livekit/protocol';
 import { EventEmitter } from 'events';
 import type { MediaAttributes } from 'sdp-transform';
@@ -650,8 +651,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       } else {
         if (dp.value?.case === 'user') {
           // compatibility
-          compatString(dp, dp.value.value, 'participantIdentity');
-          compatArray(dp, dp.value.value, 'destinationIdentities');
+          applyUserDataCompat(dp, dp.value.value);
         }
         this.emit(EngineEvent.DataPacketReceived, dp);
       }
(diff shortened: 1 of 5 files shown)


src/room/Room.ts
```diff
@@ -1466,24 +1465,40 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     pub.setSubscriptionError(update.err);
   };
 
-  private handleDataPacket = (userPacket: UserPacket, kind: DataPacket_Kind) => {
+  private handleDataPacketAny = (packet: DataPacket) => {
```
lukasIO, line 1470: nit: can we rename this to `handleDataPacket` and the current `handleDataPacket` to `handleUserPacket` ?

src/room/RTCEngine.ts
```diff
@@ -648,10 +647,37 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
         // dispatch speaker updates
         this.emit(EngineEvent.ActiveSpeakersUpdate, dp.value.value.speakers);
-      } else if (dp.value?.case === 'user') {
-        this.emit(EngineEvent.DataPacketReceived, dp.value.value, dp.kind);
-      } else if (dp.value?.case === 'transcription') {
-        this.emit(EngineEvent.TranscriptionReceived, dp.value.value);
+      } else {
+        if (dp.value?.case === 'user') {
+          // compatibility
+          if (dp.participantIdentity === '' && dp.value.value.participantIdentity !== '') {
```
lukasIO, line 653: nice! 
just to make it a bit easier to maintain, can we extract the compatibility stuff into a function? 
it seems the pattern is always `dp[myProperty] || dp.value.value[myProperty]` (empty strings are falsy in JS). Not sure if it's still elegant with typescript stuff, but having it extracted into a helper of some sort (not necessarily the one I proposed) would be nice!

lukasIO (changes requested):

src/room/RTCEngine.ts
lukasIO: > But in general, sure! I assume empty arrays are falsy too, right?

no, empty arrays are truthy :D my naive proposal won't work for both strings and arrays...

dennwc: I added the helper functions you mentioned, but I get some TS error about `DataPacket` type not supporting indexing. I might need your help on this one.

lukasIO: I had a look and my initial idea might be totally flawed. I don't really see a way to get around the typescript issues without either being overly specific about the allowed strings or converting the protobuf to JSON first and then parsing it back. Which seems pointless and not at all simplifying things. 
While verbose, your original implementation of handling these separately looks a lot preferable to me than back and forth parsing. 
I think the main reason why I pushed for that was because we had logic duplication for transcription and user packet, but it seems that is no longer the case, because the field got renamed for transcriptions to `transcribed_participant`. So I don't see a huge incentive anymore to change what you had originally.

I pushed to a branch what would work, but that's basically just extracting what you had previously already into a function 🤦 https://github.com/livekit/client-sdk-js/commit/42b8381cecabf2c32ab8c93ee1f0a317f769d44b
Feel free to ignore it entirely.

dennwc: Yeah, that looks very specific. OK, lets settle on the middle ground - I'll revert the "generic" part of the change, but will keep compat logic in a separate function, so that it's easier to read the main handler function.

dennwc: Done, pushed the updated function. Also simplified logic a bit.

lukasIO (approved):
nice!

PR #1131, opened 2024-05-08
Honor calls to track.stop() during an ongoing restart attempt

lukasIO:
closes https://github.com/livekit/components-js/issues/851

When calling `track.stop()` while a `track.restart()` is already ongoing, we'd end up with a dangling active mediastreamtrack. 

This change makes sure that if `stop` is called during a `restart` we immediately stop the track again after the restart. 

Alternatives considered:
An arguably cleaner solution would be to make the `stop` function async and await `restartLock.lock`, but that would potentially be a breaking change.

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -299,6 +299,7 @@ export default abstract class LocalTrack<
 
       await this.setMediaStreamTrack(newTrack);
       this._constraints = constraints;
+      this.emit(TrackEvent.Restarted, this);
       if (this.manuallyStopped) {
         this.log.warn(
           'track was stopped during a restart, stopping restarted track',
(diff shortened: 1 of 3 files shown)


src/room/track/LocalTrack.ts
```diff
@@ -296,7 +299,13 @@ export default abstract class LocalTrack<
...
       this._constraints = constraints;
-
+      if (this.manuallyStopped) {
+        this.log.warn(
+          'track was stopped during a restart, stopping restarted track',
+          this.logContext,
+        );
+        this.stop();
+      }
       this.emit(TrackEvent.Restarted, this);
```
lukasIO, line 309: not sure if we should still emit `TrackEvent.Restarted` if the track has been stopped. For API consistency, I'd argue it's best to still do it.

davidzhao (approved):
lg!

PR #1132, opened 2024-05-10
Ensure DD ext for svc codecs

cnderrauber:


src/room/participant/publishUtils.ts
```diff
@@ -171,6 +171,11 @@ export function computeVideoEncodings(
       });
     }
 
+    if (original.encoding.priority) {
+      encodings[0].priority = original.encoding.priority;
```
lukasIO, line 175: should we set this for every encoding if it's not using legacy SVC mode?

cnderrauber: non-legacy svc mode only has 1 encoding, do you mean legacy svc mode? it only takes first encoding paramter and use other encodings to figure out the count of spatial layers

davidzhao: good catch.. looks like priority was not being set for SVC previously.

davidzhao (approved):

lukasIO (approved):

PR #1137, opened 2024-05-13
Add leave req full reconnect simulation scenario

lukasIO:


example/index.html
```diff
@@ -172,6 +172,7 @@ <h2>Livekit Sample App</h2>
                 <option value="force-tls">Force TURN/TLS</option>
                 <option value="subscribe-all">Subscribe all</option>
                 <option value="unsubscribe-all">Unsubscribe all</option>
+                <option value="leave-full-reconnect">Leave Request Full Reconnect</option>
```
boks1971, line 175: This will not work with a locally running server I think. Have not check what region settings returns when running server locally.

boks1971 (approved):

example/index.html
boks1971: Yeah, that's right.

I think it will not work even if running cloud locally as we have all simulated regions and I am not sure if regions settings things works correctly. Not a big deal. Just a note. I will have a look later if this works in sample app and may be fix it if possible when running cloud locally too.

PR #1138, opened 2024-05-14
Add degradationPreference option for LocalVideoTrack

lukasIO:
- adds `localVideoTrack.setDegradationPreference` 
- adds `degradationPreference` to `TrackPublishOptions`
- computes a default `degradationPreference` if none is set explicitly

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -858,7 +858,7 @@ export default class LocalParticipant extends Participant {
 
     track.sender = await this.engine.createSender(track, opts, encodings);
 
-    if (track instanceof LocalVideoTrack && opts.degradationPreference) {
+    if (track instanceof LocalVideoTrack) {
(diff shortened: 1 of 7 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -857,6 +858,11 @@ export default class LocalParticipant extends Participant {
 
     track.sender = await this.engine.createSender(track, opts, encodings);
 
+    if (track instanceof LocalVideoTrack && opts.degradationPreference) {
```
cnderrauber, line 861: do we need to check opts.degradationPreference here? or just `if (track instanceof LocalVIdeoTrack)`

lukasIO: thanks, this is a leftover, you're right, we should get rid of the second check

cnderrauber (approved):

davidzhao (approved):
lgtm!

PR #1139, opened 2024-05-14
Account for spatial layers in default video encoding for SVC codecs

lukasIO:
I couldn't find any analysis on how much additional bitrate a spatial layer adds for SVC, going with cube root is more of a gut feeling. 
Please advise on what a better factor for this could be.


cnderrauber (approved):

davidzhao:
I'm not sure about this one.. let's think through and discuss?

PR #1143, opened 2024-05-16
Allow processors to be set as part of track publish options

lukasIO:
this allows users to specify processors that should be applied to tracks during publishing. 
This helps to solve use cases where users would like to start the initial publication with processors enabled already.

```ts
      videoCaptureDefaults: {
        resolution: VideoPresets.h720.resolution,
        processor: BackgroundBlur(),
      },
      audioCaptureDefaults: {
        processor: KrispNoiseFilter(),
      },
```

### DIFF
--- a/.changeset/eight-pugs-cough.md
+++ b/.changeset/eight-pugs-cough.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": minor
+---
+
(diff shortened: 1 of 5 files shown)


davidzhao (approved):
lgtm

PR #1150, opened 2024-05-30
Add support for ParticipantKind

lukasIO:


src/room/participant/Participant.ts
```diff
@@ -46,6 +47,8 @@ function qualityFromProto(q: ProtoQuality): ConnectionQuality {
   }
 }
 
+export { ParticipantKind };
```
davidzhao, line 50: we'd also need to export from index.ts.

davidzhao (approved):
lg!

PR #1157, opened 2024-06-07
Correct track replacement when resuming local track upstream

kyleparrott:
Currently if a track processor is applied to a video track while the track's upstream is paused (ie. `pauseUpstream()` has been called on it), the next call to resume upstream will replace the processed track with the plain, unprocessed track. This should correct that logic and prevent this incorrect replacement.

My app was performing the following steps when I ran into this behavior:

1. Publish a video track to the room with no processors applied
2. Pause the local video track's upstream
3. Apply a Track processor to the local track
4. Resume the local video track's upstream

I am pausing and resuming the track when I am making processor changes to prevent showing any "flashes" of unprocessed frames when changing between virtual background options.

### DIFF
--- a/.changeset/kind-nails-battle.md
+++ b/.changeset/kind-nails-battle.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Fix resumeUpstream with local track processors enabled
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
Nice, thanks for catching this!

PR #1158, opened 2024-06-10
Add RoomEvent.SignalReconnecting and ConnectionState.SignalReconnecting

lukasIO:


src/room/Room.ts
```diff
@@ -88,9 +88,10 @@ export enum ConnectionState {
   Connecting = 'connecting',
   Connected = 'connected',
   Reconnecting = 'reconnecting',
+  Resuming = 'resuming',
 }
 
-const connectionReconcileFrequency = 2 * 1000;
+const connectionReconcileFrequency = 4 * 1000;
```
lukasIO, line 94: Drive by change. 
currently with a 2s timeout the max allowed time for a publisher peer connection to establish is 6 seconds, as the peer connection manager will return a "connecting" state while publishing connection is establishing. Increasing the timeout to 12s seems reasonable to me.

PR #1162, opened 2024-06-12
Fix normalising of default device Ids in DeviceManager

lukasIO:
and add some tests

### DIFF
--- a/.changeset/dry-islands-hug.md
+++ b/.changeset/dry-islands-hug.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Fix normalising of default device Ids in DeviceManager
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
this lgtm.. though I wanted to bring up a potential drawback of normalizing device IDs.

if we left it as `default`, and the user connects to a bluetooth device, I believe default would start pointing to the "correct" device.. when the user publishes from that device.

however, if we normalized it, and remembered the device id, it would mean that when the system default changed, we would be pointing to the old default (which is incorrect)

lukasIO:
yeah, agree in general. However we currently only normalize anyways for `webAudioMix: true`. Setting `default` on an AudioContext throws an error, so we are somewhat forced to normalize it in that case.

PR #1168, opened 2024-06-20
Make metadata updates async and throw after timeout

lukasIO:
this makes metadata, name and attribute updates async and queues subsequent requests until a response has been received.


davidzhao (approved):
lgtm!

PR #1170, opened 2024-06-21
Add support for participant attributes

lukasIO:
after this gets merged, I'll update #1168 to include request error handling also for the attribute updates. 

Do we want a protocol bump for this so that clients can know whether or not the server supports attributes already?


davidzhao (approved):
lg!

PR #1180, opened 2024-07-05
Fix activeSpeakers has old participant  when participant sid changed

zesun96:


src/room/Room.ts
```diff
@@ -1407,6 +1407,10 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
   private handleSpeakersChanged = (speakerUpdates: SpeakerInfo[]) => {
     const lastSpeakers = new Map<string, Participant>();
     this.activeSpeakers.forEach((p) => {
+      const remoteParticipant = this.remoteParticipants.get(p.identity)!;
```
lukasIO, line 1410:
```suggestion
      const remoteParticipant = this.remoteParticipants.get(p.identity);
```

src/room/Room.ts
```diff
@@ -1407,6 +1407,10 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
+      if (remoteParticipant.sid != p.sid) {
```
lukasIO, line 1411:
```suggestion
      if (remoteParticipant && remoteParticipant.sid !== p.sid) {
```

lukasIO (commented):
Thanks, this makes sense, just two nits

lukasIO (approved):

PR #1181, opened 2024-07-05
Convenience method setAttribute to update a single key/value pair

davidzhao:


lukasIO:
Some thoughts:

- with https://github.com/livekit/client-sdk-js/pull/1168 still pending, we'll be in a situation where setting multiple attributes after one another will take a lot longer (as it waits for server ack)
- I'd be inclined to offer one primary way to update attributes, the current api with `setAttributes({myAttr});` actually seems to be almost (if not more concise) than explicitly requiring users to set the key separately as a string with `settAttrbute('myAttr', value)` 🤔

lukasIO (approved):

PR #1186, opened 2024-07-10
Include participant identity in CryptoError errors

hughns:


.changeset/cryptor-error-participant-identity.md
```diff
@@ -0,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Track participant identity for CryptoError errors
```
lukasIO, line 5: a tiny nit, because we use `Track` as a construct all over the SDK to avoid any confusion.
```suggestion
Include participant identity in CryptoError errors
```

lukasIO (commented):
This makes sense, thanks for the PR!

A tiny wording nit, but looks good to me otherwise

lukasIO (approved):

PR #1187, opened 2024-07-11
Make sure metadata check is async

lukasIO:


src/room/errors.ts
```diff
@@ -63,6 +65,15 @@ export class PublishDataError extends LivekitError {
   }
 }
 
+export class SignalRequestError extends LivekitError {
+  reason: ErrorResponse_Reason;
+
+  constructor(message: string, reason: ErrorResponse_Reason = ErrorResponse_Reason.UNKNOWN) {
+    super(15, message);
```
boks1971, line 72: what is the 15?

boks1971: oh the code, is there some way that it can collide with some other piece of code re-using it unintentionally? What is `code` used for?

boks1971 (approved):

src/room/errors.ts
lukasIO: `code` is our internal LiveKit error code. The `LivekitError` class has it as an error-identifier basically. We've been simply incrementing the `code` for every error we add, so I did the same for the `SignalRequestError`

PR #1188, opened 2024-07-11
Ensure permission request for listDevices works also in Firefox

lukasIO:
When calling `enumerateDevices` in Firefox it would return a list of inputs correctly populated, but omits the `audiooutput` mediakind. By filtering for the desired kind on the returned list we can make sure to request permissions if the desired kind is missing. 
This will make it possible to list and select audio output devices in Firefox.


davidzhao (approved):
lgtm

PR #1189, opened 2024-07-12
Fix wording in CryptorError debug log

zesun96:
little wrong but when I modify the debugging, the error confuses me.


lukasIO (approved):
thank you!

PR #1191, opened 2024-07-15
Emit TimesyncUpdates for local tracks

lukasIO:
Manually computing these for local tracks makes it easier to compute `activeTranscriptions` with the same API that we use for remote tracks. 
Downside is that we register this for every local track independently, but probably not a major concern.

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -54,6 +54,8 @@ export default abstract class LocalTrack<
 
   protected manuallyStopped: boolean = false;
 
+  protected lastTimeSyncUpdate = 0;
+
   private restartLock: Mutex;
 
   /**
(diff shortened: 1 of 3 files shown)


src/room/track/LocalTrack.ts
```diff
@@ -541,5 +543,23 @@ export default abstract class LocalTrack<
     this.emit(TrackEvent.TrackProcessorUpdate);
   }
 
+  protected registerTimeSyncUpdate() {
```
davidzhao, line 546: does this work for RN?

lukasIO: the browser APIs `requestAnimationFrame`, and `performance.x` should be available in RN. Haven't done any tests yet to confirm though. 
@davidliu had some concerns whether the contributing sources used for remote tracks would work in RN out of the box though

PR #1195, opened 2024-07-16
Specify pnpm version in package.json

lukasIO:
we have this field set in other JS repos already, client-sdk-js was lacking behind in that regard.


boks1971 (approved):
nice!!!

PR #1197, opened 2024-07-23
await element to be attached before calling .play on it

pabra:
otherwise Firefox 128 throws this error when 'setProcessor' is called on track: 
```
failed to play processor element
Object { room: undefined, roomID: undefined, participant: undefined, pID: undefined, trackID: undefined, source: "camera", muted: false, enabled: true, kind: "video", streamID: "{bbe7a482-313d-49e6-8d55-5dbe42d2c24f}", … } enabled: true
error: DOMException: The fetching process for the media resource was aborted by the user agent at the user's request. kind: "video"
muted: false
pID: undefined
participant: undefined
room: undefined
roomID: undefined
source: "camera"
streamID: "{bbe7a482-313d-49e6-8d55-5dbe42d2c24f}" streamTrackID: "{ae53ec2f-9257-4f7e-9525-d7b2aa0ad809}" trackID: undefined
<prototype>: Object { … }
```

Unfortunately I'm not able to share code to reproduce this error. :/

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -486,7 +486,7 @@ export default abstract class LocalTrack<
         throw TypeError('cannot set processor on track of unknown kind');
       }
 
-      attachToElement(this._mediaStreamTrack, processorElement);
+      await attachToElement(this._mediaStreamTrack, processorElement);
       processorElement.muted = true;
 
       processorElement
--- a/src/room/track/Track.ts
+++ b/src/room/track/Track.ts
@@ -338,7 +338,7 @@ export abstract class Track<
   }
 }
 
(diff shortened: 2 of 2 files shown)


lukasIO:
what problem does that error cause?
Normally this is caused by another play request and shouldn't impact the functionality. 

A reproduction of the issue would be important in order to follow up on this.

pabra:
We have not noticed any impact in functionality. So far we just noticed the error in the logs and I tracked it down to this point.

lukasIO:
got it. I think the issue here is simply that we're logging an error on `play().catch` eventhough it's not strictly true that it results in erroneous behaviour. The right fix might just be to log a warning instead of an error for this case or to log an error only for specific error types. 
Will close this PR.

PR #1198, opened 2024-07-25
Handle SignalReconnecting event in ReconnectCheck

svajunas-budrys:


lukasIO:
that makes sense, thank you for the contribution!

lukasIO (approved):

PR #1200, opened 2024-07-26
room: hook AttributesChanged

nbsp:


davidzhao (approved):
nice catch!

nbsp:
oops, forgot the changeset.

lukasIO:
thanks for catching this! 
I implemented the components' version at the same time and used that to confirm it's working, but that only listens to the ParticipantEvent's directly 🤦 

We can add the changeset in a separate PR, just mention this PR with something like `Emit RoomEvent.ParticipantAtrributes changed #1200`

PR #1207, opened 2024-07-30
Fix issues in PublishVideoCheck and PublishAudioCheck helpers

svajunas-budrys:
- It seems like `mediaType` is not working on Safari, so it was changed to `kind` with a fallback to `mediaType`.
- Timeout for video publish was increased `3000 -> 5000`
- All video outbound-rtp packetsSent are now added to the total value, only the last value was used before.

image

https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpStreamStats/kind
https://developer.mozilla.org/en-US/docs/Web/API/RTCOutboundRtpStreamStats#browser_compatibility

### DIFF
--- a/src/connectionHelper/checks/publishAudio.ts
+++ b/src/connectionHelper/checks/publishAudio.ts
@@ -21,7 +21,10 @@ export class PublishAudioCheck extends Checker {
     }
     let numPackets = 0;
     stats.forEach((stat) => {
-      if (stat.type === 'outbound-rtp' && stat.kind === 'audio') {
+      if (
+        stat.type === 'outbound-rtp' &&
+        (stat.kind === 'audio' || (!stat.kind && stat.mediaType === 'audio'))
+      ) {
         numPackets = stat.packetsSent;
       }
     });
--- a/src/connectionHelper/checks/publishAudio.ts
+++ b/src/connectionHelper/checks/publishAudio.ts
@@ -21,7 +21,7 @@ export class PublishAudioCheck extends Checker {
     }
     let numPackets = 0;
     stats.forEach((stat) => {
-      if (stat.type === 'outbound-rtp' && stat.mediaType === 'audio') {
+      if (stat.type === 'outbound-rtp' && stat.kind === 'audio') {
         numPackets = stat.packetsSent;
       }
     });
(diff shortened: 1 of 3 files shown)


src/connectionHelper/checks/publishAudio.ts
```diff
@@ -21,7 +21,7 @@ export class PublishAudioCheck extends Checker {
     }
     let numPackets = 0;
     stats.forEach((stat) => {
-      if (stat.type === 'outbound-rtp' && stat.mediaType === 'audio') {
+      if (stat.type === 'outbound-rtp' && stat.kind === 'audio') {
```
lukasIO, line 24: it would probably be best to do a feature check and fall back to `mediaType` if `kind` is not present

src/connectionHelper/checks/publishVideo.ts
```diff
@@ -21,8 +21,11 @@ export class PublishVideoCheck extends Checker {
     }
     let numPackets = 0;
     stats.forEach((stat) => {
-      if (stat.type === 'outbound-rtp' && stat.mediaType === 'video') {
-        numPackets = stat.packetsSent;
+      if (
+        stat.type === 'outbound-rtp' &&
+        (stat.kind === 'video' || (!stat.kind && stat.mediaType === 'video'))
+      ) {
+        numPackets += stat.packetsSent;
```
svajunas-budrys, line 28: I changed this so that every `stat.packetsSent` is added to the total `numPackets` value, not just the last one. While testing on safari I noticed that it has 3 `outbound-rtp` stats and only the last one was added.

src/connectionHelper/checks/publishVideo.ts
```diff
@@ -12,7 +12,7 @@ export class PublishVideoCheck extends Checker {
     const track = await createLocalVideoTrack();
     room.localParticipant.publishTrack(track);
     // wait for a few seconds to publish
-    await new Promise((resolve) => setTimeout(resolve, 3000));
+    await new Promise((resolve) => setTimeout(resolve, 5000));
```
svajunas-budrys, line 15: I increased this because it seems that Safari needs more time. Two out of three `packetsSent` values were 0, and when the last one was 0, it was failing because of how numPackets was set.

lukasIO: thanks for that!

lukasIO (approved):
thanks for the improvements!

PR #1209, opened 2024-07-31
Add RemoteTrack.setPlayoutDelay

### LINKED ISSUE #1204: Ability to set playoutDelayHint on RemoteTrack
### Describe the problem

It's sometimes desirable to introduce a bit more jitter buffer on the client-side to allow smoother playback, or to delay the audio in case the corresponding video track is going through things like HDMI. 

`RemoteTrack.setPlayoutDelay(ms)` and `RemoteTrack.getPlayoutDelay()` would give users visibility and control in the requested playback delay

### Alternatives considered

Currently this is adjustable manually by manipulating playoutDelayHint, but it seems a bit opaque and cumbersome to use.

### Importance

would make my life easier

### Additional Information

_No response_

lukasIO:
works only in Chrome today. 

closes #1204

### DIFF
--- a/src/room/track/RemoteTrack.ts
+++ b/src/room/track/RemoteTrack.ts
@@ -39,6 +39,9 @@ export default abstract class RemoteTrack<
     const onRemoveTrack = (event: MediaStreamTrackEvent) => {
       if (event.track === this._mediaStreamTrack) {
         stream.removeEventListener('removetrack', onRemoveTrack);
+        if (this.receiver && 'playoutDelayHint' in this.receiver) {
+          this.receiver.playoutDelayHint = undefined;
+        }
         this.receiver = undefined;
         this._currentBitrate = 0;
         this.emit(TrackEvent.Ended, this);
--- a/src/room/track/RemoteTrack.ts
+++ b/src/room/track/RemoteTrack.ts
@@ -8,13 +8,13 @@ export default abstract class RemoteTrack<
   TrackKind extends Track.Kind = Track.Kind,
 > extends Track<TrackKind> {
   /** @internal */
-  receiver?: RTCRtpReceiver;
+  receiver: RTCRtpReceiver | undefined;
 
   constructor(
     mediaTrack: MediaStreamTrack,
     sid: string,
     kind: TrackKind,
-    receiver?: RTCRtpReceiver,
+    receiver: RTCRtpReceiver,
     loggerOptions?: LoggerOptions,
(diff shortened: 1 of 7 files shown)


src/room/track/RemoteTrack.ts
```diff
@@ -73,6 +73,39 @@ export default abstract class RemoteTrack<
     return statsReport;
   }
 
+  /**
+   * Allows to set a playout delay (in seconds) for this track.
+   * A higher value allows for more buffering of the track in the browser
+   * and will result in a delay of media being played back of `delayInSeconds`
+   */
+  setPlayoutDelay(delayInSeconds: number): void {
+    if (this.receiver) {
```
davidzhao, line 82: if it's not an optional type anymore, is this check still needed?

src/room/track/RemoteTrack.ts
```diff
@@ -73,6 +73,39 @@ export default abstract class RemoteTrack<
...
+      if ('playoutDelayHint' in this.receiver) {
+        this.receiver.playoutDelayHint = delayInSeconds;
+      } else {
+        this.log.warn('Playout delay not supported in this browser');
+      }
+    } else {
+      this.log.warn('Cannot set playout delay, track already ended');
```
davidzhao, line 89: does this mean we'd clear this out after the track has ended?

src/room/track/RemoteTrack.ts
lukasIO: It's only not optional in the constructor.
We set `this.receiver = undefined` when the track is stopped.

src/room/track/RemoteTrack.ts
lukasIO: if the track ended, we don't have the `receiver` anymore. 
I was under the impression we're only re-using the transceiver, not the receiver itself, so we don't need to clear out anything on track ended.

davidzhao: makes sense. FYI the receiver will also be re-used. The transceiver is just a bundle of sender and receiver. So it will be the same receiver (as well as MediaStreamTrack) when re-use is active.

davidzhao (approved):
lg!

src/room/track/RemoteTrack.ts
lukasIO: oh, TIL, then I guess we should actually reset it. Thanks for the pointer!

PR #1223, opened 2024-08-14
Add firstReceivedTime and lastReceivedTime to received TranscriptionSegments

lukasIO:


src/room/Room.ts
```diff
@@ -1274,6 +1280,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     this.clearConnectionReconcile();
     this.isResuming = false;
     this.bufferedEvents = [];
+    this.transcriptionReceivedTimes.clear();
```
bcherry, line 1283: oh this is something i didn't do in swift. can a Room object be reused? (e.g. disconnect then reconnect without creating a new instance)

bcherry (approved):

src/room/Room.ts
lukasIO: yeah, room instances can be reused!

PR #1224, opened 2024-08-14
Add dedicated chat API

lukasIO:


src/room/events.ts
```diff
@@ -519,6 +520,9 @@ export enum ParticipantEvent {
    * fired on local participant only, when the first remote participant has subscribed to the track specified in the payload
    */
   LocalTrackSubscribed = 'localTrackSubscribed',
+
+  /** only emitted on local participant */
+  ChatMessage = 'chatMessage',
```
bcherry, line 525: maybe should be called ChatMessageReceived? Might also clear up some confusion to split it into ChatMessageReceived and ChatMessageSent

src/room/events.ts
```diff
@@ -519,6 +520,9 @@ export enum ParticipantEvent {
    * fired on local participant only, when the first remote participant has subscribed to the track specified in the payload
    */
   LocalTrackSubscribed = 'localTrackSubscribed',
+
+  /** only emitted on local participant */
```
bcherry, line 524: is this comment right? it looks like it's emitted on room, with participant in the args

src/room/types.ts
```diff
@@ -68,3 +68,11 @@ export interface TranscriptionSegment {
   firstReceivedTime: number;
   lastReceivedTime: number;
 }
+
+export interface ChatMessage {
+  id: string;
+  timestamp: number;
+  message: string;
+  editTimestamp?: number;
+  deleted?: boolean;
```
bcherry, line 77: is this deleted property used? I don't see code for sending or processing it

bcherry (approved):
looks good, some non-blocker questions

src/room/events.ts
lukasIO: Was trying to avoid having two different events for `sent` and `received` because it generally generates more boilerplate when having to listen to two different events. The second argument that lead me to try to combine the events under one name was that when differentiating between `sent` and `received` we'd also have to consider a different event for `updated`

src/room/types.ts
lukasIO: currently it's not, can remove it until we actually implement the logic. good call

src/room/events.ts
lukasIO: It's defined both on RoomEvents and ParticipantEvents. This comment is for the `ParticipantEvent.ChatMessage`, so it's correct here.

src/room/events.ts
bcherry: I see - I was confused.

Maybe the ParticipantEvent one should be called `localChatMessageSent`, since that's all its used for and it would then align perfectly with its handler, `onLocalChatMessageSent`. 

The RoomEvent one stay as `chatMessage` but you should add a comment explaining that its emitted for chat messages sent, received, and updated.

lukasIO: that sounds like a good idea!

PR #1228, opened 2024-08-23
Fast track publication

cnderrauber:
Add fastPublish option to establish publisher peerconnection when
joining.

Neogtiate new track without AddTrackRequest response

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -93,6 +93,12 @@ export interface InternalRoomOptions {
   e2ee?: E2EEOptions;
 
   loggerName?: string;
+
+  /**
+   * shorten the publish latency by establish publisher peerconnection when joining room, will
+   * cost few cpu time & network traffic even if there is no track has been published.
+   */
+  fastPublish: boolean;
 }
 
 /**
--- a/src/options.ts
+++ b/src/options.ts
@@ -93,12 +93,6 @@ export interface InternalRoomOptions {
   e2ee?: E2EEOptions;
 
   loggerName?: string;
-
-  /**
-   * shorten the publish latency by establish publisher peerconnection when joining room, will
-   * cost few cpu time & network traffic even if there is no track has been published.
-   */
-  fastPublish?: boolean;
 }
 
 /**
--- a/src/options.ts
+++ b/src/options.ts
@@ -98,7 +98,7 @@ export interface InternalRoomOptions {
    * shorten the publish latency by establish publisher peerconnection when joining room, will
    * cost few cpu time & network traffic even if there is no track has been published.
    */
-  fastPublish: boolean;
+  fastPublish?: boolean;
 }
 
 /**
--- a/src/room/PCTransport.ts
+++ b/src/room/PCTransport.ts
@@ -23,6 +23,8 @@ eliminate this issue.
 */
 const startBitrateForSVC = 0.7;
 
+const debounceInterval = 100;
+
 export const PCEvents = {
   NegotiationStarted: 'negotiationStarted',
   NegotiationComplete: 'negotiationComplete',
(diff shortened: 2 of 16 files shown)


src/room/PCTransport.ts
```diff
@@ -217,18 +221,32 @@ export default class PCTransport extends EventEmitter {
...
-  negotiate = debounce(async (onError?: (e: Error) => void) => {
-    this.emit(PCEvents.NegotiationStarted);
-    try {
-      await this.createAndSendOffer();
-    } catch (e) {
-      if (onError) {
-        onError(e as Error);
-      } else {
-        throw e;
+  async negotiate(onError?: (e: Error) => void) {
```
cnderrauber, line 224: negotiate immediately if time elapsed since last negotiation > debounce interval (100ms)

src/options.ts
```diff
@@ -93,6 +93,12 @@ export interface InternalRoomOptions {
   e2ee?: E2EEOptions;
 
   loggerName?: string;
+
+  /**
+   * shorten the publish latency by establish publisher peerconnection when joining room, will
+   * cost few cpu time & network traffic even if there is no track has been published.
+   */
+  fastPublish: boolean;
```
lukasIO, line 101: this should not be a required option to set
```suggestion
  fastPublish?: boolean;
```

src/room/PCTransport.ts
```diff
@@ -217,18 +221,32 @@ export default class PCTransport extends EventEmitter {
...
+    const debounceInterval = 100;
+    const doNeogiate = async () => {
```
lukasIO, line 226:
```suggestion
    const doNegotiate = async () => {
```

src/room/PCTransport.ts
```diff
@@ -217,18 +221,32 @@ export default class PCTransport extends EventEmitter {
...
+        this.negotiationPending = false;
       }
+    };
+
+    const pending = this.negotiationPending;
+    this.negotiationPending = true;
+    if (!pending && Date.now() - this.lastNegotiationAt > debounceInterval) {
+      await doNeogiate();
+    } else {
+      debounce(doNeogiate, debounceInterval);
```
lukasIO, line 247: I don't think this will have the intended effect as it creates a new debounced instance each time instead of reusing an existing one?

src/room/PCTransport.ts
lukasIO: this optimization will then only work for a single track. If someone were to publish cam + mic at the same time, one of these would get delayed. 
Wondering if we can take a different approach that would allow multiple tracks to be negotiated together without having to necessarily wait for the 100ms debounce timeout

cnderrauber: currently, the `publishTrack` method doesn't know if there is another track publishing together, I think we can provide a `publishTracks` method for this case and make `enableMicAndCamera` and `setTrackEnabled(sceenshare)` to call the new method.

src/room/PCTransport.ts
cnderrauber: oh, i misunderstand it

src/room/PCTransport.ts
lukasIO: sorry, I should have checked for this earlier, but turns out the `debounce` lib we're using has this functionality built in already: https://github.com/chodorowicz/ts-debounce?tab=readme-ov-file#function-arguments

we can just pass `{isImmediate: true}` as an option. Then we don't have to do the conditional execution handling and can get rid of the `doNegotiate` declaration again

lukasIO: pushed a commit to make use of the built-in version

lukasIO: > provide a publishTracks method for this case and make enableMicAndCamera and setTrackEnabled(sceenshare)

I think in practice the problem here will be that a lot of the client time is spent on actually acquiring the devices, so we're missing out on a lot of potential speed improvement. 
I was thinking whether we could optimistically publish empty tracks for these APIs immediately when called and then `replaceTrack` once the getUserMedia promise has resolved. Not sure how to handle initial trackinfo in that regard though.

cnderrauber: The `setLocalDescription` failed after removing the debounce and publishing two tracks at once, looks like a condition race issue. we can address the issue in a separate PR since it is a client side change, just reduce the debounce interval to 20ms in this PR.

lukasIO (approved):

PR #1237, opened 2024-09-02
Switch active device if previously selected device becomes unavailable

lukasIO:
also update active device on local track restarts


davidzhao (approved):
lg

PR #1241, opened 2024-09-03
Treat MissingKey as decryption failure to prevent spamming the logs

hughns:


src/e2ee/worker/FrameCryptor.ts
```diff
@@ -369,16 +370,19 @@ export class FrameCryptor extends BaseFrameCryptor {
...
-      this.emit(
-        CryptorEvent.Error,
-        new CryptorError(
-          `missing key at index ${keyIndex} for participant ${this.participantIdentity}`,
-          CryptorErrorReason.MissingKey,
-          this.participantIdentity,
-        ),
-      );
+      // emit an error if the key index is out of bounds but the key handler thinks we still have a valid key
+      if (this.keys.hasValidKey) {
```
lukasIO, line 374: hm, but that check is already part of the condition two lines above. Not sure what this would change.

hughns: You are quite right, I've removed the redundant logic.

The key change which is now more visible is calling `this.keys.decryptionFailure()` when a key is missing and not just when it was invalid.

lukasIO (approved):
makes sense, thanks!

PR #1245, opened 2024-09-09
Pass connect options to room from connection checkers

jespermjonsson:
The connection options is included in [Checker documentation](https://docs.livekit.io/client-sdk-js/interfaces/CheckerOptions.html)  so one might think that creating a Checker with some connection options like "rtcConfig" that it would be respected when checks are run. Some checkers seems to use parts of this information, but checkers using normal room connect functionality will not. This change should fix so all connection checkers using normal room connect flow will use this information.

### DIFF
--- a/.changeset/chilled-insects-collect.md
+++ b/.changeset/chilled-insects-collect.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
nice, thanks for spotting this!

PR #1257, opened 2024-09-26
Fix changed attribute computation

lukasIO:


davidzhao (approved):
good catch... I guess this will change the behavior of clients expecting deltas to be the other way?

lukasIO:
true, arguably there's a argument to be made that the diff _should_ have the old values, but I'm not sure.

bcherry (approved):
Can you update the docstring to explain exactly what is included in the emitted event.

PR #1260, opened 2024-09-30
Handle e2ee worker messages sequentially

lukasIO:
with sped up publication we've seen instances where the key on the worker had not been set yet while the cryptor was already trying to encode first frames. 
This PR tries to mitigate these instances threefold 
- process incoming worker messages sequentially in an async queue
- post keys before posting enable messages
- instead of throwing an error within the encode function, emit the cryptorError event and return early without enqueuing a frame


davidzhao (approved):
lgtm

PR #1268, opened 2024-10-02
Fix to allow key ring size of 256 and not 255

hughns:


lukasIO (approved):
this is great, thank you!

PR #1274, opened 2024-10-06
chore: remove replace rollup plugin

ryanleecode:
The repo doesn't even use protobuf.js anymore and instead relies on @bufbuild/protobuf-es, so this can be safely removed. Before and after the esm bundle size was 788.4KB, which means the plugin was doing nothing.


lukasIO:
makes sense, thanks!

PR #1275, opened 2024-10-06
chore: update tsconfig

ryanleecode:
update tsconfig to use the latest and greatest of typescript


lukasIO:
for backwards compatibility this is not something that we'll want to do right now.

Issue #1273, opened 2024-10-06
bundle size is too big


ryanleecode:
### Describe the bug

700kb uncompressed is too big

### Reproduction

 .

### System Info

```shell
.
```

### Severity

annoyance

lukasIO:
hey, 
too big for what? 
What usecase do you have that consumes the bundle uncompressed?

lukasIO:
https://bundlephobia.com/package/livekit-client@2.5.9 bundlephobia reports uncompressed as ~350kb. Closing this

PR #1276, opened 2024-10-06
chore: output esm as modules

ryanleecode:
We can output ESM as modules instead of one big file. Should help with https://github.com/livekit/client-sdk-js/issues/1273


lukasIO:
appreciate the PR, have you done any benchmarks comparing the file sizes of tree shaking imports with/without single file output?

lukasIO:
closing due to inactivity

PR #1282, opened 2024-10-09
RPC implementation

bcherry:
This is a full implementation of the [native RPC spec ](https://www.notion.so/livekit/RPC-Proposal-435567131ce64077aa4466fafc2fb407?pvs=4).

It includes test, documentation, and a complete in-browser working example.

The example has been tested with a local LiveKit instance and in LiveKit cloud.


src/room/participant/LocalParticipant.ts
```diff
@@ -1412,6 +1437,129 @@ export default class LocalParticipant extends Participant {
...
+   * @throws Error on failure. Details in `message`.
+   */
+  async performRpc(
+    destinationIdentity: string,
+    method: string,
+    payload: string,
+    responseTimeoutMs: number = 10000,
+  ): Promise<string> {
+    const maxRoundTripLatencyMs = 2000;
+
```
lukasIO, line 1456: throw error early if livekit server doesn't support RPCs

examples/rpc/api.js
lukasIO, line 1: in case it's required for this to be a `js` file, it would be nice to at least add `// @ts-check` to the top of the file

src/room/participant/LocalParticipant.ts
```diff
@@ -1440,6 +1588,156 @@ export default class LocalParticipant extends Participant {
...
+          error,
+        );
+        responseError = RpcError.builtIn('APPLICATION_ERROR');
+      }
+    }
+    await this.publishRpcResponse(caller.identity, requestId, responsePayload, responseError);
+  }
+
+  /** @internal */
+  async publishRpcRequest(
```
lukasIO, line 1661: would be nice if we could keep those either private or move them into `rtcEngine`. 
having them as (albeit internal) methods on the localParticipant might be a bit confusing

src/room/participant/LocalParticipant.ts
bcherry: How do I detect that?

examples/rpc/api.js
bcherry: no i was just running into some baffling type issues so I punted. I went in and finally found and resolved the issue so its back to `.ts`

src/room/participant/LocalParticipant.ts
bcherry: yep this was just an oversight. i marked them private. In the process I also realized a more complete way to test the implementation would be to mock the engine's sendDataPacket method instead of these methods, so I edited the test and it no longer has to mock these private/internal methods

src/room/participant/LocalParticipant.ts
lukasIO: `room.serverInfo` holds the version of the server (and whether it's cloud or not)

bcherry: that makes sense but I see now that we've deployed cloud but we have yet to release the livekit package itself so we don't have a new version number yet. We haven't done a release there in two months - perhaps it's about time to do so... what do you think @boks1971 ?

boks1971: Yeah, we can make a release. @davidzhao does it normally. I should learn to do it and unload it from DZ.

bcherry: @lukasIO I noticed that it's actually kinda tricky to access from here. The simplest way seems to be `this.engine.latestJoinResponse?.serverInfo?.version`. It makes me wonder I should move these methods?

In any case I don't want to block merging here and I'm not sure when exactly cloud will get its version number updated so I might do this as a follow-up.

lukasIO: yeah makes sense. 

jfyi, the method is on room: `room.serverInfo`, but `localParticipant` doesn't hold a direct reference to `room`.

src/room/participant/LocalParticipant.ts
```diff
@@ -1415,6 +1439,129 @@ export default class LocalParticipant extends Participant {
...
+      const responseTimeoutId = setTimeout(() => {
+        this.pendingResponses.delete(id);
+        reject(RpcError.builtIn('RESPONSE_TIMEOUT'));
+      }, responseTimeoutMs);
+
+      this.pendingResponses.set(id, {
+        resolve: (responsePayload: string | null, responseError: RpcError | null) => {
+          clearTimeout(responseTimeoutId);
+          if (responseError) {
+            reject(responseError);
```
lukasIO, line 1497: we might still want to clean up the pendingAck/ackTimeout here?

src/room/participant/LocalParticipant.ts
bcherry: yeah exactly which is why it's kind of a PITA to check it here. I think it would be good to make it easier to get serverInfo from localParticipant but it's not critical - for internal use i think the engine check is fine

src/room/participant/LocalParticipant.ts
bcherry: good call, fixed

src/room/participant/LocalParticipant.ts
boks1971: yeah, cloud has been deployed (maybe twice) after these changes.

src/room/Room.ts
```diff
@@ -1547,6 +1552,12 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       this.handleSipDtmf(participant, packet.value.value);
     } else if (packet.value.case === 'chatMessage') {
       this.handleChatMessage(participant, packet.value.value);
+    } else if (packet.value.case === 'rpcRequest') {
+      this.handleRpcRequest(packet.participantIdentity, packet.value.value);
```
lukasIO, line 1556: we could register these on the local participant directly (the engine is already available and we're registering some events in there already) and as a result make all those methods private in localParticipant, reducing public API surface

bcherry: ah good idea - done

lukasIO (approved):
🤖 ☁️ 📲  🚀

PR #1283, opened 2024-10-10
Additional tests for E2EE

hughns:
This actually does some encryption and decryption with test vectors.

### DIFF
--- a/src/e2ee/worker/FrameCryptor.test.ts
+++ b/src/e2ee/worker/FrameCryptor.test.ts
@@ -60,6 +60,21 @@ class TestUnderlyingSink<T> implements UnderlyingSink<T> {
 function prepareParticipantTestDecoder(
   participantIdentity: string,
   partialKeyProviderOptions: Partial<KeyProviderOptions>,
+) {
+  return prepareParticipantTest('decode', participantIdentity, partialKeyProviderOptions);
+}
+
+function prepareParticipantTestEncoder(
+  participantIdentity: string,
+  partialKeyProviderOptions: Partial<KeyProviderOptions>,
+) {
+  return prepareParticipantTest('encode', participantIdentity, partialKeyProviderOptions);
+}
+
+function prepareParticipantTest(
+  mode: 'encode' | 'decode',
+  participantIdentity: string,
+  partialKeyProviderOptions: Partial<KeyProviderOptions>,
 ): {
   keys: ParticipantKeyHandler;
   cryptor: FrameCryptor;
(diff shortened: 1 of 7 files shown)


src/e2ee/worker/FrameCryptor.ts
```diff
@@ -555,7 +555,7 @@ export class FrameCryptor extends BaseFrameCryptor {
 
     this.sendCounts.set(synchronizationSource, sendCount + 1);
 
-    return iv;
+    return new Uint8Array(iv);
```
lukasIO, line 558: any reason why you added this? looks like an unnecessary conversion

lukasIO (commented):
thank you for this effort! these are great

src/e2ee/worker/FrameCryptor.test.ts
```diff
@@ -113,154 +119,279 @@ describe('FrameCryptor', () => {
...
+    });
 
-    input.write(mockEncryptedRTCEncodedVideoFrame(1));
+    it('calls decryptionFailure on missing key and emits error', async () => {
+      vitest.useFakeTimers();
+      try {
+        const { cryptor, keys, input } = prepareParticipantTestDecoder(participantIdentity, {});
+
+        const errorListener = vitest.fn().mockImplementation((e) => {
+          console.log('error', e);
```
lukasIO, line 294: just to keep the output of the test clean, let's remove these log statements

src/e2ee/worker/FrameCryptor.test.ts
```diff
@@ -113,154 +119,279 @@ describe('FrameCryptor', () => {
...
+      try {
+        const { cryptor, keys, input, output } = prepareParticipantTestDecoder(
+          participantIdentity,
+          { ratchetWindowSize: 0 },
+        );
+
+        vitest.spyOn(crypto.subtle, 'decrypt');
+        vitest.spyOn(keys, 'decryptionFailure');
+        const errorListener = vitest.fn().mockImplementation((e) => {
+          console.log('error', e);
```
lukasIO, line 338: same as above

src/e2ee/worker/FrameCryptor.ts
hughns: I meant to say something about this....

So, the Node.js WebCrypto implementation did not like `iv` being an `ArrayBuffer`.

It would fail with:

`Failed to normalize algorithm: 'iv' of 'AesGcmParams' (passed algorithm) is not instance of ArrayBuffer, Buffer, TypedArray, or DataView.`

...even though `iv` is an `ArrayBuffer`. 🤷

This got me thinking (which I then forgot to ask about): Should we be testing across multiple platforms and running these tests inside real browsers? (the goal would be 100% coverage of `.browserslistrc`)

hughns: So, having played around with running vitest with real browsers, it feels like we could do this in CI, but would be a pain for local dev.

Another approach: if I set the vitest `environment` to `happy-dom` or `node` (instead of `jsdom`) then the test passes fine.

Would it be acceptable to set the environment to `node` or `happy-dom`? (I'm not personally familiar with the latter)

lukasIO: we use happy-dom in other repos already successfully, so would be good to switch to that!

lukasIO:
great, thank you!

lukasIO (approved):

PR #1297, opened 2024-10-24
Keep dd extension id in the session

cnderrauber:
Safari could generate different rtp extensions
for video tracks(camera and screenshare), need
to use same extension id for dependency descriptor in the offer.

### DIFF
--- a/src/room/PCTransport.ts
+++ b/src/room/PCTransport.ts
@@ -1,5 +1,5 @@
 import { EventEmitter } from 'events';
-import type { MediaDescription } from 'sdp-transform';
+import type { MediaDescription, SessionDescription } from 'sdp-transform';
 import { parse, write } from 'sdp-transform';
 import { debounce } from 'ts-debounce';
 import log, { LoggerNames, getLogger } from '../logger';
@@ -290,7 +290,7 @@ export default class PCTransport extends EventEmitter {
           }
 
           if (isSVCCodec(trackbr.codec)) {
-            this.ensureVideoDDExtensionForSVC(media);
+            this.ensureVideoDDExtensionForSVC(media, sdpParsed);
           }
 
           // TODO: av1 slow starting issue already fixed in chrome 124, clean this after some versions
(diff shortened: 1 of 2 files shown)


src/room/PCTransport.ts
```diff
@@ -503,6 +505,36 @@ export default class PCTransport extends EventEmitter {
...
+
+    if (!ddFound) {
+      if (this.ddExtID === 0) {
+        this.ddExtID = maxID + 1;
+      }
+      media.ext?.push({
+        value: this.ddExtID,
+        uri: ddExtensionURI,
+      });
+    }
```
boks1971, line 536: What happens if the first one has fewer extensions, let us say 2 and the maxID could be 2 and ddExt will get three. And the next track has three extensions and three is already taken.

boks1971: If this is all one-byte extensions (nowadays libwebrtc has so many extensions, I am guessing it may not fit in one-byte some time soon), can just keep an array of 15 slots and keep track of which slots are used and find an empty slot of first track to use for DD. And use that to munge all extensions for all the published tracks to ensure that they use the same ID?

cnderrauber: I don't think this will happen if the first munged offer already contains a extension with id=3 and the second offer reuse 3 for another extension. If this is true, we will be nerver safe to append a extension to the sdp.

boks1971: Got it. So, after munging, `libwebrtc` would updates its list and not use the munged ID? What happens if two tracks are in the same `offer` with the first one having only two extensions and the second track has three extensions. Should this code parse all tracks first to get the maximum extension and then set it?

cnderrauber: I think libwebrtc always generates same extensions for tracks but safari don't. It sounds unlikely to publish two video tracks at once and includes different extensions, but parse entire sdp is good to defense.

boks1971 (approved):
lgtm!

PR #1299, opened 2024-10-24
Copy README images when generating docs

bcherry:
This should fix the broken image on top of https://docs.livekit.io/client-sdk-js/


lukasIO (approved):
nice

PR #1300, opened 2024-10-24
Move github banners to assets directory

bcherry:
Unfortunately #1299 didn't work both due to a leading slash in the img tag (which means relative to the host not the current path) and due to the `.github/` folder not being exposed from S3/cloudfront. Instead of making changes in readme manager, I changed the doc gen script to also rewrite the image url and move the images into assets where they belong.


kath0la (approved):
LGTM!!

PR #1301, opened 2024-10-25
Add DataStream support

lukasIO:


src/room/participant/LocalParticipant.ts
```diff
@@ -1494,6 +1511,297 @@ export default class LocalParticipant extends Participant {
...
+    localP.engine.once(EngineEvent.Closing, onEngineClose);
+
+    const writer = new TextStreamWriter(writableStream, info, () =>
+      this.engine.off(EngineEvent.Closing, onEngineClose),
+    );
+
+    return writer;
+  }
+
+  async sendFile(
```
davidzhao, line 1711: will there similarly be a `streamBytes` interface for sending binary streams from memory?

davidzhao (approved):
awesome work!

src/room/participant/LocalParticipant.ts
lukasIO: Do you have use cases for this in mind already from the browser?

Was thinking we might ship that when we have concrete use cases (which I didn't have)

davidzhao: I think it's more for completeness. fine to add later

PR #1302, opened 2024-10-25
mimeTypeToVideoCodecString should not throw

davidzhao:
The current implementation is not resilient to codecs changes. If a new codec is introduced, previously working code would start failing.

### DIFF
--- a/src/room/track/utils.ts
+++ b/src/room/track/utils.ts
@@ -9,7 +9,6 @@ import {
   type ScreenShareCaptureOptions,
   type VideoCaptureOptions,
   type VideoCodec,
-  videoCodecs,
 } from './options';
 import type { AudioTrack } from './types';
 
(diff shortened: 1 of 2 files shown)


cnderrauber (approved):

src/room/track/utils.ts
```diff
@@ -188,11 +187,7 @@ export function screenCaptureToDisplayMediaStreamOptions(
 }
 
 export function mimeTypeToVideoCodecString(mimeType: string) {
-  const codec = mimeType.split('/')[1].toLowerCase() as VideoCodec;
-  if (!videoCodecs.includes(codec)) {
-    throw Error(`Video codec not supported: ${codec}`);
-  }
-  return codec;
+  return mimeType.split('/')[1].toLowerCase() as VideoCodec;
```
lukasIO, line 190: I'm not sure whether forcing this cast is the right fix. 
This incorrectly passes strings that are not part of the client defined `VideoCodec` type. 

If a method is needed that just strips away `video/` from the mimetype without throwing, I think that should return a string instead

davidzhao: that's a fair point. my main goal is to ensure it doesn't throwing just because there's a piece of data that it doesn't understand.

perhaps the more systematic question is how do we handle new video codecs that the server/browser might contain that our client isn't aware of. we'd have to be defensive about it

PR #1309, opened 2024-11-06
Make Remote Tracks `getSenderStats` method public

mpnri:
Hi.
Currently, `getSenderStats` method in remote tracks (audio & video) are private or protected and unlike local tracks, there is no way to call them from outside the class and this makes us unable to access them if we need to check Stats for remote tracks. (for debugging and metric stuff)


Mahdi-zarei:
We also need this to improve out client side monitoring.

davidzhao (approved):

PR #1325, opened 2024-11-21
Add document pip support

### LINKED ISSUE #1317: Remote participant video elements in Document PiP are wrongly detected as hidden
### Describe the bug

We use the new Chrome Document Picture in Picture mode.
As soon as the remote participants video elements are displayed in the Document PiP window for a couple of seconds their video playback stops and I see a log message in the browser console that it was set to `visible: false` likely the visibility detection isn’t fully working in PiP windows.

### Reproduction

*Repro*
Start a LiveKit call adding the remote participants video elements to the dom.
Trigger [Document PiP](https://developer.mozilla.org/en-US/docs/Web/API/Document_Picture-in-Picture_API) and move the video elements into the PiP window.

*Expected*
Playback of the remote videos continues

*Actual*
Playbeck of the remote videos is paused after a couple of seconds. (The local participant's video playback continues)

### Logs

_No response_

### System Info

```shell
Google Chrome Version 130.0.6723.58 (Official Build) (arm64)
OSX
```

### Severity

blocking an upgrade

### Additional Information

I can propose a fix for this

davideberlein:
- Add support for detecting video element visibility in Document PiP
- Drop opacity check as the intersection observer is not triggered if an element changes opacity and the check is flawed anyway as opacity of parents isn't considered.
- Add toggle document pip button to `examples/demo` page.

Closes #1317

### DIFF
--- a/.changeset/strange-pans-search.md
+++ b/.changeset/strange-pans-search.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': minor
+---
+
+Add support for detecting video element visibility in Document PiP (can be tested on the examples/demo)
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -285,6 +285,7 @@ const appActions = {
     }
 
     const pipWindow = await window.documentPictureInPicture?.requestWindow();
+    setButtonState('toggle-pip-button', 'Close PiP', true);
     // Copy style sheets over from the initial document
     // so that the views look the same.
     [...document.styleSheets].forEach((styleSheet) => {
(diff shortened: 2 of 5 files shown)


lukasIO (commented):
This looks great, thank you @davideberlein ! 

For the demo page two small nits: 
It would be nice if the pip toggle could keep track of it's current state (enabled vs disabled) and there are still some exceptions spewing in the console after popping out the window where some updates are still trying to render to the window.document directly instead of to the pip document.

davideberlein:
> ions spewing in the console after popping out the window where some updates are still trying to render to the window.document directly instead of to the pip document

Good finding there, I missed 2 stops which I added and cleaned up the rest.

lukasIO (approved):
looks good, excited to support this new possibility :)

PR #1328, opened 2024-11-26
chore: expose reconnection types

wuhkuh:
I wanted to implement my own `ReconnectPolicy`, but was unable to use the interface, due to the fact that these were not exposed.


lukasIO (approved):
thanks for the PR! that makes sense

PR #1329, opened 2024-11-28
Fix processor passing in CreateLocalTracks options

lukasIO:
closes https://github.com/livekit/track-processors-js/issues/50


lukasIO:
need to investigate why the tests are failing with this.

nbsp (approved):

PR #1336, opened 2024-12-03
Fix blocking main thread on parallel publishing requests

lukasIO:


chrisackermann (approved):
lgtm - thanks for the quick fix!

PR #1338, opened 2024-12-03
Deep clone the options so we do not set the processor to undefined

toger5:


lukasIO:
thanks, this doesn't work as it would be trying to deep clone a class instance, fixed the issue in #1339

PR #1342, opened 2024-12-05
Fix sdp connection address mismatch issue

cnderrauber:
Chrome could generate sdp with c=IN IP4 <ipv6 addr> in edge case and return error when set sdp. This is not a sdk error but correct it if the issue detected.


examples/rpc/README.md
```diff
@@ -6,8 +6,8 @@ A working multi-participant live demo of the LiveKit RPC feature.
 
 1. Create `.env.local` with `LIVEKIT_API_KEY`, `LIVEKIT_API_SECRET`, and `LIVEKIT_URL`
 1. Install dependencies: `pnpm install`
-2. Start server: `pnpm dev`
-3. Open browser to local URL (typically http://localhost:5173)
-4. Press the button to watch the demo run
+1. Start server: `pnpm dev`
+1. Open browser to local URL (typically http://localhost:5173)
+1. Press the button to watch the demo run
```
boks1971, line 11: Looks like some formatter getting confused?

src/room/PCTransport.ts
```diff
@@ -630,3 +632,15 @@ function extractStereoAndNackAudioFromOffer(offer: RTCSessionDescriptionInit): {
...
+    // Chrome could generate sdp with c = IN IP4 <ipv6 addr>
+    // in edge case and return error when set sdp.This is not a
+    // sdk error but correct it if the issue detected.
+  if (media.connection && media.connection.version === 4 && media.connection.ip.indexOf(':')) {
+    log.debug(
+      `media connection address ${media.connection.ip} mismatched with version ${media.connection.version}, replace it with 0.0.0.0`,
+    );
+    media.connection.ip = '0.0.0.0';
+  }
+}
```
boks1971, line 646: Is there a need to check for the other way? It says IP6 and has a v4 address?

boks1971 (approved):
lgtm

src/room/PCTransport.ts
cnderrauber: I don't know if it is a real case.. maybe worth to add.

PR #1344, opened 2024-12-05
fix: set attributes of remote participants earlier

holzgeist:
Otherwise the attributes are not set during the "ParticipantConnected" event, which makes it difficult to react to this event accordingly.

I didn't test it, but if I understand the code correctly, there is one potentially problematic side effect: "ParticipantAttributesChanged" is not emitted anymore on connect. Which sounds fine to me, as the attributes didn't change from any previous setting, they just started that way. However, it might break workflows


lukasIO (approved):
thanks for the PR, that's a good catch!

Issue #1346, opened 2024-12-06
VoiceAssistantControlBar change device do not show correct selection.


tcper:
### Describe the bug

![image](https://github.com/user-attachments/assets/4740c512-a390-4f5a-b1d3-e87dbffa3423)
Change device, always show the first selection.

### Reproduction

```
<LiveKitRoom
        connect={token !== undefined}
        token={token}
        serverUrl="wss://rtc.metag-ai.com"
        room={room}
      >
<VoiceAssistantControlBar controls={{ leave: false }} />
</LiveKitRoom>
```

Debug the code from livekit-client/livekit-react

Room.ts
```
try {
        success = (
          await Promise.all(tracks.map((t) => t.videoTrack?.setDeviceId(deviceConstraint)))
        ).every((val) => val === true);
      } catch (e) {
        this.options.videoCaptureDefaults!.deviceId = prevDeviceId;
        throw e;
      }
```

success variable always got `false` value,
make the later event emit not work

```
if (deviceHasChanged && success) {
      this.localParticipant.activeDeviceMap.set(kind, deviceId);
      this.emit(RoomEvent.ActiveDeviceChanged, kind, deviceId);
    }
```

### System Info

```shell
Model Name:	MacBook Air
  Model Identifier:	Mac14,2
  Model Number:	Z15W001ADCH/A
  Chip:	Apple M2
  Total Number of Cores:	8 (4 performance and 4 efficiency)
  Memory:	16 GB
  System Firmware Version:	11881.1.1
  OS Loader Version:	11881.1.1
  Serial Number (system):	RMGJXG0QJN
  Hardware UUID:	96DD2714-C497-5025-95DF-EEEE6A98706F
  Provisioning UDID:	00008112-000E20462631401E
  Activation Lock Status:	Enabled


Chrome
```

### Severity

annoyance

PR #1347, opened 2024-12-06
fix: #1346 bugfix, change device show on selection

tcper:
```
success = (
          await Promise.all(tracks.map((t) => t.audioTrack?.setDeviceId(deviceConstraint)))
        ).every((val) => val === true);
```

```
async setDeviceId(deviceId: ConstrainDOMString): Promise<boolean> {
    if (
      this._constraints.deviceId === deviceId &&
      this._mediaStreamTrack.getSettings().deviceId === unwrapConstraint(deviceId)
    ) {
      return true;
    }
    this._constraints.deviceId = deviceId;
    if (!this.isMuted) {
      await this.restartTrack();
    }
    return (
      this.isMuted || unwrapConstraint(deviceId) === this._mediaStreamTrack.getSettings().deviceId
    );
  }
```

Select device not affect on the device selection, and setDeviceId()'s success is not the real selected result.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -835,7 +835,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
    * @internal for testing
    */
   async simulateScenario(scenario: SimulationScenario, arg?: any) {
-    let postAction = () => {};
+    let postAction = () => { };
     let req: SimulateScenario | undefined;
     switch (scenario) {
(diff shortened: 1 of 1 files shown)


lukasIO:
@tcper what exactly is the issue you're observing? `success` has the purpose of making sure that the device has actually been switched on the track itself.

lukasIO:
this should be fixed by #1357

PR #1350, opened 2024-12-09
Keep backwards compatibility for attributes updates on creation

lukasIO:
follow up for #1344 to keep backwards compatibility


davidzhao (commented):
do you think we need this level of compat? the previous behavior seems incorrect. (we don't trigger similar events for metadata upon connect)

davidzhao (approved):

PR #1357, opened 2024-12-19
Lukas/default device handling

lukasIO:


bcherry (approved):

davidzhao (approved):
nice work tracking down the behavior differences!

PR #1358, opened 2024-12-29
[feature] Allow changing arbitrary options during room lifetime (`disconnectOnPageLeave` implemented)

Philzen:
This is a proposal (or an RFC if you want) implementing a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) on the `Room.options` property. This enables to switch options on or off that otherwise can only be configured once on instantiation. Potential applications could be to have `adaptiveStream` or `dynacast` switches in the example applications, but basically any option can be made configurable at runtime with it.

As a proof-of-concept and first application, this PR implements updating `disconnectOnPageLeave` after the Room has been created, because that was my [XY-Problem](https://xyproblem.info/) that led to this: I wanted to hook up a `beforeunload` handler so a user confirmation is shown should they navigate away during a LiveKit session that is being recorded. The problem i discovered was that the room was being disconnected anyway after the user chose to stay on the page in the confirm dialog.  Digging into the library i realized i had to initialize the Room with `disconnectOnPageLeave: false` – however i felt i don't want to lose this very sensible default behavior. So i implemented this Proxy, which enables to simply `room.options.disconnectOnPageLeave = false` when recording starts and my custom `beforeunload` handler is registered, and vice versa do `room.options.disconnectOnPageLeave = true` when recording stops.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -235,9 +235,11 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       this.setupE2EE();
     }
 
+    /**
+     * Proxy for interception of changes to InternalRoomOptions after Room has been instantiated.
+     */
     this.options = new Proxy(this.options, {
       set: (target: InternalRoomOptions, key: keyof InternalRoomOptions, value) => {
-        console.log("Proxy is intercepting!", key, value)
         if (key == 'disconnectOnPageLeave' && value !== target[key]) {
           value === true ? this.registerUnloadEvents() : this.unregisterUnloadEvents()
         }
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -234,6 +234,19 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     if (this.options.e2ee) {
       this.setupE2EE();
     }
+
+    const self = this
+    this.options = new Proxy(this.options, {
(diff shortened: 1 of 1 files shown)


lukasIO:
thanks for the suggestion!

I think we'd want to expose dedicated (feature based) methods instead of relying on a proxy for these kind of things. 
I don't see an immediate use case (simulcast and dynacast would be rather tricky as that needs updates communicated to the server as well and would - today - have to result in a reconnect). 

WRT your use case of custom page leave behaviour, I think it makes sense to leave `disconnectOnPageLeave: false` and implement the logic for page leave within your app (the defaults events are rather straight forward to hook into)

Philzen:
I can completely follow your point to expose dedicated (feature based) methods instead of relying on a proxy for these kind of things as being a solid design decision. So i'm just leaving this feature branch here which – as said – was just a suggestion.

>  I think it makes sense to leave `disconnectOnPageLeave: false` and implement the logic for page leave within your app (the defaults events are rather straight forward to hook into)

This is of course what i ended up doing, at least for the moment as i didn't want to rely on my custom build of client-sdk-ks ;)
The only thing that i don't like about this is that i have now effectively duplicated perfectly fine event handler logic from the sdk with the sole purpose of being able to enable/disable it after instantiation. As i am sure i won't be the last person that needs something like that, maybe it's worth considering this use case. Of course  can see two options here:

Option A would be to expose dedicated methods as you said.
Option B would be to implement a "prevent accidental navigation" feature directly within the sdk (which may be out of scope).

PR #1362, opened 2025-01-06
Ensure maxFps applies for very low framerates

lukasIO:


src/room/participant/publishUtils.ts
```diff
@@ -352,13 +355,20 @@ function encodingsFromPresets(
...
       rid,
       scaleResolutionDownBy: Math.max(1, size / Math.min(preset.width, preset.height)),
       maxBitrate: preset.encoding.maxBitrate,
     };
-    if (preset.encoding.maxFramerate) {
-      encoding.maxFramerate = preset.encoding.maxFramerate;
+    // ensure that the sourceFramerate is the highest framerate applied across all layers so that the
+    // original encoding doesn't get bumped unintentionally by any of the other layers
+    const maxFramerate =
+      sourceFramerate && preset.encoding.maxFramerate
```
davidliu, line 367: Is it possible for just one of the values to exist?

davidliu (approved):

src/room/participant/publishUtils.ts
lukasIO: Yeah, I think so.
If only one of them exists there should be no need to consider the sourceFramerate

PR #1378, opened 2025-01-22
Replace internal instanceof checks with typeguards

lukasIO:


bcherry (approved):
seems good although why not just use the instance methods like `isLocal` that you added to the types themselves?

lukasIO:
unfortunately that doesn't narrow the types

bcherry:
ah got it that makes sense - still new to typescript but now I understand those method signatures... nice!

PR #1384, opened 2025-01-24
Document error codes in readme/reference docs

chrisackermann:
Need to provide developers with easier reference for error codes in order to help improve their client-side logging and diagnosis of issues


kath0la (approved):
Thanks for adding this!

PR #1385, opened 2025-01-24
Set names of errors and reasonName for enums

lukasIO:
inspired by https://github.com/livekit/client-sdk-js/pull/1384

improve default readability of custom LiveKit errors


davidzhao (approved):
lg!

PR #1395, opened 2025-02-03
Allow audio processing for react native without AudioContext

davidliu:


src/room/track/LocalAudioTrack.ts
```diff
@@ -187,6 +187,8 @@ export default class LocalAudioTrack extends LocalTrack<Track.Kind.Audio> {
       };
       this.log.debug(`setting up audio processor ${processor.name}`, this.logContext);
 
+      // RN won't have AudioContext.
+      // @ts-expect-error
```
lukasIO, line 191: wondering if we can move this expect-error up to the processorOptions definition, into the audioContext line to make sure we're only expecting/ignoring that particular field not being set

davidliu: Updated it to force treat the AudioContext as non-null, I think that should solve it?

lukasIO (approved):

src/room/track/LocalAudioTrack.ts
lukasIO: thanks!

PR #1399, opened 2025-02-07
Add BackupCodecPolicy to TrackPublishDefaults

cnderrauber:
When using advance codecs and backup codec enabled, the client can use REGRESSION or SIMULCAST codec policy to choose subscriber bandwidth efficiency or publisher device/bandwidth efficiency.

### DIFF
--- a/.changeset/honest-hornets-fail.md
+++ b/.changeset/honest-hornets-fail.md
@@ -3,3 +3,7 @@
 ---
 
 Add backupCodecPolicy to TrackPublishDefaults
+
+The default policy of backup codec is `codec regression` for maxium compatibility, which means the publisher stops sending primary codec and all subscribers will receive backup codec even primary codec is supported. 
+It changes the default behavior `multi-codec simulcast` in the previous version, will not break the functionality of the previous version but only cause potential extra bandwidth usage. The user can set the policy to `multi-codec simulcast` to keep the previous behavior.
+ 
--- a/.changeset/honest-hornets-fail.md
+++ b/.changeset/honest-hornets-fail.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Add backupCodecPolicy to TrackPublishDefaults
--- a/.changeset/honest-hornets-fail.md
+++ b/.changeset/honest-hornets-fail.md
@@ -1,5 +1,5 @@
 ---
-'livekit-client': patch
+'livekit-client': minor
 ---
 
 Add backupCodecPolicy to TrackPublishDefaults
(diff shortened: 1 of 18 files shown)


src/room/track/options.ts
```diff
@@ -12,8 +12,7 @@ export interface TrackPublishDefaults {
   videoEncoding?: VideoEncoding;
 
   /**
-   * Multi-codec Simulcast
-   * VP9 and AV1 are not supported by all browser clients. When backupCodec is
+   * Advance codecs (VP9/AV1/H265) are not supported by all browser clients. When backupCodec is
```
lukasIO, line 15:
```suggestion
   * Advanced codecs (VP9/AV1/H265) are not supported by all browser clients. When backupCodec is
```

src/room/track/options.ts
```diff
@@ -24,6 +23,20 @@ export interface TrackPublishDefaults {
    */
   backupCodec?: true | false | { codec: BackupVideoCodec; encoding?: VideoEncoding };
 
+  /**
+   * When backup codec is enabled, there are two options to decide whether to
+   * send the primary codec at the same time:
+   *   * codec regression: publisher stops sending primary codec and all subscribers
+   *       will receive backup codec even primary codec is supported. It is the default
```
lukasIO, line 30:
```suggestion
   *       will receive backup codec even if the primary codec is supported on their browser. It is the default
```

src/room/track/options.ts
```diff
@@ -24,6 +23,20 @@ export interface TrackPublishDefaults {
...
+   *       behavior and provides maximum compatibility and reduced device performance
```
lukasIO, line 31:
```suggestion
   *       behavior and provides maximum compatibility. It also reduces CPU
```

.changeset/honest-hornets-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'livekit-client': patch
```
lukasIO, line 2:
```suggestion
'livekit-client': minor
```

.changeset/honest-hornets-fail.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add backupCodecPolicy to TrackPublishDefaults
```
lukasIO, line 5: I feel like we should call it out in more detail in here that this is changing a default behaviour.

cnderrauber: we can use the comment of codec policy as the detail and describe the behavior change, append them in this changeset file?

lukasIO (approved):

.changeset/honest-hornets-fail.md
lukasIO: sounds good, no there's no format required

PR #1400, opened 2025-02-07
Add updateText API

lukasIO:
to support editing of messages

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -6,6 +6,7 @@ import {
   DataPacket_Kind,
   DataStream_Chunk,
   DataStream_Header,
+  DataStream_OperationType,
   DataStream_Trailer,
   DisconnectReason,
   JoinResponse,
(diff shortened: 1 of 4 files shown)


bcherry (commented):
so this allows reopening a closed a stream to rewrite the text?  is it a requirement that the stream was already closed and maybe that it was actually your stream to begin with?

lukasIO:
> so this allows reopening a closed a stream to rewrite the text?

not really, this is a new stream with a different operationType (`update` instead of `create`) in order to differentiate edits/updates from first-time sends

> is it a requirement that the stream was already closed?

yes, you should only update streams that have previously completed, we don't have a good way to enforce this unfortunately

> that it was actually your stream to begin with

yeah, that's another short coming of not having a centralised message hub that can enforce these kind of restrictions. The SFU overrides the participant identity if it's not coming from an agent to ensure the sender_identity cannot easily be spoofed by other clients. So the app logic for displaying the chat would have to do the check to ensure that the original message is not being overridden if the participant_identities don't match. This requires however that the app has the original message still in memory. 

Also happy with converting this back to a draft if we don't want to support this rn

PR #1402, opened 2025-02-11
added new connection tests:

davidzhao:
- connectionProtocol: tests video publishing over UDP vs TCP
- cloudRegion: tests closest Cloud regions and selects the most performant


lukasIO (approved):
nice!

PR #1356, opened 2024-12-16
Allow version range from `@livekit/protocol` dependency

Philzen:
I realized when migrating to v2 that yarn resolved two separate versions of `@livekit/protocol`.

This should ensure when installing it along with other SDKs (such as `livekit-server-sdk`) to always resolve to a common compatible version without having to dedupe the dependencies.

It's just a suggestion, it may be the case that the livekit team has a very specific reason for pinning the version in this js sdk?

PR #1405, opened 2025-02-11
Allow semver range for @livekit/protocol

Philzen:
Harmonizing it with livekit-server-sdk, allowing for a semver range so package managers can resolve to a single compatible version instead of two (which is currently the case). 

Hence i'd like to re-raise this, to save consumers from having to manually dedupe their dependencies each time.

For further information, see https://github.com/livekit/client-sdk-js/pull/1356#issuecomment-2573005093 and https://github.com/livekit/client-sdk-js/pull/1356#issuecomment-2573011526

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   },
   "dependencies": {
     "@livekit/mutex": "1.1.1",
-    "@livekit/protocol": "1.33.0",
+    "@livekit/protocol": "^1.33.0",
     "events": "^3.3.0",
     "loglevel": "^1.8.0",
     "sdp-transform": "^2.14.1",


lukasIO:
we're relying on exact versions within the SDK implementation, we don't want to support a range for that reason, sorry.

Philzen:
> we're relying on exact versions within the SDK implementation, we don't want to support a range for that reason, sorry.

I hear you – so if the protocol follows semver, using an exact minimum version would suffice, wouldn't it?

This does not seem to be a problem in https://github.com/livekit/node-sdks/blob/a1d26a149340535b132b6360d397f01cd6bd374c/packages/livekit-server-sdk/package.json#L45, so why would it be for livekit-client SDK?

lukasIO:
the protocol repo doesn't strictly follow semver. 

Secondly, and more importantly, the issue you're running into is one of type mismatch. 
this repo _bundles_ all dependencies at build time. The range specified in package.json has no effect on this at all. 
Historically this used to be different for livekit-server-sdk (it only transpiled from ts to js without a bundling step) hence a range could have made more sense if the goal ever was to dedupe dependencies.

Issue #1369, opened 2025-01-13
livekit-client should export `TrackType` for seamless communication with livekit-server-sdk backend


Philzen:
### Describe the bug

Following the discussion at https://github.com/livekit/client-sdk-js/issues/1355#issuecomment-2573140926 i went ahead and tried to replace `TrackType` (imported from the transitive dependency `@livekit/protocol` which following [this comment](https://github.com/livekit/client-sdk-js/issues/1355#issuecomment-2572914598) should be discouraged / not necessary) with `Track.Kind` or `Track.Source` and came to realize that this is not viable for us.

### Explanation

Our backend implements a service – which maps type-safe as a GraphQL-mutation to the frontend – like this:

```ts
export const muteParticipant = async ({ input: { roomName, identity, type } }) => {
  try {
      const participant = await roomServiceClient.getParticipant(urlName, identity)

      // select tracks to mute by `TrackType.AUDIO` or `TrackType.VIDEO` ↓
      const tracks = participant.tracks.filter((track) => track.type === type)
      for (const track of tracks) {
        await roomServiceClient.mutePublishedTrack(urlName, identity, track.sid, true)
      )
    } catch (e) {...}
}
```

Evaluating the possible alternatives to `TrackType` here:

1. AFAIK the only alternative to `track.type` here is `track.source` – but that is not what we want here, as – compared to `TrackType` the `TrackSource` type is less deterministic (i.e. it has `TrackSource.UNKNOWN` which we wouldn't known how to handle here) and also the source of the track is not relevant to this mutation at all.   
  :warning: Also – `Track.Source` in livekit-client does not map to `TrackSource` from livekit-server-sdk/protocol at all … it's actually a pitfall, b/c the former is an enum mapping to string values, while the latter is an enum mapping to integers. 
  We could use `Track.sourceToProto(s: Source)` here, but that is an internal method and also seems more complex than necessary.
2. `TrackKind` doesn't seem exposed at this point at all – but even if it was, it bears the same caveat regarding type mismatch than above – `Track.Kind` from livekit-client is a string-enum, while `TrackKind` from livekit-server-sdk is integer-based. Again, it also has `Track.Kind.Unknown` (`TrackKind.UNKNOWN`) respectively, which we don't exactly know what to make out of it.  
  Again, `Track.kindToProto(k: Kind)` could be used in livekit-client, but i don't believe that's an option that the designers of the SDK would want to encourage.
[...]

PR #1406, opened 2025-02-11
Export TrackType enum as value entity

Philzen:
Just upgraded to v2.8.1 and realized #1370 did not really solve #1369 yet.

As per https://github.com/microsoft/TypeScript/issues/40344, a type import does not suffice to be able to put the enum to any use.


lukasIO (approved):

lukasIO:
thanks!

PR #1410, opened 2025-02-16
Add text auto chunking to text streams

lukasIO:
JS equivalent for https://github.com/livekit/python-sdks/pull/361


davidzhao (approved):
lg!

PR #1414, opened 2025-02-19
Improve utf8 text split and add unit test

lukasIO:
Previously the splitUtf8 method looked at the string length (in characters) which is wrong as we're actually interested in the byte length of the string. 
This PR encodes the string inside of the split function and operates on Uint8Arrays for the splitting


davidzhao (approved):
lg!

PR #1418, opened 2025-02-25
Reject publishing with insufficient permissions present

lukasIO:


boks1971 (approved):
lgtm, but one question
do apps now call publish again so that SDK can do `AddTrack` when it has permissions or is there a need to cache the publish intent and do an `AddTrack` when permissions are granted?

davidzhao (approved):
nice!

paulwe (approved):

lukasIO:
> do apps now call publish again so that SDK can do AddTrack when it has permissions or is there a need to cache the publish intent and do an AddTrack when permissions are granted?

I don't think we should cache publish intent. 
The current behaviour is similar, were publishing will eventually fail, just without providing adequate user feedback.

PR #1431, opened 2025-03-07
Automatically attempt to resume suspended audio contexts on click

lukasIO:
the `createAudioAnalyser` helper ist standalone and creates a new AudioContext. 

If this audio context is constructed prior to any user interaction on the page it might stay suspended. 
This PR registers a click listener on the document body in order to attempt to resume the context on any click on the page


davidzhao (approved):
lg!

PR #1435, opened 2025-03-11
Fix ignored constraints in LocalTrack.restart

### LINKED ISSUE #1434: LocalTrack.restart() method ignores user-provided constraints
### Describe the bug

First, I'd like to thank the LiveKit team for providing an excellent client SDK. I'm reporting an issue that occurs due to code changes in the recent PR #1388. 

- **Expected Behavior**: When calling LocalTrack.restart(constraints), the provided constraints should be accurately applied when restarting the media track.
- **Actual Behavior**: Even when users explicitly provide constraints, the function internally extracts deviceId and otherConstraints from this._constraints rather than from the user-provided constraints. As a result, the deviceId in user-provided constraints is completely ignored, causing camera or microphone switching operations to not work as intended.

#### Issue in the Code
The issue occurs in the restart function in LocalTrack.ts:
```typescript
protected async restart(constraints?: MediaTrackConstraints) {
  // ...
  if (!constraints) {
    constraints = this._constraints;
  }
  const { deviceId, ...otherConstraints } = this._constraints; // Root cause: always using this._constraints
  // ...
}
```
In the code above, when extracting deviceId and otherConstraints, it always uses this._constraints instead of the user-provided constraints.

### Reproduction

1. Create a local audio track using the livekit-client
2. Create a constraints object with a new deviceId to switch to a different microphone
3. Call localTrack.restart(newConstraints)
4. Obtained a mediastream without the new deviceId applied
```typescript
// Reproduction code
import { LocalTrack, Room, Track, TrackEvent } from 'livekit-client';

async function testConstraintsBug() {
  // Connect to a room and get the local participant
  const room = new Room();
  await room.connect('ws://your-livekit-url', 'token');
  const localParticipant = room.localParticipant;
  
  // Get initial audio track
  const micPublication = localParticipant.getTrackPublication(Track.Source.Microphone);
  const audioTrack = micPublication?.track as LocalTrack;
  
  if (!audioTrack) {
    console.error('No microphone track found');
    return;
  }
  
  // Get available audio devices
  const audioDevices = await Room.getLocalDevices('audioinput');
  
  if (audioDevices.length > 1) {
    // Get current device ID
    const currentSettings = audioTrack.getMediaStreamTrack().getSettings();
    const currentDeviceId = currentSettings.deviceId;
    
    // Find a different device
    const newDevice = audioDevices.find(d => d.deviceId !== currentDeviceId);
    
    if (newDevice) {
      // Try to switch to the new device
      const newConstraints = {
        deviceId: { exact: newDevice.deviceId },
      };
      
      await audioTrack.restart(newConstraints);
    }
  }
}
```

### System Info

```shell
OS: macOS Ventura 13.4
Browser: Chrome 133.0.6943.55
Package Version: livekit-client 2.9.2
```

### Severity

blocking an upgrade

### Additional Information

_No response_

rktguswjd:
resolve #1434

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -311,10 +311,8 @@ export default abstract class LocalTrack<
     this.manuallyStopped = false;
     const unlock = await this.restartLock.lock();
     try {
-      if (!constraints) {
-        constraints = this._constraints;
-      }
-      const { deviceId, ...otherConstraints } = this._constraints;
+      constraints = { ...this._constraints, ...constraints ?? {} };
+      const { deviceId, ...otherConstraints } = constraints;
       this.log.debug('restarting track with constraints', { ...this.logContext, constraints });
 
       const streamConstraints: MediaStreamConstraints = {
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -311,9 +311,15 @@ export default abstract class LocalTrack<
     this.manuallyStopped = false;
     const unlock = await this.restartLock.lock();
     try {
-      constraints = { ...this._constraints, ...(constraints ?? {}) };
-      const { deviceId, ...otherConstraints } = constraints;
-      this.log.debug('restarting track with constraints', { ...this.logContext, constraints });
+      const trackConstraints: MediaTrackConstraints = {
+        ...this._constraints,
+        ...(constraints ?? {}),
+      };
+      const { deviceId, ...otherConstraints } = trackConstraints;
+      this.log.debug('restarting track with constraints', {
+        ...this.logContext,
+        constraints: trackConstraints,
+      });
 
       const streamConstraints: MediaStreamConstraints = {
         audio: false,
@@ -345,7 +351,7 @@ export default abstract class LocalTrack<
       this.log.debug('re-acquired MediaStreamTrack', this.logContext);
 
       await this.setMediaStreamTrack(newTrack);
-      this._constraints = constraints;
+      this._constraints = trackConstraints;
       this.emit(TrackEvent.Restarted, this);
       if (this.manuallyStopped) {
         this.log.warn(
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -343,6 +343,8 @@ export default abstract class LocalTrack<
       // 'A MediaStreamTrack ended due to a capture failure`
       this._mediaStreamTrack.stop();
 
+      this._constraints = trackConstraints;
+
       // create new track and attach
       const mediaStream = await navigator.mediaDevices.getUserMedia(streamConstraints);
       const newTrack = mediaStream.getTracks()[0];
@@ -351,7 +353,6 @@ export default abstract class LocalTrack<
       this.log.debug('re-acquired MediaStreamTrack', this.logContext);
 
       await this.setMediaStreamTrack(newTrack);
-      this._constraints = trackConstraints;
       this.emit(TrackEvent.Restarted, this);
       if (this.manuallyStopped) {
         this.log.warn(
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -311,7 +311,7 @@ export default abstract class LocalTrack<
     this.manuallyStopped = false;
     const unlock = await this.restartLock.lock();
     try {
-      constraints = { ...this._constraints, ...constraints ?? {} };
+      constraints = { ...this._constraints, ...(constraints ?? {}) };
       const { deviceId, ...otherConstraints } = constraints;
       this.log.debug('restarting track with constraints', { ...this.logContext, constraints });
 
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -311,15 +311,11 @@ export default abstract class LocalTrack<
     this.manuallyStopped = false;
     const unlock = await this.restartLock.lock();
     try {
-      const trackConstraints: MediaTrackConstraints = {
-        ...this._constraints,
-        ...(constraints ?? {}),
-      };
-      const { deviceId, ...otherConstraints } = trackConstraints;
-      this.log.debug('restarting track with constraints', {
-        ...this.logContext,
-        constraints: trackConstraints,
-      });
+      if (!constraints) {
+        constraints = this._constraints;
+      }
+      const { deviceId, ...otherConstraints } = constraints;
+      this.log.debug('restarting track with constraints', { ...this.logContext, constraints });
 
       const streamConstraints: MediaStreamConstraints = {
         audio: false,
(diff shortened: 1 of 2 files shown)


src/room/track/LocalTrack.ts
```diff
@@ -311,10 +311,8 @@ export default abstract class LocalTrack<
     this.manuallyStopped = false;
     const unlock = await this.restartLock.lock();
     try {
-      if (!constraints) {
-        constraints = this._constraints;
-      }
-      const { deviceId, ...otherConstraints } = this._constraints;
+      constraints = { ...this._constraints, ...(constraints ?? {}) };
```
lukasIO, line 314: this mutates the passed constraints, we'd want to create a separate local variable for this

lukasIO (commented):
thanks, just one thing that needs fixing

src/room/track/LocalTrack.ts
```diff
@@ -343,6 +343,8 @@ export default abstract class LocalTrack<
       // 'A MediaStreamTrack ended due to a capture failure`
       this._mediaStreamTrack.stop();
 
+      this._constraints = trackConstraints;
+
```
rktguswjd, line 347: I am modifying the order of operations in the restart method. I moved this._constraints = trackConstraints; before the setMediaStream call. Please review these changes!

The reason for this change is to prevent potential loss of deviceId information. Here's the current issue:
1. If we get a new track without a specified deviceId
2. Then in setMediaStreamTrack, we do `this._constraints = newTrack.getConstraints();` (which might obtain a new deviceId)
3. And then back in restart, we do `this._constraints = trackConstraints;`
4. This could result in losing the deviceId information

lukasIO: > we do this._constraints = newTrack.getConstraints();

I don't believe this is true, `getConstraints` should always return the constraints that have been used to construct the mediastreamtrack, see https://w3c.github.io/mediacapture-main/#dom-constrainablepattern-getconstraints

rktguswjd: I agree with your concern! Let me explain the situation in detail:
1. Let's say trackConstraints has undefined deviceId
2. streamConstraints becomes { audio: true }
3. getUserMedia is called with streamConstraints to obtain newTrack
4. setMediaStreamTrack is called
5. In setMediaStreamTrack, since newTrack exists, newTrack.getConstraints() is assigned to this.constraints
6. At this point, due to steps 2 and 3, the browser automatically selects an appropriate microphone, so that microphone's deviceId is also assigned to this.constraints
7. After setMediaStreamTrack execution completes, trackConstraints is overwritten to this.constraints
8. However, due to step 1, trackConstraints' deviceId is undefined

If anything is unclear or if you disagree with my explanation, please let me know! 🙏

- as-is
```typescript
const { deviceId } = { ...this._constraints, ...constraints?? {} } // deviceId is undefined
await this.setMediaStreamTrack(newTrack); //  this._constraints = { deviceId: "anything" }
this._constraints = trackConstraints; // this._constraints = { deviceId: undefined }
// or
const { deviceId } = { ...this._constraints, ...constraints?? {} } // deviceId is "abcd"
await this.setMediaStreamTrack(newTrack); //  this._constraints = { deviceId: "abcd" }
this._constraints = trackConstraints; // this._constraints = { deviceId: "abcd" }
```

- to-be
```typescript
const { deviceId } = { ...this._constraints, ...constraints?? {} } // deviceId is undefined
this._constraints = trackConstraints; // this._constraints = { deviceId: undefined }
await this.setMediaStreamTrack(newTrack);  //  this._constraints = { deviceId: "anything" }
//or
const { deviceId } = { ...this._constraints, ...constraints?? {} } // deviceId is "abcd"
this._constraints = trackConstraints; // this._constraints = { deviceId: "abcd" }
await this.setMediaStreamTrack(newTrack);  //  this._constraints = { deviceId: "abcd" }
```

lukasIO: oh, maybe you're right and this
>  that were the argument to the most recent successful invocation 

means that it's exclusively the last call to `applyconstraints` which would always be missing the device Id in our case.

rktguswjd: I agree with you, and I think my explanation was incomplete.
I was concerned about cases where this.constraints's deviceId becomes undefined before creating newTrack, like this:
```typescript
const trackConstraints = { ...this._constraints, ...constraints?? {} }
const { deviceId } = trackConstraints // deviceId is undefined
// omitted
const newTrack = await navigator.mediaDevices.getUserMedia({ audio: true }).getTracks()[0]
await this.setMediaStreamTrack(newTrack); //  this._constraints = { deviceId: "anything" }
this._constraints = trackConstraints; // this._constraints = { deviceId: undefined }
```
Even if such situations don't occur, I think there's no guarantee that `trackConstraints` and `newTrack.getConstraints()` will always be identical in the current code. Therefore, I believed it would be safer to assign `newTrack.getConstraints()` to this.constraints at the end.

If anything is unclear or if you disagree with my explanation, please let me know🙏

src/room/track/LocalTrack.ts
```diff
@@ -311,11 +311,15 @@ export default abstract class LocalTrack<
     this.manuallyStopped = false;
     const unlock = await this.restartLock.lock();
     try {
-      if (!constraints) {
-        constraints = this._constraints;
-      }
-      const { deviceId, ...otherConstraints } = this._constraints;
```
lukasIO, line 317: @rktguswjd thinking more about this I think the correct fix is simply to change this line to 
```
      const { deviceId, ...otherConstraints } = constraints;
```

rktguswjd: Aha, I thought we needed to merge this._constraints with the given constraints. I see that wasn't the intention! I'll fix it.

src/room/track/LocalTrack.ts
rktguswjd: @lukasIO Do you agree with this change? While this is a different context from the original issue we were trying to solve, I had provided an additional suggestion. If it's not valid, we can rollback this :)

src/room/track/LocalTrack.ts
lukasIO: thanks, if you can just revert that other change now, then we're good to go! 

The reason the constraints are overridden _after_ setMediaStreamTrack is because the constraints passed in by the user are the _desired_ constraints. We want to reuse the desired constraints after, too. And as we don't merge any constraints together this should be sufficient

src/room/track/LocalTrack.ts
lukasIO: yeah, please revert this. If you have a reproduction of a case where this causes an actual problem feel free to open a separate issue/PR and I'll take a look :)

lukasIO (approved):
thank you for catching this!

rktguswjd:
@lukasIO Oh, sorry. I did a force push and the changeset commit disappeared🥲

PR #1437, opened 2025-03-12
log ice candidates as debug

haydenbr:


lukasIO (approved):
that makes sense, thanks!

lukasIO:
expecting a patch release will be rolled out tomorrow!

PR #1440, opened 2025-03-13
Fix trailing slash handling in WebSocket URL pathname

### LINKED ISSUE #1439: Connection failure issue when the WebSocket URL for the LiveKit server has no path.
### Describe the bug

If the WebSocket URL passed to the connect function does not contain a path, the connection fails, as shown in the image below.
This appears to be due to the inability to set an empty string for urlObj.pathname.
This issue has been occurring since updating to version 2.9.6.

Image

### Reproduction

1. Pass a WebSocket URL without a path when calling the connect function (e.g., 'wss://192.168.0.211:8073').
2. An extra, unnecessary slash is added to the WebSocket URL, causing the connection to fail (e.g., 'wss://192.168.0.211:8073//rtc').

### Logs

```shell

```

### System Info

```shell
System:
    OS: Windows 10 Pro
    CPU: Intel(R) Core(TM) i9-10900F CPU @ 2.80GHz   2.81 GHz
    Memory: 16 GB / 32.00 GB
  Binaries:
    Node: 17.9.1
    Yarn: 1.22.22
    npm: 10.9.0
  Browsers:
    Chrome: 134.0.6998.35
  npmPackages:
    livekit-client: 2.9.4 => 2.9.6
```

### Severity

blocking an upgrade

### Additional Information

_No response_

Tanney-102:
resolve #1439 

Previously, the code attempted to remove the trailing slash from urlObj.pathname by setting it to an empty string and then appending "/rtc". However, since URL objects always return "/" for an empty path when a host is present, the existing approach never truly removed the slash. This PR updates the logic to explicitly check if the pathname ends with a slash. If it does, we simply append "rtc". Otherwise, we append "/rtc". This ensures the final path is correctly formed without relying on the pathname being set to an empty string.

## References
- https://url.spec.whatwg.org/#dom-url-pathname
- https://url.spec.whatwg.org/#url-path-serializer
- https://url.spec.whatwg.org/#example-url-components

### DIFF
--- a/.changeset/cyan-berries-divide.md
+++ b/.changeset/cyan-berries-divide.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Fix trailing slash handling in WebSocket URL pathname
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -260,7 +260,7 @@ export class SignalClient {
     this.connectOptions = opts;
     const urlObj = new URL(toWebsocketUrl(url));
     // strip trailing slash
-    const hasTrailingSlash = urlObj.pathname.endsWith('/')
+    const hasTrailingSlash = urlObj.pathname.endsWith('/');
     urlObj.pathname += hasTrailingSlash ? 'rtc' : '/rtc';
 
     const clientInfo = getClientInfo();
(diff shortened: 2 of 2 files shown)


lukasIO:
thanks for catching this!

PR #1443, opened 2025-03-15
Use string instead of passing url object to WebSocket constructor

davidliu:
React-Native (and WebSockets JS spec) can only take a string in the WebSocket constructor.

Details here: https://github.com/facebook/react-native/issues/49382

### DIFF
--- a/.changeset/empty-crews-remember.md
+++ b/.changeset/empty-crews-remember.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


davidzhao (approved):

IIvexII:
Thanks for fixing this issue, I was stuck for like 6 hours straight trying everything i could without having any idea, what was causing it.

lukasIO (approved):

PR #1456, opened 2025-03-26
Add preferred codec for backup codec policy

cnderrauber:
relates https://github.com/livekit/protocol/pull/1002

Also rename the enums according to js naming convention.

### DIFF
--- a/src/room/track/options.ts
+++ b/src/room/track/options.ts
@@ -389,8 +389,12 @@ export function isBackupCodec(codec: string): codec is BackupVideoCodec {
 }
 
 export enum BackupCodecPolicy {
-  REGRESSION = 0,
-  SIMULCAST = 1,
+  // codec regression is preferred, the sfu will try to regress codec if possible but not guaranteed
+  PreferRegression = 0,
+  // multi-codec simulcast, publish both primary and backup codec at the same time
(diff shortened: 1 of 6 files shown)


lukasIO:
there's a mismatch between the protocol defined backupcodecpolicy and the one redefined in the SDK, that's why the CI is failing

src/room/track/options.ts
```diff
@@ -389,8 +389,12 @@ export function isBackupCodec(codec: string): codec is BackupVideoCodec {
 }
 
 export enum BackupCodecPolicy {
-  REGRESSION = 0,
-  SIMULCAST = 1,
+  // codec regression is preferred, the sfu will try to regress codec if possible but not guaranteed
+  PreferRegression = 0,
```
lukasIO, line 393: not sure if renaming these to camel case is worth it.
1. it breaks for users who used the enum previously
2. it needs the forced typecast with `as unknown` which will hide incompatibilities 

I think it's better to leave them as uppercase for the reasons above

cnderrauber: for `2`, I think no user use it now so it is not a problem. if uppercase is ok here, I can revert the naming change.

lukasIO (approved):
thanks!

PR #1465, opened 2025-04-01
Defer publishing until signal is connected

lukasIO:
This PR defers publishing attempts until signal is connected. 
There's a 15s timeout between publishing attempt until signal connection is connected.

Previous behaviour was that publishing on a room that's not connected would simply fail. 

This change allows for patterns where device acquisition can happen in parallel to room connection, e.g. 

```ts
await Promise.all([room.connect(...), room.localParticipant.enableMicrophoneAndCamera()]);
```

### DIFF
--- a/.changeset/eight-horses-raise.md
+++ b/.changeset/eight-horses-raise.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": minor
+---
+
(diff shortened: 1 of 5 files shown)


davidzhao (approved):
lgtm

PR #1467, opened 2025-04-02
fix `hostname//rtc/validate` issue

edmond-baton:


src/api/SignalClient.ts
```diff
@@ -322,9 +321,13 @@ export class SignalClient {
             this.state = SignalConnectionState.DISCONNECTED;
             clearTimeout(wsTimeout);
             try {
-              const validateURL = new URL(urlObj.toString());
-              validateURL.protocol = `http${validateURL.protocol.substring(2)}`;
-              validateURL.pathname += '/validate';
+              // Build validate URL relative to the original rtcUrl base
+              const validateURL = new URL('validate', rtcUrl);
+              // Reset search params for validate request if needed, or keep rtcUrl's? Assuming reset for now.
```
lukasIO, line 326: we need to keep these for the validate URL

lukasIO (commented):
thanks for tackling this, this looks a lot cleaner!

There's been a bunch of movement in this code path over the last versions so I'm going to follow up with a PR that adds testing and extracts some of these methods. 

The PR looks good to go apart from the search param reset

edit: actually there's a problem here with the usage of `new URL(path, base)` as the URL constructor will only consider paths as base until the _last_ trailing slash, so we would get different behaviours between trailing and non trailing slash URLs

lukasIO:
for reference, these are the tests that need to pass: https://github.com/livekit/client-sdk-js/pull/1468

lukasIO:
thanks for reporting this, I went ahead and merged #1468 with a testable fix

PR #1479, opened 2025-04-10
Default to exact matching and retry on unspecified ideal narrowing

lukasIO:
for `createLocalTracks` this PR changes our stream acquisition handling to default to `exact` matching if neither `ideal` nor `exact` have been explicitly specified by the user

1. user passes in `{ deviceId: "my-device-id" }`: Try with exact match if that throws -> try again with "ideal"
2.  `{ deviceId: { exact: "my-device-id"  } }`:  Same behaviour as before, throws overconstrainederror if it cannot be fulfilled
3. `{ deviceId: { ideal: "my-device-id" }}`: Browser's ideal behaviour is respected


davidzhao (approved):
lgtm

PR #1485, opened 2025-04-11
Request audio/video in create helpers

lukasIO:
previously those helpers were passing audio and video respectively as `undefined` if no explicit options were set.


davidzhao (approved):
lg

PR #1490, opened 2025-04-17
Add option (`KeyProviderOptions`) to allowKeyExtraction.

toger5:
This is helpful for systems where only the ratcheted key should be shared with new participants so that new joiners cannot decrypt historic media from the call. It provides an option so that the emitted ratcheted key can be extracted (read as a js variable) and therfore can be shared (in a secure way) over the network to the newjoiner

This touches a couple of ares:
 - the `importKey` method gets an optional property to import an extractable key
  - The e2ee.worker will use the new importKey property based on the data in the `RatchetRequestMessage`
  - The E2eeManager will read the keyProviderOptions and send use the `allowKeyExtraction` to configure the `RatchetRequestMessage`

### DIFF
--- a/src/e2ee/E2eeManager.ts
+++ b/src/e2ee/E2eeManager.ts
@@ -233,7 +233,7 @@ export class E2EEManager
   private postRatchetRequest(
     participantIdentity?: string,
     keyIndex?: number,
-    extracable?: boolean,
+    extractable?: boolean,
   ) {
     if (!this.worker) {
       throw Error('could not ratchet key, worker is missing');
@@ -243,7 +243,7 @@ export class E2EEManager
       data: {
         participantIdentity: participantIdentity,
         keyIndex,
-        extracable,
+        extractable,
       },
     };
     this.worker.postMessage(msg);
--- a/src/e2ee/E2eeManager.ts
+++ b/src/e2ee/E2eeManager.ts
@@ -222,11 +222,19 @@ export class E2EEManager
     keyProvider
       .on(KeyProviderEvent.SetKey, (keyInfo) => this.postKey(keyInfo))
       .on(KeyProviderEvent.RatchetRequest, (participantId, keyIndex) =>
-        this.postRatchetRequest(participantId, keyIndex),
+        this.postRatchetRequest(
+          participantId,
+          keyIndex,
+          keyProvider.getOptions().allowKeyExtraction,
+        ),
       );
   }
 
-  private postRatchetRequest(participantIdentity?: string, keyIndex?: number) {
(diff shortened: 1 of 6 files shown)


BillCarsonFr:
For the usecase where you send only the ratcheted key to the new joiner, we might also want to have a little delay to distribute the key before starting using it.
Is it possible with this change?

BillCarsonFr:
> For the usecase where you send only the ratcheted key to the new joiner, we might also want to have a little delay to distribute the key before starting using it. Is it possible with this change?

After discussion you are right, this won't help much to delay.

src/e2ee/E2eeManager.ts
```diff
@@ -222,11 +222,19 @@ export class E2EEManager
...
+          keyProvider.getOptions().allowKeyExtraction,
+        ),
       );
   }
 
-  private postRatchetRequest(participantIdentity?: string, keyIndex?: number) {
+  private postRatchetRequest(
+    participantIdentity?: string,
+    keyIndex?: number,
+    extracable?: boolean,
```
lukasIO, line 236: nit
```suggestion
    extractable?: boolean,
```

src/e2ee/worker/e2ee.worker.ts
```diff
@@ -119,11 +119,11 @@ onmessage = (ev) => {
 async function handleRatchetRequest(data: RatchetRequestMessage['data']) {
   if (useSharedKey) {
     const keyHandler = getSharedKeyHandler();
-    await keyHandler.ratchetKey(data.keyIndex);
+    await keyHandler.ratchetKey(data.keyIndex, undefined, data.extracable);
```
lukasIO, line 122: also only a small nit, just to make the behaviour slightly more obvious here

```suggestion
    await keyHandler.ratchetKey(data.keyIndex, true, data.extracable);
```

lukasIO (commented):
thanks for this! 

Just two nits looks good to me otherwise

src/e2ee/types.ts
```diff
@@ -74,6 +74,7 @@ export interface RatchetRequestMessage extends BaseMessage {
   data: {
     participantIdentity?: string;
     keyIndex?: number;
+    extracable?: boolean;
```
lukasIO, line 77:
```suggestion
    extractable?: boolean;
```

lukasIO (approved):
thank you 🙏

BillCarsonFr:
Notice that ratcheted key as they are now will fail if `extractable` is true:
`SyntaxError: KDF keys must set extractable=false`

PR #1492, opened 2025-04-18
Convert URL Scheme When Creating WebSocket Object

### LINKED ISSUE #1491: WebSocket Connection Fails
### Describe the bug

After version 2.10.0, there have been instances where the connect function fails to create a WebSocket object when provided with a URL using the `https://` scheme. This issue has occurred in older versions of certain browsers. (There is no problem in the latest version of browser.)

- Expected Behavior: The connect function should successfully establish a WebSocket connection even when the URL uses the `https://` scheme.
- Actual Behavior: The connect function fails to create a WebSocket object when provided with a URL using the `https://` scheme in certain versions of some browsers.

### Reproduction

1. Use the Safari browser to call the connect function of SignalClient with a URL using the `https://` scheme.
2. Observe the WebSocket connection error. // Wrong url scheme for WebSocket ...

### Logs

```shell

```

### System Info

```shell
OS: macOS Ventura 13.4
Browser: Safari 16.5(18615.2.9.11.4)
Package Version: livekit-client 2.9.2
```

### Severity

annoyance

### Additional Information

_No response_

rktguswjd:
resolve #1491 
Add logic to convert the scheme to `ws://` or `wss://` when creating the rtcUrl to ensure a successful WebSocket connection. Use the toWebsocketUrl function for this purpose.

### DIFF
--- a/.changeset/breezy-cups-laugh.md
+++ b/.changeset/breezy-cups-laugh.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Convert URL Scheme When Creating WebSocket Object
--- a/src/api/utils.test.ts
+++ b/src/api/utils.test.ts
@@ -9,6 +9,13 @@ describe('createRtcUrl', () => {
     expect(result.toString()).toBe('wss://example.com/rtc');
   });
 
+  it('should create a basic RTC URL with http protocol', () => {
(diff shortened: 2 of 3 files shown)


lukasIO (approved):
Thanks for the fix and for adding the test case! 
This got lost in some edit, but was the intention from the start

PR #1503, opened 2025-04-25
feature: E2E Allow sharing ratcheted material out-of-band

BillCarsonFr:
One of the interesting thing with ratcheting is that when there is a new member joining a call, it would be possible to only share with him the ratcheted key instead of rotating the existing key and reshare to everyone.

Unfortunatly with the exisiting code it is not possible to share the ratcheted key material out-of-band because it is not exportable (as per [webcrypto limitiation](https://github.com/w3c/webcrypto/pull/113))

This PR introduce a new event `RatchetRequestCompleted` that will be emitted after a `RatchetRequest` has been completed.
This event contains a ArrayBuffer, that can be transmitted out-of-band and used by a new joiner.

### DIFF
--- a/src/e2ee/worker/ParticipantKeyHandler.ts
+++ b/src/e2ee/worker/ParticipantKeyHandler.ts
@@ -2,7 +2,7 @@ import { EventEmitter } from 'events';
 import type TypedEventEmitter from 'typed-emitter';
 import { workerLogger } from '../../logger';
 import { KeyHandlerEvent, type ParticipantKeyHandlerCallbacks } from '../events';
-import type { KeyProviderOptions, KeySet } from '../types';
+import type { KeyProviderOptions, KeySet, RatchetResult } from '../types';
 import { deriveKeys, importKey, ratchet } from '../utils';
 
 // TODO ParticipantKeyHandlers currently don't get destroyed on participant disconnect
@@ -25,7 +25,7 @@ export class ParticipantKeyHandler extends (EventEmitter as new () => TypedEvent
 
   private keyProviderOptions: KeyProviderOptions;
 
-  private ratchetPromiseMap: Map<number, Promise<CryptoKey>>;
+  private ratchetPromiseMap: Map<number, Promise<RatchetResult>>;
 
   private participantIdentity: string;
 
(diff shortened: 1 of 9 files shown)


src/e2ee/worker/ParticipantKeyHandler.ts
```diff
@@ -126,17 +126,13 @@ export class ParticipantKeyHandler extends (EventEmitter as new () => TypedEvent
...
-        );
-
+        const chainKey = await ratchet(currentMaterial, this.keyProviderOptions.ratchetSalt);
+        const newMaterial = await importKey(chainKey, currentMaterial.algorithm.name, 'derive');
         if (setKey) {
           await this.setKeyFromMaterial(newMaterial, currentKeyIndex, true);
           this.emit(
-            KeyHandlerEvent.KeyRatcheted,
-            newMaterial,
+            KeyHandlerEvent.RatchetRequestCompleted,
```
lukasIO, line 134: Currently this change would stop emitting the `KeyRatcheted` event which would be a breaking change. 
We could emit _both_ events, but I'm wondering if it's cleaner to just add the `chainKey` param to the `KeyRatcheted` payload after the key index. 
That should be - practically - non breaking and leave the events in place as they are

lukasIO (commented):
thanks for this! This seems to be in line with what the PR you linked describes as the recommended workaround (using deriveBits).

src/e2ee/worker/ParticipantKeyHandler.ts
BillCarsonFr: Agree, I was a bit confused because of the `true` in `this.setKeyFromMaterial(newMaterial, currentKeyIndex, true)` is already emiting `.KeyRatcheted` event

BillCarsonFr: @lukasIO I added back the initial emitted event in order to stay non-breaking  (see bd60545b3a25392225faa4aec551bc5e1129530a)

But it is strange that this is emitted twice in a row (the second time it is with the actual strecthed secret used to encrypt). I suspect there is a bug somewhere.

The goal here is to stay non-breaking.

lukasIO: uff, you're right! Not sure how I didn't pick that up earlier, sorry!

Timing wise the second event (emitted from within setKeySet) probably makes more sense. 
Leaves the question whether we'd pass through the ratcheted key or emit a separate event. 

we could have `emitRatchetEvent` take `bool | ArrayBuffer` for passing

BillCarsonFr: @lukasIO I am back to this. I just pushed a new commit in order to keep the existing RatchetEvent instead of using a separate event

lukasIO (approved):
this looks good to me, thanks for making this more useful!

BillCarsonFr:
> this looks good to me, thanks for making this more useful!

Hi @lukasIO, I was wondering if you have an idea of when this change could be merged? any ETA for a release?
Thx, Valere.

lukasIO:
sorry @BillCarsonFr, simply forgot to merge this after my testing! 

Thanks again!

PR #1511, opened 2025-05-09
Support moving participant to another room

cnderrauber:


src/room/Room.ts
```diff
@@ -589,6 +589,24 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
           this.localParticipant,
         );
+      })
+      .on(EngineEvent.RoomMoved, (roomMoved) => {
+        this.log.debug('room moved', roomMoved);
+
+        if (roomMoved.room) {
+          this.handleRoomUpdate(roomMoved.room);
+        }
+        this.emit(RoomEvent.Moved, roomMoved.room?.name ?? '', roomMoved.token);
```
lukasIO, line 599: should this happen before disconnecting remote participants or after 🤔

cnderrauber: the particiant has been moved then other participant disappered so the application can know it is caused by participant moved, do you prefer to fire the move event after disconnect remote particpant?

lukasIO: it's tricky, my feeling is that would align better if the participants disconnect before moving as at least the room will then still be the "correct" one at that point in time

lukasIO: this is also the order of events on disconnect

src/room/Room.ts
```diff
@@ -596,12 +596,13 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
         if (roomMoved.room) {
           this.handleRoomUpdate(roomMoved.room);
         }
-        this.emit(RoomEvent.Moved, roomMoved.room?.name ?? '', roomMoved.token);
 
         this.remoteParticipants.forEach((participant, identity) => {
           this.handleParticipantDisconnected(identity, participant);
         });
 
+        this.emit(RoomEvent.Moved, roomMoved.room?.name ?? '', roomMoved.token);
```
lukasIO, line 604: one last thing, under which circumstances would we expect `room` to not be defined on the `roomMoved` message here? 
maybe we could raise an error if it's missing instead of the `?? ''`

lukasIO (approved):

src/room/Room.ts
cnderrauber: it should not happen, just avoid compile error

PR #1517, opened 2025-05-14
Add ParticipantActive event to signal data message readiness

lukasIO:


src/room/events.ts
```diff
@@ -203,6 +203,13 @@ export enum RoomEvent {
    */
   ParticipantAttributesChanged = 'participantAttributesChanged',
 
+  /**
+   * Emitted when the participant's state changes to active and is ready to send/receive data messages
```
boks1971, line 207: maybe upper case `ACTIVE` to indicate that it is a state value? Not a strong suggestion though.

src/room/participant/Participant.ts
```diff
@@ -224,12 +229,14 @@ export default class Participant extends (EventEmitter as new () => TypedEmitter
...
     this._setAttributes(info.attributes);
+    if (info.state === ParticipantInfo_State.ACTIVE) {
+      this.emit(ParticipantEvent.Active);
+    }
     if (info.permission) {
       this.setPermissions(info.permission);
     }
     // set this last so setMetadata can detect changes
     this.participantInfo = info;
-    this.log.trace('update participant info', { ...this.logContext, info });
```
boks1971, line 232: intentional remove?

src/room/participant/Participant.ts
```diff
@@ -224,12 +229,14 @@ export default class Participant extends (EventEmitter as new () => TypedEmitter
     this._setName(info.name);
     this._setMetadata(info.metadata);
     this._setAttributes(info.attributes);
+    if (info.state === ParticipantInfo_State.ACTIVE) {
```
boks1971, line 232: I think this probably needs some prev state? This should emit on transition from not-ACTIVE -> ACTIVE only right?

src/room/participant/Participant.ts
lukasIO: yeah, we added this way back to debug some unexpected participant state updates, but haven't needed it since

src/room/participant/Participant.ts
lukasIO: uff, of course, thanks!

lukasIO: addressed in [df8ef65](https://github.com/livekit/client-sdk-js/pull/1517/commits/df8ef6513b60fe71c2d279b4c0d29046577df6aa)

lukasIO:
thanks for the review and sorry that I just pushed another commit with a little helper. 

I realised most people might want a `participant.waitUntilActive()` helper to not have to write the boiler plate each time with the event listener. 
Was slightly more involved than expected as it needs to reject the promise on participant disconnects

boks1971:
> thanks for the review and sorry that I just pushed another commit with a little helper.
> 
> I realised most people might want a `participant.waitUntilActive()` helper to not have to write the boiler plate each time with the event listener. Was slightly more involved than expected as it needs to reject the promise on participant disconnects

Nice!!!

bcherry (approved):

PR #1519, opened 2025-05-16
Fix svc encoding for safari 18.4

cnderrauber:
Safari18.4 requires scaleResolutionDown to be set
with legacy svc api.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -216,7 +216,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       room: this.latestJoinResponse?.room?.name,
       roomID: this.latestJoinResponse?.room?.sid,
       participant: this.latestJoinResponse?.participant?.identity,
-      pID: this.latestJoinResponse?.participant?.sid,
+      pID: this.participantSid,
     };
   }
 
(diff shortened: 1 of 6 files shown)


src/room/RTCEngine.ts
```diff
@@ -532,6 +532,9 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
     this.client.onRoomMoved = (res: RoomMovedResponse) => {
       this.participantSid = res.participant?.sid;
+      if (this.latestJoinResponse) {
+        this.latestJoinResponse.room = res.room;
+      }
```
boks1971, line 537: should this be setting other participants also from MoveResponse? Maybe, the only thing used from lastesJoinResponse is the room?

lukasIO (approved):
thank you!

src/room/RTCEngine.ts
cnderrauber: It is a internal property and only room and server info (verison, nodeID...) are used, but `RoomMoved` will not update server info, so it is ok here.

boks1971: Would be good to add server info to room moved at some point later.

cnderrauber: MoveParticipant will not change the server for client, maybe worth to add it to `ReconnectResponse` since migration will change server node.

boks1971: Good point. Makes sense.

PR #1521, opened 2025-05-17
[BUG] Remove duplicate media devices

yashjain-99:
Fixes: [components-js/issues/1139](https://github.com/livekit/components-js/issues/1139) 

- Currently in chrome/firefox we see multiple instances of same device.
- Fixed it by removing duplicates based on groupId and kind.


lukasIO:
thanks for the PR! 

we initially filtered these devices out, but decided to change that a while back because they're not actually the same device. 
In chrome one of them will have `(Default)` appended to the label, which means if you were to change your default device on an OS level when the default is selected it will also switch in the app.

PR #1526, opened 2025-05-19
[FIX] Update active device immediately for muted audio tracks

yashjain-99:
Fixes: [components-js/issues/1134](https://github.com/livekit/components-js/issues/1134) 

- When changing audio devices while tracks are muted, the UI didn’t update because muted tracks don’t trigger automatic device restarts.
- Users expected device changes to reflect immediately in the UI, even when muted.
- Fix: Added `shouldTriggerImmediateDeviceChange` flag to detect muted tracks and trigger update if true.

### DIFF
--- a/.changeset/ninety-baboons-act.md
+++ b/.changeset/ninety-baboons-act.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Update active device immediately for muted audio tracks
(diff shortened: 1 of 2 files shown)


KenKolano:
There were similar problems for cameras that were off, does this only address this for mics?

yashjain-99:
> There were similar problems for cameras that were off, does this only address this for mics?

Ideally it should work for video devices as well. I tested mocking multiple video devices it is working as expected.

KenKolano:
Sorry to report, but the camera selection is still broken with this patch.

Showing current results with the 2.14.0 client version.

https://github.com/user-attachments/assets/c53908a0-0cab-400b-b5d8-71c3c207fea7

PR #1543, opened 2025-06-03
Virtual camera plugin standard proposal.

Vislouhi:
Iframe runs virtual camera UI and functionality.
It receives, processes and sends media with WebRTC.
Can be used to integrate any speech-driven real-time avatars in web conference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 Use this SDK to add realtime video, audio and data features to your JavaScript/TypeScript app. By connecting to <a href="https://livekit.io/">LiveKit</a> Cloud or a self-hosted server, you can quickly build applications such as multi-modal AI, live streaming, or video calls with just a few lines of code.
 <!--END_DESCRIPTION-->
 
-> **New feature**: Custom virtual camera.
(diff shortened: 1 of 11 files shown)


Vislouhi:
Thanks for reviewing my commit. Unlike avatars in agents I propose an
avatar that is generated right at the user device without any backend.
No server exists that can run an agent with an avatar. It is a virtual
camera that transforms audio to video in realtime on the user's device.
Noone can use an agent for this purpose. I am suggesting a standard for
real-time transformation of audio to video on frontend.
I am not insisting on the way I did that, I want to open the discussion.

Here what is written on link you gave:
Overview <https://docs.livekit.io/agents/integrations/avatar/#overview>

Virtual avatars add lifelike video output for your voice AI agents. You can
integrate a variety of providers to LiveKit Agents with just a few lines of
code.


My proposal is not an avatar for an AI agent. But speaking avatar for
the user who has no camera or does not want to use it.

lukasIO:
Sorry for the misunderstanding! 

This is definitely an interesting use case. I feel this mainly needs a usage example rather than deeply integration in the SDK?

I'm not sure I understand what parts you'd like to standardise around this?

PR #1546, opened 2025-06-11
e2e reliablility for data channel

cnderrauber:
* Resend reliable message from last sequence received by server after connection resumed
* Report last received reliable message from other participants on resuming

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -5,6 +5,7 @@ import {
   ClientConfiguration,
   type ConnectionQualityUpdate,
   DataChannelInfo,
+  DataChannelReceiveState,
   DataPacket,
   DataPacket_Kind,
   DisconnectReason,
@@ -76,11 +77,14 @@ import {
   supportsAddTrack,
   supportsTransceiver,
 } from './utils';
+import { DataPacketBuffer } from '../utils/dataPacketBuffer';
+import { TTLMap } from '../utils/ttlmap';
 
 const lossyDataChannel = '_lossy';
 const reliableDataChannel = '_reliable';
 const minReconnectWait = 2 * 1000;
 const leaveReconnect = 'leave-reconnect';
+const reliabeReceiveStateTTL = 30_000; 
 
 enum PCState {
   New,
@@ -177,6 +181,12 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
   private loggerOptions: LoggerOptions;
 
   private publisherConnectionPromise: Promise<void> | undefined;
+
+  private reliableDataSequence: number = 1;
+
+  private reliableMessageBuffer = new DataPacketBuffer(); 
+
+  private reliableReceivedState: TTLMap<string, number> = new TTLMap(reliabeReceiveStateTTL);
 
   constructor(private options: InternalRoomOptions) {
     super();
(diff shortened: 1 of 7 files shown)


src/room/RTCEngine.ts
```diff
@@ -677,6 +690,15 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       }
       const dp = DataPacket.fromBinary(new Uint8Array(buffer));
 
+      if (dp.sequence > 0 && dp.participantSid !== '') {
+        const lastSeq = this.reliableReceivedState.get(dp.participantSid);
+        if (lastSeq && dp.sequence <= lastSeq) {
+          // ignore duplicate or out-of-order packets in reliable channel
```
lukasIO, line 696: will it be required to guard against this across client implementations in general?

src/room/RTCEngine.ts
```diff
@@ -1151,19 +1180,42 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   /* @internal */
   async sendDataPacket(packet: DataPacket, kind: DataPacket_Kind) {
-    const msg = packet.toBinary();
-
     // make sure we do have a data connection
     await this.ensurePublisherConnected(kind);
 
+    if (kind === DataPacket_Kind.RELIABLE) {
```
lukasIO, line 1186: hmm, thinking this.ensurePublisherConnection might race for parallel `sendDataPacket` s so I'm not sure if it's guaranteed to respect invocation order.

src/room/RTCEngine.ts
```diff
@@ -1151,19 +1180,42 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
     }
 
     this.updateAndEmitDCBufferStatus(kind);
   }
 
+  private async resendReliableMessagesForResume(lastMessageSeq: number) {
+    await this.ensurePublisherConnected(DataPacket_Kind.RELIABLE);
+    const dc = this.dataChannelForKind(DataPacket_Kind.RELIABLE);
+    if (dc) {
+      this.reliableMessageBuffer.popToSequence(lastMessageSeq);
```
lukasIO, line 1211: are we sure that sending them at this point will actually succeed?

Is it safe to pop here already?

src/room/RTCEngine.ts
cnderrauber: It is double check that the server has sequence check too, add it here to avoid rare case that the client receives inflight message from old node when switching connection during migrtaiton.

src/room/RTCEngine.ts
cnderrauber: There is no `await` between increase sequence and `dc.Send` so the messages in the wire are in order and we can guaranteed the message has been received in the order which is sent. The `sendxxx` methods are `async` so I think the developer needs to wait for the first call completed before the next call if the message order is matter?

src/room/RTCEngine.ts
cnderrauber: The `dc.Send` will append those message to sctp's buffer and sent by the sctp transport, the transport will take care of sending the message to server side and it only fails when the ice connection is failed.

src/room/RTCEngine.ts
lukasIO: yeah, that's right.
I just wonder what happens to our buffer logic here if the message order doesn't match the sequence number. 
We simply pop and don't really look at the actual sequence number order in the buffer, right?

cnderrauber: No, the message order in the buffer should be same as the sequence number. I assume the message will be in order since there is no `async` statements between increase the number and push to buffer, is it wrong in the typescript?

lukasIO: got it, yeah, I think your assumption is correct!

PR #1547, opened 2025-06-12
Drop outdated sdp answers and forward offer ids

lukasIO:
depends on https://github.com/livekit/protocol/pull/1105

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -491,8 +491,10 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       if (!this.pcManager) {
         return;
       }
-      const answer = await this.pcManager.createSubscriberAnswerFromOffer(sd);
-      this.client.sendAnswer(answer, offerId);
+      const answer = await this.pcManager.createSubscriberAnswerFromOffer(sd, offerId);
+      if (answer) {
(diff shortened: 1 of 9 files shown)


src/room/RTCEngine.ts
```diff
@@ -1390,16 +1398,22 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-          ? toProtoSessionDescription({
-              sdp: previousOffer.sdp,
-              type: previousOffer.type,
-            })
+          ? toProtoSessionDescription(
+              {
+                sdp: previousOffer.sdp,
+                type: previousOffer.type,
+              },
+              this.latestRemoteOfferId,
```
boks1971: this is not getting used on server side currently, but just for correctness, this should not be `lastRemoteOfferId`?

In fact, `latestRemoteOfferId` is probably not needed as the regular answer path has the remote offer id in the closure and does not need state variable. And sync state does not use the id. So, maybe, it can be dropped?

bcherry (approved):

src/room/RTCEngine.ts
lukasIO: makes sense, thanks for catching that! I'll also remove it from `previousAnswer`

PR #1551, opened 2025-06-16
Ensure offerIds are ignored for 0 values

lukasIO:


boks1971 (commented):
Sorry about this @lukasIO . Is this due to not accepting `lastOfferId` in `SyncState`? If that is the case, we can add it and we can accept it on server side if that makes this more robust.

boks1971:
Got it. Will take a look.

PR #1552, opened 2025-06-17
Stop MediaStream tracks at the end of the video check

svajunas-budrys:
Stops MediaStream tracks at the end of the video check to release the camera and prevent it from staying active.


lukasIO (approved):
this makes sense, thanks!

PR #1553, opened 2025-06-18
wip cpu constrained publishing detection

lukasIO:


examples/demo/demo.ts
```diff
@@ -166,7 +166,13 @@ const appActions = {
     await room.prepareConnection(url, token);
     const prewarmTime = Date.now() - startTime;
     appendLog(`prewarmed connection in ${prewarmTime}ms`);
-
+    room.localParticipant.on(ParticipantEvent.LocalTrackCpuConstrained, (track, publication) => {
+      console.log('track is cpu constrained, lowering capture resolution', track, publication);
+      track.restartTrack({
```
davidzhao: should we turn off simulcast & reset layers too?

lukasIO: my hope was that this would happen automatically as browsers don't allow for simulcast layers to be added below a certain original resolution. E.g. max of 2 layers for 360p as defined [here](https://chromium.googlesource.com/external/webrtc/+/HEAD/media/engine/simulcast.cc#100). But maybe we're sidestepping that step by calling replaceTrack with a lower resolution? 🤷 


If we need to do it explicitly, then turning off simulcast will have two paths: 
- turn off simulcast for vp8/h264
- override scalabilityMode for SVC codecs to `L1T3` and ensure simulcast layers for backup codecs gets disabled

probably makes sense to hide this complexity behind something like a `track.reduceLayers()` helper?

davidzhao: with SVC I'm not sure if that behavior is inherited.. so it'd be preferable to set them explicitly

also for the example, I think we shouldn't restart the track each time this signal is received. since when it flips back, we'd restart the track again. It'd be great if it's ran once per session - anytime the CPU is an issue, we'd just downgrade to 360p for the rest of the session.

lukasIO: yeah, explicitly setting makes sense to me. 

>  I think we shouldn't restart the track each time this signal is received

the event is emitted only once for a particular mediastreamtrack track right now. 
Only after it's being restarted, we reset the flag to allow the event to be emitted again.
Exclusively setting the layers without changing the capture resolution would still be potentially overloading the CPU though because when thinking about background processors reducing the input resolution will have a much more significant performance impact that scaling the resolution down on the sender.

lukasIO: I added a `prioritizePerformance` method in https://github.com/livekit/client-sdk-js/pull/1553/commits/588fbe1dcc50856ca8b0f1accd3c622e5af42527, let me know what you think

davidzhao (approved):

PR #1559, opened 2025-06-19
 Fix transceiver reuse in chrome when using e2ee

lukasIO:
the reason #1554 didn't work for Safari was an uncaught type error in the script transform path. 

this PR re-applies the previous fix and adds types for the options passed to the transform event in order to prevent similar issues from happening again.

### DIFF
--- a/src/e2ee/worker/e2ee.worker.ts
+++ b/src/e2ee/worker/e2ee.worker.ts
@@ -257,10 +257,12 @@ function handleSifTrailer(trailer: Uint8Array) {
 }
 
 // Operations using RTCRtpScriptTransform.
+// @ts-ignore
 if (self.RTCTransformEvent) {
   workerLogger.debug('setup transform event');
   // @ts-ignore
   self.onrtctransform = (event: RTCTransformEvent) => {
+    // @ts-ignore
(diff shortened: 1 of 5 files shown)


src/e2ee/worker/e2ee.worker.ts
```diff
@@ -259,14 +262,14 @@ if (self.RTCTransformEvent) {
...
-    // @ts-ignore monkey patching non standard flag
-    transformer.handled = true;
-    const { kind, participantIdentity, trackId, codec } = transformer.options;
+
+    const { kind, participantIdentity, trackId, codec } =
+      transformer.options as ScriptTransformOptions;
     const cryptor = getTrackCryptor(participantIdentity, trackId);
     workerLogger.debug('transform', { codec });
-    cryptor.setupTransform(kind, transformer.readable, transformer.writable, trackId, codec);
+    cryptor.setupTransform(kind, transformer.readable, transformer.writable, trackId, false, codec);
```
lukasIO, line 273: this missing boolean was causing the issues in Safari

bcherry (approved):

davidzhao (approved):

PR #1562, opened 2025-06-20
Wait for participant in onTrack

lukasIO:


davidzhao:
I'm not certain that this is still needed.. the server should be handling this by sending down metadata first. are you seeing any issues?

PR #1569, opened 2025-06-29
allow manual controls even when adaptiveStream is enabled

davidzhao:
this would allow the user to manipulate video dimensions even when adaptive stream is in use.

it would allow us to override requested dims when running in low-power mode, but still be able to use adaptiveStream normally

### DIFF
--- a/src/room/track/RemoteTrackPublication.ts
+++ b/src/room/track/RemoteTrackPublication.ts
@@ -23,9 +23,11 @@ export default class RemoteTrackPublication extends TrackPublication {
 
   protected disabled: boolean = false;
 
-  protected currentVideoQuality?: VideoQuality = VideoQuality.HIGH;
+  protected videoDimensionsAdaptiveStream?: Track.Dimensions;
 
-  protected videoDimensions?: Track.Dimensions;
+  protected requestedVideoDimensions?: Track.Dimensions;
+
+  protected requestedVideoQuality?: VideoQuality;
 
   protected fps?: number;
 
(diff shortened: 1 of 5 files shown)


src/room/track/RemoteTrackPublication.ts
```diff
@@ -308,11 +310,16 @@ export default class RemoteTrackPublication extends TrackPublication {
       disabled: this.disabled,
       fps: this.fps,
     });
-    if (this.videoDimensions) {
-      settings.width = Math.ceil(this.videoDimensions.width);
-      settings.height = Math.ceil(this.videoDimensions.height);
-    } else if (this.currentVideoQuality !== undefined) {
-      settings.quality = this.currentVideoQuality;
+
+    // prefer explicit control if set
```
lukasIO: this approach effectively disables adaptive stream after quality has been set manually. 

I'm wondering whether setVideoQuality together with adaptiveStream should behave more like a `maxQuality` setting, so that adaptive stream could still allow to optimise for smaller tiles than requested?

src/room/track/RemoteTrackPublication.ts
```diff
@@ -297,25 +303,79 @@ export default class RemoteTrackPublication extends TrackPublication {
...
+            this.videoDimensionsAdaptiveStream.width * this.videoDimensionsAdaptiveStream.height <
+            minDimensions.width * minDimensions.height;
+          if (smallerAdaptive) {
+            this.log.debug('using adaptive stream dimensions instead of requested', {
+              ...this.logContext,
+              ...this.videoDimensionsAdaptiveStream,
+            });
+            minDimensions = this.videoDimensionsAdaptiveStream;
+          }
+        } else if (this.requestedMaxQuality !== undefined) {
```
davidzhao: nit: there's a bit of code duplication here. would be cleaner if layer / dimension matching was pulled out above

lukasIO (approved):

PR #1573, opened 2025-07-01
[FIX] Update active device immediately for muted video tracks

yashjain-99:
Fixes: [components-js/issues/1134 Muted video tracks issue](https://github.com/livekit/components-js/issues/1134#issuecomment-2877490017) 

- When changing video input devices while tracks are muted, the UI didn’t update because muted tracks don’t trigger automatic device restarts.
- Similar fix as done in PR: #1526


lukasIO (approved):
thanks!

PR #1575, opened 2025-07-01
Switch bundler from rollup to rolldown

lukasIO:
some reasons why I believe a switch makes sense:
- improves build speed significantly
- plugin interface is still rollup compatible
- built in lowering methods to support browser target syntax
- decreases bundle size a bit


1egoman (approved):
Generally makes sense to me! Two general thoughts:

1. I'm assuming you've tested this all locally and have high confidence everything is working. But I'm just wondering if there's any packaging related edge cases that require more extensive verification (ie, maybe one thing that comes to mind is verifying everything still works when imported into a commonjs / esm context properly still, etc).

2. It looks like the format of the output package has changed slightly (ie, `./dist/src` -> `./dist`, some files being named differently, etc). Is there value in matching the exact same paths / file names as the old rollup build? The reason I ask - users may have logic in their downstream apps which expects files to be at certain locations so changing file paths I think could be seen as a potential breaking change.

.github/workflows/release.yaml
```diff
@@ -2,6 +2,11 @@ name: Release
 
 on:
   workflow_dispatch:
+    inputs:
+      snapshot:
+        description: 'Publish a snapshot release'
+        type: boolean
+        default: false
```
1egoman, line 9: Just generally very excited to see this addition!

thought: Is it worthwhile to also add a text input here to allow configuration of the "personalized part" per [this](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md#versioning-your-packages)? ie, `0.0.0-foo-THE_TIME_YOU_DID_THIS` and publishing to tag `foo`. I really like doing this so that I can give a tester `livekit-client@foo` and every time I push a new version they can just pull down the latest.

lukasIO: I'd like to avoid publishing to too many different tags as it'll start to clutter the npmjs releases page. Do you think it's still worth while to add if it's _only_ for the snapshot prefix?

1egoman: I think it would still be worthwhile for just the snapshot prefix, because then at least searching through lists you could determine which tags were associated with what and it wouldn't just be opaque date strings. I've definitely done a bit of spelunking through snapshot builds to compare the result of different ones and it's annoying to try to correlate without that.

lukasIO: added in [4480ed6](https://github.com/livekit/client-sdk-js/pull/1575/commits/4480ed648f9cdea714bdacbe97f234adffe5e366)

PR #1590, opened 2025-07-07
Don't add dd extension for safari

cnderrauber:
When pubishing both camera and screenshare in
Safari 18.5, the browser could generate sdp with
dd extension in camera and without it in screenshare, munge sdp to add it could cause inconsistent ext id in two media sections and a warning in sfu side.

### DIFF
--- a/src/room/PCTransport.ts
+++ b/src/room/PCTransport.ts
@@ -6,7 +6,7 @@ import { debounce } from 'ts-debounce';
 import log, { LoggerNames, getLogger } from '../logger';
 import { NegotiationError, UnexpectedConnectionState } from './errors';
 import type { LoggerOptions } from './types';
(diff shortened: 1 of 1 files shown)


src/room/PCTransport.ts
```diff
@@ -316,7 +316,7 @@ export default class PCTransport extends EventEmitter {
               return true;
             }
 
-            if (isSVCCodec(trackbr.codec)) {
+            if (isSVCCodec(trackbr.codec) && !isSafari()) {
```
lukasIO, line 319: could this also occur on Webkit based browsers on e.g. iPads? 
Just wondering if we should do `isSafariBased` instead

cnderrauber: Don't have a device to confirm iOS but all warning logs in cloud are coming from mac. We can change it to safariBased later if iOS reports the error too.

PR #1595, opened 2025-07-08
Add e2ee for data channel messages

lukasIO:
depends on https://github.com/livekit/protocol/pull/1127

Most importantly for a frictionless adoption this PR ships _decryption_ support. 
This means if at a later stage encryption will be turned on by default, all clients that have the decryption support from this PR will continue to work.

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -983,7 +983,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
    * @internal for testing
    */
   async simulateScenario(scenario: SimulationScenario, arg?: any) {
-    let postAction = () => {};
+    let postAction = async () => {};
     let req: SimulateScenario | undefined;
     switch (scenario) {
       case 'signal-reconnect':
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## 2.15.7
+
+### Patch Changes
+
+- Ensure permission event is only emitted once for local participant - [#1643](https://github.com/livekit/client-sdk-js/pull/1643) ([@lukasIO](https://github.com/lukasIO))
+
+- internal typing fix - add missing async to postAction - [#1644](https://github.com/livekit/client-sdk-js/pull/1644) ([@1egoman](https://github.com/1egoman))
+
+- Ensure mid is always interpreted as string - [#1641](https://github.com/livekit/client-sdk-js/pull/1641) ([@lukasIO](https://github.com/lukasIO))
+
 ## 2.15.6
 
 ### Patch Changes
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "livekit-client",
-  "version": "2.15.6",
+  "version": "2.15.7",
   "description": "JavaScript/TypeScript client SDK for LiveKit",
   "main": "./dist/livekit-client.umd.js",
   "unpkg": "./dist/livekit-client.umd.js",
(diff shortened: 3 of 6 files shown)


src/room/Room.ts
```diff
@@ -355,11 +363,20 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
   private setupE2EE() {
-    if (this.options.e2ee) {
-      if ('e2eeManager' in this.options.e2ee) {
-        this.e2eeManager = this.options.e2ee.e2eeManager;
+    // when encryption is enabled via `options.encryption`, we enable data channel encryption
+
+    const dcEncryptionEnabled = false;
+    const e2eeOptions = this.options.e2ee;
+
+    // TODO(dc-e2ee): add this back in
```
bcherry: is this important in this PR?

bcherry (approved):
lgtm besides the todo.

src/room/Room.ts
lukasIO: this needs to be uncommented once we expose the `options.encryption` field. 

Because this PR doesn't actually declare that, we leave the logic commented out

src/e2ee/utils.ts
```diff
@@ -176,3 +177,18 @@ export function writeRbsp(data_in: Uint8Array): Uint8Array {
   }
   return new Uint8Array(dataOut);
 }
+
+export function asEncryptablePacket(packet: DataPacket): EncryptedPacketPayload | undefined {
+  if (
+    packet.value?.case !== 'sipDtmf' &&
+    packet.value?.case !== 'metrics' &&
```
davidliu, line 184: Looks like metrics is encryptable?

https://github.com/livekit/protocol/commit/72e862b1785637bba4347b7330954070b4de875b#diff-65ff73125c901a7593f1b81ac2dfa070f5d617c4fa3aebb570d36618526c3d89R313

lukasIO: In theory yes, the question is how that would handled for the dashboard were you'd want to see the metrics

src/e2ee/E2eeManager.ts
```diff
@@ -55,11 +68,26 @@ export class E2EEManager
...
   private keyProvider: BaseKeyProvider;
 
-  constructor(options: E2EEManagerOptions) {
+  private decryptDataRequests: Map<string, Future<DecryptDataResponseMessage['data']>> = new Map();
+
+  private encryptDataRequests: Map<string, Future<EncryptDataResponseMessage['data']>> = new Map();
+
+  private dataChannelEncryptionEnabled: boolean;
+
+  constructor(options: E2EEManagerOptions, dcEncryptionEnabled: boolean) {
```
pblazej, line 77: Do we really need to expose `dcEncryptionEnabled` here, or just assign based on the type of `options` (old vs new)?

src/room/Room.ts
pblazej: @lukasIO are you okay with creating the "new" options type right now and making it internal? I need that mostly for integration tests (to drive the feature), but maybe that's more elegant than the opt-in flag `dcEnabled`.

src/e2ee/E2eeManager.ts
lukasIO: in JS `E2EEManager` is an internal class, so this is not really a big concern to me. 
We can of course also make it part of the E2EEManagerOptions

src/room/Room.ts
lukasIO: for testing or in general? 

I've been thinking more about the "general" route and I think we should go forward with exposing a new options field now and deprecating the old one. v3 will remove the old one then.

src/e2ee/E2eeManager.ts
lukasIO: ah, I see why I didn't do this in the first place, as the `E2EEManagerOptions` are actually exposed to users under an alias right now.

src/room/Room.ts
pblazej: Yup, deprecating now makes much more sense, as e.g. you can deprecate old events/delegates as well, without changing the signatures "silently" so people no longer get their events (enriched with `EncryptionType`)

PR #1598, opened 2025-07-15
Add support for react-native preconnect audio

davidliu:


src/room/participant/LocalParticipant.ts
```diff
@@ -554,10 +554,14 @@ export default class LocalParticipant extends Participant {
...
 
         for (const localTrack of localTracks) {
+          const opts: TrackPublishOptions = {
+            ...this.roomOptions.publishDefaults,
+            ...options,
+          };
           if (
             source === Track.Source.Microphone &&
             isAudioTrack(localTrack) &&
-            publishOptions?.preConnectBuffer
```
davidliu, line 560: This didn't seem to include the roomOptions.publishDefaults, so changed to grab those for this check.

lukasIO (approved):
lgtm!

src/room/participant/LocalParticipant.ts
```diff
@@ -1331,7 +1336,7 @@ export default class LocalParticipant extends Participant {
             });
             const writer = await this.streamBytes({
               name: 'preconnect-buffer',
-              mimeType: 'audio/opus',
```
davidliu, line 1334: @lukasIO btw, this ends up changing from 'audio/opus' to 'audio/webm;codecs=opus' on JS, this won't have any problems will it?

lukasIO: thanks for flagging, we'll need to add this to the agent side, the agent expects audio/opus

lukasIO: cc @longcw

davidzhao: @davidliu I see you changed it to be compatible with python agents [here](https://github.com/livekit/client-sdk-js/pull/1598/files#diff-cf84874acd5677364a55762127f033eb5a9e7ca344a666b6305c3d76452bff02R1308).

should we have this override or to report the actual type? currently Python agents could handle both, even if the format isn't accurate. In case we want the latter, I have a PR [here](https://github.com/livekit/agents/pull/2952).

davidliu: I'll revert the change so that JS reports the actual type.

PR #1601, opened 2025-07-16
Ensure that if the signal layer is disconnected, signalling messages aren't sent in another place

1egoman:
Reproduction instructions within `agent-starter-react`:
1. Click "connect", and quickly (within a few hundred ms) click "end call"
2. A `cannot send signal request before connected, type: leave` error gets logged

What is happening:
1. `room.connect(...)` is called
2. Execution eventually gets to here, the first `await` after `disconnectLock` is unlocked: https://github.com/livekit/client-sdk-js/blob/8e787a29b3c0850ed49fbf365595eaf66cb370a2/src/room/Room.ts#L813
3. `engine.join(...)` is called, which is async, so the event loop gets taken back
4. an event fires, which the event loop picks up, which calls `room.disconnect()`
5. `room.disconnect()` expects the engine ws to be set up here: https://github.com/livekit/client-sdk-js/blob/8e787a29b3c0850ed49fbf365595eaf66cb370a2/src/api/SignalClient.ts#L672-L678
6. so the error gets logged

There seems to be a bunch of ad-hoc checks all over the place to drop messages if `SignalClient.isDisconnected`  is set, but the trouble is any promise `await` could result in other code resetting that state. So, really, it seems after every `await` this state has to be checked once again to verify that the room hasn't been disconnected in the intervening time. Maybe there's a more comprehensive solution here?

Another possible way to address this could be to extend the time the `disconnectLock` mutex is acquired on the connect end, but it sounds like talking with @lukasIO that the goal is to allow `room.disconnect()` to be run as quickly as possible during a `room.connect(...)`, so I opted to patch it on the disconnect end.

### DIFF
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -672,6 +672,7 @@ export class SignalClient {
     if (this.isDisconnected) {
       // Skip requests if the signal layer is disconnected
       // This can happen if an event is sent in the mist of room.connect() initializing
+      this.log.debug(`skipping signal request (type: ${message.case}) - SignalClient disconnected`);
       return;
     }
     if (!this.ws || this.ws.readyState !== this.ws.OPEN) {


src/api/SignalClient.ts
```diff
@@ -669,6 +669,11 @@ export class SignalClient {
     if (this.signalLatency) {
       await sleep(this.signalLatency);
     }
+    if (this.isDisconnected) {
+      // Skip requests if the signal layer is disconnected
+      // This can happen if an event is sent in the mist of room.connect() initializing
```
lukasIO, line 674: let's maybe add a debug log here just so that we have some insight into when this is happening?

1egoman: Good idea, added!

PR #1604, opened 2025-07-18
Skip track publish if the room is signalling client is disconnected

1egoman:
Previously, `setMicrophoneEnabled` would attempt to publish a given track, and would fail if `this.engine` wasn't properly initialized. When this occurred, a "cannot publish track when not connected" error toast would be shown in the application. 

This could happen via a scenario like the below:
1. `setMicrophoneEnabled` called
2. (a few ms go by)
3. `room.disconnect()` called, which starts tearing down the engine
4. [These lines](https://github.com/livekit/client-sdk-js/blob/b3ac2ace6a693ea04818406835feab195f551d91/src/room/participant/LocalParticipant.ts#L1164-L1166) are hit, the condition passes, and an error is thrown

So, explicitly check now that before calling `this.publish(...)` that the signalling client is connected and healthy, otherwise drop the message.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -891,31 +891,43 @@ export default class LocalParticipant extends Participant {
     }
     const publishPromise = new Promise<LocalTrackPublication>(async (resolve, reject) => {
       try {
-        if (this.engine.client.currentState !== SignalConnectionState.CONNECTED) {
-          this.log.debug('deferring track publication until signal is connected', {
-            ...this.logContext,
-            track: getLogContextFromTrack(track),
-          });
+        switch (this.engine.client.currentState) {
+          case SignalConnectionState.CONNECTING:
+          case SignalConnectionState.RECONNECTING:
+            this.log.debug('deferring track publication until signal is connected', {
+              ...this.logContext,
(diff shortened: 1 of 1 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -891,30 +891,45 @@ export default class LocalParticipant extends Participant {
...
+              const publication = await this.publish(track, opts, isStereo);
+              resolve(publication);
+            } catch (e) {
+              reject(e);
+            }
+            break;
+          }
+
+          case SignalConnectionState.DISCONNECTING:
+          case SignalConnectionState.DISCONNECTED: {
```
lukasIO: there's still a case here with `DISCONNECTED`, where the call to `setMicrophonEnabled` happens _before_ a room.connect call. Ideally we would still want to defer it in that case.

1egoman: @lukasIO Makes sense - I've made some updates and now opted to do this in a different way / at a different level of abstraction which I think doesn't suffer from that case you mentioned! See the pull request description for details.

lukasIO (approved):
nice, this makes sense to me!

PR #1611, opened 2025-07-29
Datastream - `TextStreamReader`/`ByteStreamReader` read timeouts and abort signals

1egoman:
Adds the ability to include an `AbortSignal` as a parameter when reading byte or text data from a datastream. This permits both manual cancellation (seems useful) but also importantly timeouts using `AbortSignal.timeout(...)`:

## Code overview
```typescript
room.registerTextStreamHandler('lk.chat', async (reader /* TextStreamReader */, participant) => {
  // Async iterator example
  for await (const chunk of reader.withAbortSignal(AbortSignal.timeout(5000))) {
    console.log('Received chunk:', chunk);
  }

  // OR:

  // `.readAll()` example
  const message = await reader.readAll({ signal: AbortSignal.timeout(5000) });
});
```

## VIdeo demo
Note that both of these are using a button-press to cancel the in flight read operation, but with `AbortSignal.timeout(...)` it could just as easily be driven by a timeout instead:

#### Aborting a chat message (uses `TextStreamReader`)
https://github.com/user-attachments/assets/47081155-971c-47a6-a0da-92119b0b9c9a

#### Aborting a file upload (uses `ByteStreamReader`)
https://github.com/user-attachments/assets/10bf79fc-4df2-4f42-ab7d-02d884fe7a1d

### DIFF
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -57,6 +57,8 @@ let currentRoom: Room | undefined;
 
 let startTime: number;
 
+let chatDataStreamReaderAbortController: AbortController | undefined;
+
 const searchParams = new URLSearchParams(window.location.search);
 const storedUrl = searchParams.get('url') ?? 'ws://localhost:7880';
 const storedToken = searchParams.get('token') ?? '';
@@ -259,32 +261,17 @@ const appActions = {
       });
 
     room.registerTextStreamHandler('lk.chat', async (reader, participant) => {
+      chatDataStreamReaderAbortController = new AbortController();
+      (<HTMLButtonElement>$('cancel-chat-receive-button')).style.display = 'block';
+
       const info = reader.info;
-      if (info.size) {
-        handleChatMessage(
-          {
-            id: info.id,
-            timestamp: info.timestamp,
-            message: await reader.readAll({ signal: AbortSignal.timeout(1000) }),
-          },
-          room.getParticipantByIdentity(participant?.identity),
-        );
-      } else {
-        let message = '';
-        try {
-          for await (const chunk of reader.withAbortSignal(AbortSignal.timeout(5000))) {
-            message += chunk;
-            handleChatMessage(
-              {
-                id: info.id,
-                timestamp: info.timestamp,
-                message,
-              },
-              room.getParticipantByIdentity(participant?.identity),
-            );
-          }
-        } catch (err) {
-          message += "ERROR";
+
+      let message = '';
+      try {
+        for await (const chunk of reader.withAbortSignal(
+          chatDataStreamReaderAbortController.signal,
+        )) {
+          message += chunk;
           handleChatMessage(
             {
               id: info.id,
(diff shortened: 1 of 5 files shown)


examples/demo/demo.ts
```diff
@@ -259,30 +261,44 @@ const appActions = {
...
-      if (info.size) {
-        handleChatMessage(
-          {
-            id: info.id,
-            timestamp: info.timestamp,
-            message: await reader.readAll(),
-          },
-          room.getParticipantByIdentity(participant?.identity),
-        );
-      } else {
```
1egoman, line 272: Just want to make sure there's not something I'm missing - the `handleChatMessage` calls in both sides of this `if` seem like they do the same thing? Other than the `appendLog` call at the bottom of the `else` case, am I right this is a no-op and can be simplified like I did here?

src/room/StreamReader.ts
```diff
@@ -40,35 +45,77 @@ export class ByteStreamReader extends BaseStreamReader<ByteStreamInfo> {
...
+   * Injects an AborrSignal, which if aborted, will terminate the currently active
+   * stream iteration operation.
+   *
+   * Note that when using AbortSignal.timeout(...), the timeout applies across
+   * the whole iteration operation, not just one individual chunk read.
+   */
+  withAbortSignal(signal: AbortSignal) {
+    this.signal = signal;
+    return this;
+  }
```
1egoman: Note that this mechanism here injects this `AbortSignal` in a way that it gets used in the next read operation no matter what it is. I think this provides a pretty slick interface but the mutating behavior could maybe be a bit surprising.

I considered here somehow doing a deep clone of `this` and then assigning `signal`, but doing that deep clone proved to be challenging since there's a lot of state in this class - maybe it's a better long term path, though. 🤔

src/room/StreamReader.ts
```diff
@@ -137,22 +211,37 @@ export class TextStreamReader extends BaseStreamReader<TextStreamInfo> {
...
             };
           }
-        } catch (error) {
-          // TODO handle errors
-          return { done: true, value: undefined };
+        } catch (err) {
+          cleanup();
+          throw err;
         }
       },
```
1egoman, line 218: There was a subtle bug / resource leak in the previous implementation here. If `reader` were to throw an error, `reader.releaseLock()` wouldn't be called and the reader would be left in a permanently open state, blocking other consumers.

src/room/StreamReader.ts
```diff
@@ -40,35 +45,77 @@ export class ByteStreamReader extends BaseStreamReader<ByteStreamInfo> {
...
-        reader.releaseLock();
+        cleanup();
         return { done: true, value: undefined };
       },
     };
   }
 
-  async readAll(): Promise<Array<Uint8Array>> {
+  /**
+   * Injects an AborrSignal, which if aborted, will terminate the currently active
```
lukasIO:
```suggestion
   * Injects an AbortSignal, which if aborted, will terminate the currently active
```

src/room/StreamReader.ts
```diff
@@ -137,22 +211,37 @@ export class TextStreamReader extends BaseStreamReader<TextStreamInfo> {
...
-        reader.releaseLock();
+        cleanup();
         return { done: true, value: undefined };
       },
     };
   }
 
-  async readAll(): Promise<string> {
+  /**
+   * Injects an AborrSignal, which if aborted, will terminate the currently active
```
lukasIO:
```suggestion
   * Injects an AbortSignal, which if aborted, will terminate the currently active
```

PR #1613, opened 2025-07-31
Handle more data stream errors

1egoman:
After looking at a few other sdks worked, especially the swift sdk, I noticed [there were some data stream-related error cases](https://github.com/livekit/client-sdk-swift/blob/f37bbd260d61e165084962db822c79f995f1a113/Sources/LiveKit/DataStream/StreamError.swift) that this sdk didn't properly handle. Most of these are invariants that shouldn't ever be true in practice and largely only will happen when packets are malformed on the sender end.

So, try to handle these error cases and throw errors to make invalid data fail in a predictable way. A list of error cases:
- If the remote participant disconnects halfway through sending a data stream message, make any currently active `DataStreamReader.readAll` / `DataStreamReader` async iteration operations reject with `new DataStreamError("...", DataStreamErrorReason.AbnormalEnd)`.
- If a client sends text chunks containing data which can not be decoded as a valid utf-8 string, throw a `new  DataStreamError("...", DataStreamError.DecodeFailed)`
- If a client sends a text/byte header containing a `totalSize` parameter and after reading all chunks, the byte lengths all summed up don't equal the expected `totalSize`, then either throw `new DataStreamError("...", DataStreamErrorReason.LengthIncomplete)` if too small, or throw `new DataStreamError("...", DataStreamErrorReason.LengthExceeded)` if too large.
- If `room.registerTextStreamHandler(...)` / `room.registerByteStreamHandler(...)` is called for a topic which already has a handler registered, throw `new DataStreamError("...", DataStreamErrorReason.HandlerAlreadyRegistered)`

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -61,7 +61,7 @@ import {
   roomOptionDefaults,
   videoDefaults,
 } from './defaults';
-import { ConnectionError, ConnectionErrorReason, UnsupportedServer } from './errors';
+import { ConnectionError, ConnectionErrorReason, DataStreamError, DataStreamErrorReason, UnsupportedServer } from './errors';
 import { EngineEvent, ParticipantEvent, RoomEvent, TrackEvent } from './events';
 import LocalParticipant from './participant/LocalParticipant';
 import type Participant from './participant/Participant';
@@ -289,7 +289,10 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
 
   registerTextStreamHandler(topic: string, callback: TextStreamHandler) {
     if (this.textStreamHandlers.has(topic)) {
-      throw new TypeError(`A text stream handler for topic "${topic}" has already been set.`);
+      throw new DataStreamError(
+        `A text stream handler for topic "${topic}" has already been set.`,
+        DataStreamErrorReason.HandlerAlreadyRegistered,
+      );
     }
     this.textStreamHandlers.set(topic, callback);
   }
@@ -300,7 +303,10 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
 
   registerByteStreamHandler(topic: string, callback: ByteStreamHandler) {
     if (this.byteStreamHandlers.has(topic)) {
-      throw new TypeError(`A byte stream handler for topic "${topic}" has already been set.`);
+      throw new DataStreamError(
+        `A byte stream handler for topic "${topic}" has already been set.`,
+        DataStreamErrorReason.HandlerAlreadyRegistered,
+      );
     }
     this.byteStreamHandlers.set(topic, callback);
   }
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -1847,6 +1847,14 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       const stream = new ReadableStream({
         start: (controller) => {
           streamController = controller;
+
+          if (this.textStreamControllers.has(streamHeader.streamId)) {
+            throw new DataStreamError(
+              `A data stream read is already in progress for a stream with id ${streamHeader.streamId}.`,
+              DataStreamErrorReason.AlreadyOpened,
+            );
+          }
+
           this.byteStreamControllers.set(streamHeader.streamId, {
             info,
             controller: streamController,
@@ -1884,6 +1892,14 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       const stream = new ReadableStream<DataStream_Chunk>({
         start: (controller) => {
           streamController = controller;
+
+          if (this.textStreamControllers.has(streamHeader.streamId)) {
+            throw new DataStreamError(
+              `A data stream read is already in progress for a stream with id ${streamHeader.streamId}.`,
+              DataStreamErrorReason.AlreadyOpened,
+            );
+          }
+
           this.textStreamControllers.set(streamHeader.streamId, {
             info,
             controller: streamController,
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -61,7 +61,13 @@ import {
   roomOptionDefaults,
   videoDefaults,
 } from './defaults';
-import { ConnectionError, ConnectionErrorReason, DataStreamError, DataStreamErrorReason, UnsupportedServer } from './errors';
+import {
+  ConnectionError,
+  ConnectionErrorReason,
+  DataStreamError,
+  DataStreamErrorReason,
+  UnsupportedServer,
+} from './errors';
 import { EngineEvent, ParticipantEvent, RoomEvent, TrackEvent } from './events';
 import LocalParticipant from './participant/LocalParticipant';
 import type Participant from './participant/Participant';
@@ -1649,18 +1655,20 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       return;
     }
 
+    // Terminate any in flight data stream receives from the given participant
     const streamsBeingSentByDisconnectingParticipant = [
       ...Array.from(this.textStreamControllers.values()),
       ...Array.from(this.byteStreamControllers.values()),
-    ].filter(controller => controller.sendingParticipantIdentity === identity)
+    ].filter((controller) => controller.sendingParticipantIdentity === identity);
     if (streamsBeingSentByDisconnectingParticipant.length > 0) {
-      const topics = streamsBeingSentByDisconnectingParticipant.map(
-        controller => controller.info.topic
-      );
-      throw new DataStreamError(
-        `Participant ${identity} disconnected in the middle of sending data on these topics: ${topics.join(', ')}`,
-        DataStreamErrorReason.AbnormalEnd,
-      );
+      for (const controller of streamsBeingSentByDisconnectingParticipant) {
+        controller.outOfBandFailureRejectingFuture.reject?.(
+          new DataStreamError(
+            `Participant ${identity} unexpectedly disconnected in the middle of sending data`,
+            DataStreamErrorReason.AbnormalEnd,
+          ),
+        );
+      }
     }
 
     participant.trackPublications.forEach((publication) => {
@@ -1826,7 +1834,6 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
   private async handleStreamHeader(streamHeader: DataStream_Header, participantIdentity: string) {
     if (streamHeader.contentHeader.case === 'byteHeader') {
       const streamHandlerCallback = this.byteStreamHandlers.get(streamHeader.topic);
-
       if (!streamHandlerCallback) {
         this.log.debug(
           'ignoring incoming byte stream due to no handler for topic',
@@ -1834,7 +1841,10 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
         );
         return;
       }
+
       let streamController: ReadableStreamDefaultController<DataStream_Chunk>;
+      const outOfBandFailureRejectingFuture = new Future<never>();
+
       const info: ByteStreamInfo = {
         id: streamHeader.streamId,
         name: streamHeader.contentHeader.value.name ?? 'unknown',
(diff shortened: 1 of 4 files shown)


src/room/Room.ts
```diff
@@ -1643,6 +1655,22 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
+    if (streamsBeingSentByDisconnectingParticipant.length > 0) {
+      for (const controller of streamsBeingSentByDisconnectingParticipant) {
+        controller.outOfBandFailureRejectingFuture.reject?.(
+          new DataStreamError(
+            `Participant ${identity} unexpectedly disconnected in the middle of sending data`,
+            DataStreamErrorReason.AbnormalEnd,
+          ),
+        );
+      }
+    }
```
1egoman: **New Error Case:** If the remote participant disconnects halfway through sending a data stream message, make any currently active `DataStreamReader.readAll` / `DataStreamReader` async iteration operations reject with `new DataStreamError("...", DataStreamErrorReason.AbnormalEnd)`.

*Note this is technically backwards incompatible with the current implementation, though this case is probably uncommon so I'm not considering it that big of a deal.*

Screenshot 2025-07-30 at 4 54 29 PM

src/room/StreamReader.ts
```diff
@@ -199,16 +240,33 @@ export class TextStreamReader extends BaseStreamReader<TextStreamInfo> {
...
+            let decodedResult;
+            try {
+              decodedResult = decoder.decode(value.content);
+            } catch (err) {
+              throw new DataStreamError(
+                `Cannot decode datastream chunk ${value.chunkIndex} as text: ${err}`,
+                DataStreamErrorReason.DecodeFailed,
+              );
+            }
+
```
1egoman: **New Error Case:** If a client sends text chunks containing data which can not be decoded as a valid utf-8 string, throw a `new  DataStreamError("...", DataStreamError.DecodeFailed)`

Note that previously, the `TextDecoder` wasn't being constructed with `fatal: true` [(more info)](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/fatal), which meant that it would skip bad bytes / try to do the best it could with any input specified.

Because of that, this is technically backwards incompatible - but also, I'd think that if folks are sending malformed payloads, it would be better for it to explicitly break then implicitly work like it did previously?

src/room/StreamReader.ts
```diff
@@ -16,15 +17,42 @@ abstract class BaseStreamReader<T extends BaseStreamInfo> {
...
+        `Not enough chunk(s) received - expected ${this.totalByteSize} bytes of data total, only received ${this.bytesReceived} bytes`,
+        DataStreamErrorReason.LengthIncomplete,
+      );
+    } else if (this.bytesReceived > this.totalByteSize) {
+      throw new DataStreamError(
+        `Extra chunk(s) received - expected ${this.totalByteSize} bytes of data total, received ${this.bytesReceived} bytes`,
+        DataStreamErrorReason.LengthExceeded,
+      );
+    }
+  }
```
1egoman: **New Error Case:** If a client sends a text/byte header containing a `totalSize` parameter and after reading all chunks, the byte lengths all summed up don't equal the expected `totalSize`, then either throw `new DataStreamError("...", DataStreamErrorReason.LengthIncomplete)` if too small, or throw `new DataStreamError("...", DataStreamErrorReason.LengthExceeded)` if too large.

Note that this is technically backwards incompatible, though again it seems like it would be better for this to explicitly throw to uncover malformed data then let these issues go unnoticed.

Screenshot 2025-07-31 at 9 24 33 AM
Screenshot 2025-07-31 at 9 25 13 AM

src/room/Room.ts
```diff
@@ -289,7 +295,10 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
   registerTextStreamHandler(topic: string, callback: TextStreamHandler) {
     if (this.textStreamHandlers.has(topic)) {
-      throw new TypeError(`A text stream handler for topic "${topic}" has already been set.`);
+      throw new DataStreamError(
+        `A text stream handler for topic "${topic}" has already been set.`,
+        DataStreamErrorReason.HandlerAlreadyRegistered,
+      );
     }
     this.textStreamHandlers.set(topic, callback);
   }
```
1egoman: **Updated Error Case**: If `room.registerTextStreamHandler(...)` / `room.registerByteStreamHandler(...)` is called for a topic which already has a handler registered, throw `new DataStreamError("...", DataStreamErrorReason.HandlerAlreadyRegistered)`

An error was already thrown here, but now it is using the new error format.

src/room/Room.ts
```diff
@@ -1827,30 +1857,46 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
         start: (controller) => {
           streamController = controller;
+
+          if (this.textStreamControllers.has(streamHeader.streamId)) {
+            throw new DataStreamError(
+              `A data stream read is already in progress for a stream with id ${streamHeader.streamId}.`,
+              DataStreamErrorReason.AlreadyOpened,
+            );
+          }
+
```
1egoman: **New Error Case:** If a user tries to read from a text/byte data stream that is already being read from somewhere else, explicitly throw `new DataStreamError("...", DataStreamErrorReason.AlreadyOpened)`.

Note this is technically backwards incompatible, though previously if you tried to do this, your pre-existing data stream reader would stop getting data because the `this.byteStreamControllers` / `this.textStreamControllers` map entry would get overriden. So, I think throwing an explicit error is probably much better than the previous implicit / broken state that would occur.

src/room/errors.ts
```diff
@@ -111,6 +111,39 @@ export class SignalRequestError extends LivekitError {
...
+  reasonName: string;
+
+  constructor(message: string, reason: DataStreamErrorReason) {
+    super(16, message);
+    this.name = 'DataStreamError';
+    this.reason = reason;
+    this.reasonName = DataStreamErrorReason[reason];
+  }
+}
+
```
1egoman, line 147: It wasn't clear to me exactly how this `LivekitError` class worked when it comes to error codes. Each error in this file seems to have its own code (ie, `super(12, message);`, `super(13, message);`, etc) but there are some discontinuities?

I just used the next one in the series (`16`) - where do these come from? Did I do the right thing here? Do these error codes map to some other list, maybe in the docs somewhere?

src/room/Room.ts
```diff
@@ -1643,6 +1655,22 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
     }
 
+    // Terminate any in flight data stream receives from the given participant
+    const streamsBeingSentByDisconnectingParticipant = [
+      ...Array.from(this.textStreamControllers.values()),
+      ...Array.from(this.byteStreamControllers.values()),
+    ].filter((controller) => controller.sendingParticipantIdentity === identity);
+    if (streamsBeingSentByDisconnectingParticipant.length > 0) {
+      for (const controller of streamsBeingSentByDisconnectingParticipant) {
+        controller.outOfBandFailureRejectingFuture.reject?.(
```
lukasIO: it would be good to also clean up everything else in the controller map on a fatal error of a specific stream

src/room/StreamReader.ts
lukasIO: Is this considering the case with `totalByteSize` being `0` ? 
I think we should ignore totalByteSize if it's set to `0` in addition to when it's undefined

src/room/StreamReader.ts
```diff
@@ -37,6 +65,8 @@ abstract class BaseStreamReader<T extends BaseStreamInfo> {
 export class ByteStreamReader extends BaseStreamReader<ByteStreamInfo> {
   protected handleChunkReceived(chunk: DataStream_Chunk) {
     this.bytesReceived += chunk.content.byteLength;
+    this.validateBytesReceived();
```
lukasIO: how would you expect a user to catch this? 
Looks like this would always be thrown as an uncaught exception right now?

lukasIO: how would a user catch this? 
Looks like this would always be thrown as an uncaught exception right now?

src/room/StreamReader.ts
```diff
@@ -151,7 +189,10 @@ export class TextStreamReader extends BaseStreamReader<TextStreamInfo> {
       return;
     }
     this.receivedChunks.set(index, chunk);
+
     this.bytesReceived += chunk.content.byteLength;
+    this.validateBytesReceived();
```
lukasIO: same question as above about the uncaught exception here

src/room/StreamReader.ts
lukasIO: I'm not yet convinced that switching to `fatal` mode is desirable here. 
Imagine a stream with a bunch of valid data and one chunk gets – for whatever reason – corrupted but the stream would continue with valid data afterwards.

src/room/errors.ts
```diff
@@ -111,6 +111,39 @@ export class SignalRequestError extends LivekitError {
...
+  AbnormalEnd,
+
+  // Incoming chunk data could not be decoded.
+  DecodeFailed,
+
+  // Read length exceeded total length specified in stream header.
+  LengthExceeded,
+
+  // Read length less than total length specified in stream header.
+  LengthIncomplete,
```
lukasIO: if we borrow the names from Swift (which I think makes sense for consistency), then let's not change some of them?

src/room/errors.ts
lukasIO: I think the discontinuities were an effort to group certain types of errors, e.g. connection related errors starting with `1`. 

It's not done particularly well though. I think we could either give DataStreamError it's own new first digit or keep `16` from your proposal. I don't have strong opinion on this. 

I'm tagging @ladvoc as he came up with the Swift data stream errors and also wanted to explore unifying error messages across SDKs. 
If there was a clear path already for unified error codes + messages then we might as well adopt that now for new errors gradually

lukasIO (commented):
this will be great to have. Just some minor notes and thoughts

src/room/StreamReader.ts
1egoman: @lukasIO I don't have a strong opinion, but lean towards including it. A few notes:
- My understanding is other sdks throw an error in this case as well, so it might be nice to standardize behavior
- In regards to the scenario you outlined: while the behavior you describe could maybe be desired, I could also see it being a source of surprising bugs - imagine you sent a large text document split over multiple packets, and when it was reassembled on the other end, a chunk was missing in the middle, and downstream user code after that then operated on the invalid input without realizing it was invalid in some way. I would think that it would be much better to get an explicit error which could be caught that would let a downstream user choose how to handle this case (retrying, dropping data, etc)

If there are truly cases where both behaviors are desired, it could maybe be an additional configuration parameter? Though then maybe the conversation would be more about which value should be the default...

src/room/StreamReader.ts
1egoman: @lukasIO I hadn't considered this was a case where a user would want to catch this - it seems like if the `totalByteSize` specified disagreed with the actual length of the message in bytes, the resolution would always be fixing the `totalByteSize` value sent on the sender end, and once that was done this error would go away here (receiver end).

Can you describe a case where catching this would be valuable?

That being said, maybe there is some value in all errors being raised by a `StreamReader` presenting themselves as promise rejections returned from the reader, so maybe that is rationale enough.

1egoman: See https://github.com/livekit/client-sdk-js/pull/1613#discussion_r2251529441

src/room/StreamReader.ts
1egoman: See https://github.com/livekit/client-sdk-js/pull/1613#discussion_r2251529441

src/room/StreamReader.ts
lukasIO: thanks, agree that those are some solid arguments in favor of fatal mode

src/room/StreamReader.ts
1egoman: Actually, looking at this in more detail - `handleChunkReceived` is always called as part of the iterator `next`, so I think the errors would bubble up via the `readAll` / async iteration as would be expected?

src/room/errors.ts
1egoman: Ah oops, yea makes sense.

@lukasIO Also worth noting also here, the enum value order (and therefore, their corresponding assigned values) don't line up with the swift sdk since some of those errors aren't relevant in the js sdk and were omitted from the enum. Given this, maybe there's an argument to explicitly assign to numerical values from the swift sdk, if those are to be the standard?

src/room/StreamReader.ts
1egoman: Discussed on a call, [my above comment ](https://github.com/livekit/client-sdk-js/pull/1613#discussion_r2251569871) seems to be correct, so lukas is good with this approach

src/room/Room.ts
1egoman: Good point - assuming I understand what you are asking correctly, I think this commit addresses the concern: https://github.com/livekit/client-sdk-js/pull/1613/commits/b27c325e723811cabfbc2e9943195473f93fd59f.

src/room/errors.ts
1egoman: Done [here](https://github.com/livekit/client-sdk-js/pull/1613/commits/8f7bc06ebf3104a20e9cd8159f843f3267b6a2de), and I also explicitly assigned each enum value to be equal to its swift counterpart. Maybe that's too heavy handed, idk, but the more I think about it the more I think if they are going to line up name wise, they should also line up value wise.

PR #1614, opened 2025-07-31
Reabstract data stream logic into new `IncomingDataStreamManager` / `OutgoingDataStreamManager`

1egoman:
Looking at a few other livekit sdk implementations, a few others have abstracted their data stream logic into two classes or structs often called `IncomingDataStreamManager` (which handles all datastream packet parsing logic, calling `registerTextStreamHandler` functions, etc), and `OutgoingDataStreamManager` (which contains implementations for sending data over the `RTCPeerConnection` like `streamBytes`, `sendText`, etc).

Here's the relevant files in the swift sdk ([incoming](https://github.com/livekit/client-sdk-swift/blob/f37bbd260d61e165084962db822c79f995f1a113/Sources/LiveKit/DataStream/Incoming/IncomingStreamManager.swift), [outgoing](https://github.com/livekit/client-sdk-swift/blob/f37bbd260d61e165084962db822c79f995f1a113/Sources/LiveKit/DataStream/Incoming/OutgoingStreamManager.swift)), android sdk ([incoming](https://github.com/livekit/client-sdk-android/blob/caa474d284b29b5b3dd8fb16f3424dbdf50949a3/livekit-android-sdk/src/main/java/io/livekit/android/room/datastream/incoming/IncomingDataStreamManager.kt#L90), [outgoing](https://github.com/livekit/client-sdk-android/blob/caa474d284b29b5b3dd8fb16f3424dbdf50949a3/livekit-android-sdk/src/main/java/io/livekit/android/room/datastream/outgoing/OutgoingDataStreamManager.kt#L56)), and the rust sdk ([incoming](https://github.com/livekit/rust-sdks/blob/d2fb275fc3aea09212b80036db8b41b396647731/livekit/src/room/data_stream/incoming.rs#L232), [outgoing](https://github.com/livekit/rust-sdks/blob/d2fb275fc3aea09212b80036db8b41b396647731/livekit/src/room/data_stream/outgoing.rs#L277-L280)).

I think this could be a good idea for the javascript sdk to do as well, as it removes complexity from the large / hard to reason about `Room` class. If this seems like a good direction, I could see there being some other relatively self contained subsystems that this could be done with to try to make things smaller and easier to understand.

### DIFF
--- a/src/room/data-stream/incoming/IncomingDataStreamManager.ts
+++ b/src/room/data-stream/incoming/IncomingDataStreamManager.ts
@@ -4,19 +4,15 @@ import {
   DataStream_Header,
   DataStream_Trailer,
 } from '@livekit/protocol';
-import log, { /* LoggerNames, getLogger */ } from '../../logger';
+import log from '../../../logger';
+import { type ByteStreamInfo, type StreamController, type TextStreamInfo } from '../../types';
+import { bigIntToNumber } from '../../utils';
 import {
   type ByteStreamHandler,
   ByteStreamReader,
   type TextStreamHandler,
   TextStreamReader,
-} from '../StreamReader';
-import {
-  type ByteStreamInfo,
-  type StreamController,
-  type TextStreamInfo,
-} from '../types';
-import { bigIntToNumber } from '../utils';
+} from './StreamReader';
 
 export default class IncomingDataStreamManager {
   private log = log;
--- a/src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
+++ b/src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
@@ -9,10 +9,9 @@ import {
   DataStream_TextHeader,
   DataStream_Trailer,
 } from '@livekit/protocol';
-import log, { type StructuredLogger } from '../../logger';
-import type RTCEngine from '../RTCEngine';
-import { ByteStreamWriter, TextStreamWriter } from '../StreamWriter';
-import { EngineEvent } from '../events';
+import log, { type StructuredLogger } from '../../../logger';
+import type RTCEngine from '../../RTCEngine';
+import { EngineEvent } from '../../events';
 import type {
   ByteStreamInfo,
   SendFileOptions,
@@ -20,8 +19,9 @@ import type {
   StreamBytesOptions,
   StreamTextOptions,
   TextStreamInfo,
-} from '../types';
-import { numberToBigInt, splitUtf8 } from '../utils';
+} from '../../types';
+import { numberToBigInt, splitUtf8 } from '../../utils';
+import { ByteStreamWriter, TextStreamWriter } from './StreamWriter';
 
 const STREAM_CHUNK_SIZE = 15_000;
 
--- a/src/room/types.ts
+++ b/src/room/types.ts
@@ -43,12 +43,15 @@ export type StreamBytesOptions = {
   streamId?: string;
   mimeType?: string;
   totalSize?: number;
-}
+};
 
-export type SendFileOptions = Pick<StreamBytesOptions, 'topic' | 'mimeType' | 'destinationIdentities'> & {
+export type SendFileOptions = Pick<
+  StreamBytesOptions,
+  'topic' | 'mimeType' | 'destinationIdentities'
+> & {
   onProgress?: (progress: number) => void;
   encryptionType?: Encryption_Type.NONE;
-}
+};
 
 export type DataPublishOptions = {
   /**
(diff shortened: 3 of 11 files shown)


src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -0,0 +1,172 @@
...
+import { type ByteStreamInfo, type StreamController, type TextStreamInfo } from '../../types';
+import { bigIntToNumber } from '../../utils';
+import {
+  type ByteStreamHandler,
+  ByteStreamReader,
+  type TextStreamHandler,
+  TextStreamReader,
+} from './StreamReader';
+
+export default class IncomingDataStreamManager {
```
lukasIO, line 17: what are your thoughts on when this should be disposed/cleared?

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -0,0 +1,315 @@
...
+const STREAM_CHUNK_SIZE = 15_000;
+
+/**
+ * Manages sending custom user data via data channels.
+ * @internal
+ */
+export default class OutgoingDataStreamManager {
+  protected engine: RTCEngine;
+
+  protected log: StructuredLogger = log;
```
lukasIO: with this method I believe we will use the room's logging context. 
Can we instead pass the room's parent logger into the constructor together with the engine?

src/room/types.ts
```diff
@@ -35,6 +35,24 @@ export interface StreamTextOptions {
...
+  mimeType?: string;
+  totalSize?: number;
+};
+
+export type SendFileOptions = Pick<
+  StreamBytesOptions,
+  'topic' | 'mimeType' | 'destinationIdentities'
+> & {
+  onProgress?: (progress: number) => void;
+  encryptionType?: Encryption_Type.NONE;
```
lukasIO, line 53: as long as we don't actually support encrypting the files, let's not expose this as an option?

1egoman: @lukasIO I can remove it, but [it is there in the current released SDK](https://github.com/livekit/client-sdk-js/blob/8332ccacaca794c0cefe5a85596be79343c1c729/src/room/participant/LocalParticipant.ts#L1904), so I think it'd technically be a (small) breaking change.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: @lukasIO I had assumed that the lifecycle of both `IncomingDataStreamManager` and `OutgoingDataStreamManager` would roughly line up with the `Room` they are associated with. As far as I can tell, there aren't any resources being managed that need to be explicitly deinitialized, so relying on the garbage collector seemed good enough - is there something you see that I'm missing?

(If so, I can add some sort of manual destructor that the `Room` could call.)

src/room/types.ts
lukasIO: Sorry, I missed that, yeah, let's leave it in then!

src/room/data-stream/incoming/IncomingDataStreamManager.ts
lukasIO: The main thing I'm wondering about is whether we should clear it on room disconnection 🤔

1egoman: @lukasIO Ah, I see - fwiw, the controllers / handlers state doesn't seem to get cleared in the existing `Room` version of this logic after disconnecting.

That seems like it might be a mistake / bug though - can you confirm it is? If so I can clear these after room disconnection in this new implementation.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
1egoman: I think [this](https://github.com/livekit/client-sdk-js/pull/1614/commits/cbb331400095a9a2826066ad35af04f18fd0c53d) is what you are surfacing here - did I get this right? If not, I might need some clarification.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
lukasIO: Yeah, I think it would definitely be an improvement to add cleanup on disconnect

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
lukasIO: that's it, thanks!

lukasIO (approved):

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: Cool, done: https://github.com/livekit/client-sdk-js/pull/1614/commits/45980728cf3ca1f5f5d0ad4c994ec9af876a0309

PR #1618, opened 2025-08-04
feat(e2ee): add h265 as supported codec for encryption

lukasIO:


src/e2ee/worker/naluUtils.ts
lukasIO, line 1: extracted NAL related parsing to a util file to keep FrameCryptor a bit cleaner

cloudwebrtc (approved):
the code looks pretty good!

PR #1619, opened 2025-08-04
Add RTCEngine`SupersededBy` event

1egoman:
## Problem / Concern
`RTCEngine` gets fully town down and recreated initially when an initial connection is created and also whenever a reconnection occurs. This means that any class that has a reference to `engine` must be informed of this new change so that it can keep its reference to `engine` current.

The current means this is being done is by calling a `setupEngine` method on all dependents - right now, `LocalParticipant` and `E2EEManager`, plus soon `OutgoingDataStreamManager`. This is relatively coupled though because the room needs to call that method on each of these classes!

## Potential solution
The existing approach _works_, but is inelegant. To attempt to solve this in a slightly more elegant way, I've added a new `SupersededBy` event to the `RTCEngine` which is fired on the old engine whenever a new engine is generated, which allows anybody who has the engine to always have the latest copy.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -94,8 +94,10 @@ enum PCState {
   Closed,
 }
 
+const EngineEventEmitter = EventEmitter as new () => TypedEventEmitter<EngineEventCallbacks>;
+
 /** @internal */
-export default class RTCEngine extends (EventEmitter as new () => TypedEventEmitter<EngineEventCallbacks>) {
+export default class RTCEngine extends EngineEventEmitter {
   client: SignalClient;
 
   rtcConfig: RTCConfiguration = {};
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -1580,6 +1580,7 @@ export type EngineEventCallbacks = {
   offline: () => void;
   signalRequestResponse: (response: RequestResponse) => void;
   signalConnected: (joinResp: JoinResponse) => void;
+  supersededBy: (newEngine: RTCEngine) => void;
 };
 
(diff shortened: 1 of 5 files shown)


src/room/Room.ts
```diff
@@ -610,12 +611,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
       });
 
-    if (this.localParticipant) {
-      this.localParticipant.setupEngine(this.engine);
-    }
-    if (this.e2eeManager) {
-      this.e2eeManager.setupEngine(this.engine);
-    }
+    oldEngine?.emit(EngineEvent.SupersededBy, this.engine);
   }
```
1egoman, line 615: Previously, `n` function calls were required. Now, only one publish does the same thing!

src/room/RTCEngine.ts
```diff
@@ -94,8 +94,10 @@ enum PCState {
   Closed,
 }
 
+const EngineEventEmitter = EventEmitter as new () => TypedEventEmitter<EngineEventCallbacks>;
+
 /** @internal */
-export default class RTCEngine extends (EventEmitter as new () => TypedEventEmitter<EngineEventCallbacks>) {
+export default class RTCEngine extends EngineEventEmitter {
```
1egoman, line 100: Admittedly this is a little weird - `EngineEventCallbacks` now contains a reference to `RTCEngine`, so putting this expression directly within the `extends` creates a reference cycle that typescript doesn't seem to like. Breaking it up like this seems to solve that problem, though.

lukasIO (approved):
nice!

1egoman:
I'm going to close this, I think it's a good idea but I don't have the bandwidth to fix the conflicts

PR #1620, opened 2025-08-08
lock all APIs that change underlying track with the same lock

lukasIO:
for local tracks we currently have both a processor lock and a restart lock

We can run into situations where the processor loses track of the actual underlying mediastreamtrack if the track is being restarted while the processor is setting up.

In order to avoid this race, this PR removes the processor lock and instead uses the restart lock also in processor related method calls. 
The biggest risk with this change is that there could be unforeseen situations in which two restartlock acquirations block each other. I did some testing to try to force this, but wasn't able to and also cannot spot anything that would allow this in the code, but asking for reviews to take a close look on whether you can spot anything that might lead to a deadlock with this change.

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -57,7 +57,7 @@ export default abstract class LocalTrack<
 
   protected processor?: TrackProcessor<TrackKind, any>;
 
-  protected processorLock: Mutex;
+  // protected processorLock: Mutex;
 
   protected audioContext?: AudioContext;
 
@@ -65,7 +65,7 @@ export default abstract class LocalTrack<
 
   protected localTrackRecorder: LocalTrackRecorder<typeof this> | undefined;
 
-  private restartLock: Mutex;
+  protected restartLock: Mutex;
 
   /**
    *
@@ -86,9 +86,14 @@ export default abstract class LocalTrack<
     this.providedByUser = userProvidedTrack;
     this.muteLock = new Mutex();
     this.pauseUpstreamLock = new Mutex();
-    this.processorLock = new Mutex();
     this.restartLock = new Mutex();
-    this.setMediaStreamTrack(mediaTrack, true);
+    this.restartLock.lock().then(async (unlock) => {
+      try {
+        await this.setMediaStreamTrack(mediaTrack, true);
+      } finally {
+        unlock();
+      }
+    });
 
     // added to satisfy TS compiler, constraints are synced with MediaStreamTrack
     this._constraints = mediaTrack.getConstraints();
@@ -171,27 +176,22 @@ export default abstract class LocalTrack<
     }
     let processedTrack: MediaStreamTrack | undefined;
     if (this.processor && newTrack) {
-      const unlock = await this.processorLock.lock();
-      try {
-        this.log.debug('restarting processor', this.logContext);
-        if (this.kind === 'unknown') {
-          throw TypeError('cannot set processor on track of unknown kind');
-        }
-
-        if (this.processorElement) {
-          attachToElement(newTrack, this.processorElement);
-          // ensure the processorElement itself stays muted
-          this.processorElement.muted = true;
-        }
-        await this.processor.restart({
-          track: newTrack,
-          kind: this.kind,
-          element: this.processorElement,
-        });
-        processedTrack = this.processor.processedTrack;
-      } finally {
-        unlock();
+      this.log.debug('restarting processor', this.logContext);
+      if (this.kind === 'unknown') {
+        throw TypeError('cannot set processor on track of unknown kind');
       }
+
+      if (this.processorElement) {
+        attachToElement(newTrack, this.processorElement);
+        // ensure the processorElement itself stays muted
+        this.processorElement.muted = true;
+      }
+      await this.processor.restart({
+        track: newTrack,
+        kind: this.kind,
+        element: this.processorElement,
+      });
+      processedTrack = this.processor.processedTrack;
     }
     if (this.sender && this.sender.transport?.state !== 'closed') {
       await this.sender.replaceTrack(processedTrack ?? newTrack);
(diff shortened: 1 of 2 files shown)


src/room/track/LocalTrack.ts
```diff
@@ -171,27 +176,22 @@ export default abstract class LocalTrack<
     }
     let processedTrack: MediaStreamTrack | undefined;
     if (this.processor && newTrack) {
-      const unlock = await this.processorLock.lock();
```
lukasIO, line 174: locking here isn't needed anymore as every method invoking `setMediaStreamTrack` is locking now in the caller

src/room/track/LocalTrack.ts
```diff
@@ -57,15 +57,15 @@ export default abstract class LocalTrack<
...
+  // protected processorLock: Mutex;
 
   protected audioContext?: AudioContext;
 
   protected manuallyStopped: boolean = false;
 
   protected localTrackRecorder: LocalTrackRecorder<typeof this> | undefined;
 
-  private restartLock: Mutex;
+  protected restartLock: Mutex;
```
1egoman: Nitpick but maybe it makes sense to rename this to something more generic if it now is being used for both functions?

src/room/track/LocalTrack.ts
```diff
@@ -57,15 +57,15 @@ export default abstract class LocalTrack<
 
   protected processor?: TrackProcessor<TrackKind, any>;
 
-  protected processorLock: Mutex;
+  // protected processorLock: Mutex;
```
1egoman: If there's only one mutex in play now, IMO it makes sense to outright get rid of the old one.

```suggestion
```

src/room/track/LocalTrack.ts
```diff
@@ -290,36 +290,42 @@ export default abstract class LocalTrack<
...
+      this.log.debug('replace MediaStreamTrack', this.logContext);
+      await this.setMediaStreamTrack(track);
+      // this must be synced *after* setting mediaStreamTrack above, since it relies
+      // on the previous state in order to cleanup
 
-    if (stopProcessor && this.processor) {
-      await this.stopProcessor();
+      if (stopProcessor && this.processor) {
+        await this.stopProcessor();
+      }
```
1egoman: I think there's a potential deadlock situation here - `this.stopProcessor` within has a `this.restartLock.lock()` call but `replaceTrack` has already taken out a lock on `this.restartLock` at this point.

Maybe this `if` block could be moved underneath the `try` / `finally`?


One more example of a similar type of situation: https://github.com/livekit/client-sdk-js/blob/664b21ad7844dac0e1bdaebfdeafd87baec7c89f/src/room/track/LocalTrack.ts#L542-L544

1egoman (commented):
General thought: I wonder if there's a good way to somehow make it clear through naming of private methods that they do / do not take out a lock on a mutex. Or maybe even leaving a note in a docs comment or something... I don't have a concrete recommendation but I think something could maybe be done there.

src/room/track/LocalTrack.ts
lukasIO: good catch! 

yeah, this looks problematic. 
Even if moving it around for the first instance would work, I don't see a good way of mitigating it by moving for the second problematic situation you found :/

we could have a `dangerouslySkipLock` param on the stopProcessor, but that's not exactly neat.

src/room/track/LocalTrack.ts
lukasIO: forgot to push the commit that did this 🤦

src/room/track/LocalTrack.ts
lukasIO: sure, I'll just throw out the first things that come to my head (don't have any great ideas)
- `trackLock`
- `acquisitionLock`
- `trackChangeLock`

anything particular you had in mind?

1egoman: Nothing particular, any of those three sound fine!

src/room/track/LocalTrack.ts
1egoman: Maybe (naming aside) there could be a public / private half of `stopProcessor`, like:

```tsx
  async stopProcessor(...) {
    const unlock = this.mutex.lock();
    try {
      await this.internalStopProcessor(...);
    } finally {
      unlock();
    }
  }

  private async internalStopProcessor(...) {
    // Existing stopProcessor code goes here
  }
```

And then things that needed to call `stopProcessor` within a pre-existing mutex could call `internalStopProcessor` instead? But like, if this were the path that were explored, naming is going to be key to ensure this isn't crazy confusing and that maybe ties into my previous comment.

lukasIO: I like the suggestion! 

I'm taking your thumbs up on the decorator-mutex-documentation to mean you'd be on board with that. Additionally documenting it directly on the internalStopProcessor method what the expectations around the caller are.  
Does that sound good enough as a confusion limitation strategy to you?

1egoman: Yep I think that could make sense! Two minor things that come to mind:
  - Would a decorator / higher order function show up in any sort of docs ingestion tooling in any way? ie, language server tools, [built artifacts like this](https://docs.livekit.io/reference/client-sdk-js/modules.html), etc.
  - Not all mutex locks would necessarily wrap a function. That's probably the common case and in this situation it holds true, but I'm thinking of cases like `room.connect()` where the same mutex is locked / unlocked multiple times

FWIW, I'm less worried about this specific case and more just in general given all the async behavior at play in this package. So if you want to start with something lightweight / continue to discuss it, and one of us do a separate pass applying whatever makes sense everywhere I'd be fine with that approach too!

src/room/track/LocalTrack.ts
lukasIO: changed it to `trackChangeLock`

src/room/track/LocalTrack.ts
lukasIO: yeah makes sense. 

I left out the decorators part for now (here's the [PR](https://github.com/livekit/client-sdk-js/pull/1621)) because additionally to the concerns you mentioned I'm not actually sure this would be compatible with react native.

Let me know if the documentation part sounds ok to you or whether I should add anything additionally

PR #1627, opened 2025-08-14
Check for encrypted track if room unencrypted, and if so, emit an event

1egoman:
If a remote participant joins a room and sends an encrypted track, but the local participant / room is not configured to support encryption, throw an error.

Screenshot 2025-08-14 at 5 02 57 PM

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -1409,13 +1409,17 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       }
     }
 
-    participant.addSubscribedMediaTrack(
+    const publication = participant.addSubscribedMediaTrack(
       mediaTrack,
       trackId,
       stream,
       receiver,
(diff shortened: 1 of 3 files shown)


src/room/Room.ts
```diff
@@ -1409,13 +1409,20 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
       trackId,
       stream,
       receiver,
       adaptiveStreamSettings,
     );
+
+    if (publication?.isEncrypted && !this.isE2EEEnabled) {
+      this.emit(
+        RoomEvent.EncryptionError,
+        new Error('Encrypted track received, but room does not have encryption enabled!'),
```
lukasIO: maybe adding the participant identity and the `source` of the publication in the error message would be helpful

1egoman: Good idea, done:

Screenshot 2025-08-15 at 9 45 38 AM

PR #1632, opened 2025-08-20
fix(e2ee): h264 publishing with e2ee enabled

lukasIO:
we got reports around h264 encrypted tracks publishing only a single layer and also not showing up on the remote side. 

The fix for this is twofold:

- the compare function for the `e2eeSimulcastSupport` was checking the wrong way around
- we now pass codec information to the frame cryptor on track published so that we don't have to rely on the nalu parsing logic for figuring out whether or not the codec is h264 or h265 on Safari


bcherry (approved):

fkwp:
@lukasIO just tested this one and it still looks like if the iPhone iOS client is first client publishing e2ee webcam track it for what ever reason is claiming to use vp8... 

though I can see `workerLogger.info('updated codec', vp8, h264)` trying to update the codec to h264...

What is the easiest way for me to repo with meet.livekit.io (or a local version of it)

PR #1633, opened 2025-08-26
Fix local package links breaking on build

lukasIO:
previously the rollup build would delete contents of `dist` on every build, which breaks local linking of packages.


1egoman (approved):

1egoman:
Ah oops approved slightly prematurely, it looks like there's a related error in one of the github action tasks that is keeping ci from passing:

> Command "build:clean" not found. Did you mean "pnpm run build:watch"?
> Error: The process '/home/runner/setup-pnpm/node_modules/.bin/pnpm' failed with exit code 1

PR #1635, opened 2025-08-27
Add Safari speaker selection support

### LINKED ISSUE #1568: Add support for iOS26 Speaker Selection API
### Describe the problem

Finally Apple is supporting "the Speaker Selection API" on iOS and iPadOS as pointed out here: https://developer.apple.com/documentation/safari-release-notes/safari-26-release-notes#WebRTC

After testing it successfully using https://webrtc.github.io/samples/src/content/devices/input-output/ I can confirm switching the output device between all available output devices works including: Earpiece Speaker, Main Speaker and connected headsets like wired or bluetooth headsets.

Using https://meet.livekit.io I can see the different output devices in the settings. However, selecting them does not change the actual output device.

Removing iOS specific workarounds for iOS >= 26

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

fkwp:
## Support iOS 26 Speaker Selection API

Apple added `HTMLMediaElement.setSinkId()` in iOS/iPadOS 26 and Safari 26. However, calling it directly on `<audio>` elements backed by WebRTC remote tracks is silently ignored. It seems iOS renders that audio through the WebRTC engine's internal pipeline (→ AVAudioSession), bypassing the AVPlayer pipeline that `setSinkId` controls.

This PR adds a shared WebAudio relay element per AudioContext: all remote tracks mix into one `MediaStreamAudioDestinationNode`, played via a single `<audio>` element on which `setSinkId` is called once during the user gesture in `switchActiveDevice`. iOS grants the permission once for the lifetime of that element, so participants joining later route correctly without their own user gesture.

  RTCPeerConnection remote track A ─┐
  RTCPeerConnection remote track B ─┼─→ shared MediaStreamDestinationNode ─→ shared  ← setSinkId() here
  RTCPeerConnection remote track C ─┘
          (each via AudioContext + GainNode)

### Changes
- `isSafariSpeakerSelectionSupported()` — detects iOS/Safari 26+
- `supportsSetSinkId()` — returns `true` for iOS 26+ (skips feature-detection that misfires before DOM attachment)
- `getOrCreateSharedRelay()` / `disposeSharedRelay()` — lazy AudioContext-scoped relay setup
- `RemoteAudioTrack.connectWebAudio` routes the gain node to the shared destination when `AudioContext.setSinkId` is unavailable
- `Room.switchActiveDevice` calls `setSinkId` on the shared relay for the iOS 26 path; per-participant `setAudioOutput` path retained for Chrome (echo cancellation) and other
browsers
- `selectDefaultDevices` / `handleDeviceChange` no longer block audiooutput auto-switching on iOS 26
- `supportsSetSinkId` exported alongside `isSafariSpeakerSelectionSupported`

fixes https://github.com/livekit/client-sdk-js/issues/1568

My testing stack is https://github.com/livekit-examples/meet as reference app and meanwhile the following stack is working on my end:
- this PR
- https://github.com/livekit/components-js/pull/1328
- https://github.com/livekit-examples/meet/pull/529


src/room/utils.ts
```diff
@@ -196,6 +196,19 @@ export function isSafariSvcApi(browser?: BrowserDetails): boolean {
...
+export function isSafariSpeakerSelectionSupported(browser?: BrowserDetails): boolean {
+  if (!browser) {
+    browser = getBrowser();
+  }
+  // Safari (macOS or iOS) since version 26
+  // https://developer.apple.com/documentation/safari-release-notes/safari-26-release-notes#WebRTC
+  return (
+    (browser?.name === 'Safari' && compareVersions(browser.version, '26') >= 0) ||
+    (browser?.os === 'iOS' && !!browser?.osVersion && compareVersions(browser.osVersion, '26') >= 0) ||
+    (browser?.os === 'macOS' && !!browser?.osVersion && compareVersions(browser.osVersion, '26') >= 0)
```
lukasIO: this check is not connected to checking against Safari, I believe?

lukasIO (commented):
Thanks for the PR! 

I believe  the speaker selection API comes with some more caveats (explicit user permission must be granted for output selection if no microphone is acquired yet). 

I haven't tested this, but probably needs its own selection path via `MediaDevices.selectAudioOutput` ?

1egoman:
@fkwp If it's helpful, a few months back I spent a bit of time attempting to further adapt your changes to get ios speaker selection to work. I wasn't able to get it all the way there (and the effort has since been deprioritized), but maybe some of this code could be useful in your effort here: https://github.com/livekit/client-sdk-js/tree/ios-speaker-selection

fkwp:
> @fkwp If it's helpful, a months back I spent a bit of time attempting to further adapt your changes to get ios speaker selection to work. I wasn't able to get it all the way there (and the effort has since been deprioritized), but maybe some of this code could be useful in your effort here: https://github.com/livekit/client-sdk-js/tree/ios-speaker-selection

@1egoman nice Ill definitely will look into this. I used https://github.com/livekit-examples/meet as reference app and meanwhile the following stack is working on my end (given you also publish audio):
- this PR
- https://github.com/livekit/components-js/pull/1328
- https://github.com/livekit-examples/meet/pull/529

1egoman:
Per comment left in #1836, some more notes from when I attempted to dig into this on my own, in case they would be helpful to anybody else attempting to introduce speaker selection support:

-----------

I was able to get the api working on my test iphone 11, but was NOT able to get the api to work when I integrated it within `livekit-client`.

Importantly, if I add a button into the demo app which on tap reaches into the DOM, gets the active audio element playing remote participant audio, and manually calls setSinkId on it, this works and works every single time without fail. So the APIs definitely exist on my test iphone.

A few things that might be part of the problem with integrating into livekit-client:

- setSinkId has a requirement in that it must be triggered by a user gesture. Right now the setSinkId implementation in livekit-client isn't always (at least based off of my minimal testing) triggered in a user gesture path in the demo. I made some changes in [my WIP branch](https://github.com/livekit/client-sdk-js/tree/ios-speaker-selection) (added an "apply" button next to the output device selector select element as onChange doesn't count as a user gesture on ios apparently) and was I think able to fix this but it needs some more thorough testing.

- I have a suspicion there's some sort of race condition going on. I can do the same exact sequence of events (ie, connect, switch from "default speaker" -> "receiver" -> "speaker") and get different results each time. There's something going on that makes this non deterministic and I (or a LLM) couldn't figure it out on my own.

PR #1638, opened 2025-08-29
Use WebSocketStream for sequential signal processing

lukasIO:
This is part of a couple of PRs that work towards signalling handling improvements.

As a first step this PR only swaps out `WebSocket` with a `WebSocketStream` which is based on https://developer.mozilla.org/en-US/docs/Web/API/WebSocketStream.

In follow ups I'd like to move closer to a signalling api layer with a swappable transport that holds references to `ReadableStream<SignalResponse>` and `WritableStream<SignalRequest>`.

### DIFF
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -44,9 +44,9 @@ import log, { LoggerNames, getLogger } from '../logger';
 import { ConnectionError, ConnectionErrorReason } from '../room/errors';
 import CriticalTimers from '../room/timers';
 import type { LoggerOptions } from '../room/types';
-import { getClientInfo, isReactNative, sleep } from '../room/utils';
+import { Future, getClientInfo, isReactNative, sleep } from '../room/utils';
 import { AsyncQueue } from '../utils/AsyncQueue';
-import { WebSocketStream } from './WebSocketStream';
+import { type WebSocketConnection, WebSocketStream } from './WebSocketStream';
 import { createRtcUrl, createValidateUrl, parseSignalResponse } from './utils';
 
 // internal options
@@ -258,7 +258,7 @@ export class SignalClient {
     return res;
   }
 
-  private connect(
+  private async connect(
     url: string,
     token: string,
     opts: ConnectOpts,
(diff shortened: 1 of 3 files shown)


src/api/SignalClient.ts
```diff
@@ -267,189 +270,201 @@ export class SignalClient {
...
+      const handleSignalConnected = (
+        connection: WebSocketConnection,
+        firstMessage?: SignalResponse,
+      ) => {
+        this.state = SignalConnectionState.CONNECTED;
+        clearTimeout(wsTimeout);
+        abortSignal?.removeEventListener('abort', abortHandler);
+        this.startPingInterval();
+        this.startReadingLoop(connection.readable.getReader(), firstMessage);
+        console.info('setting stream writer');
```
1egoman: Should this be using `log`, and not `console`?

src/api/SignalClient.ts
```diff
@@ -267,189 +270,201 @@ export class SignalClient {
...
+        this.streamWriter = connection.writable.getWriter();
+
+        this.ws!.closed.then((closeInfo) => {
```
1egoman: I'd think the `!` should be a `?`, since `this.ws` isn't set on the first time `connect` runs? But since this is right in the middle of the happy path / I'm assuming you've tested this and it works, maybe there's something I'm not seeing here...

src/api/SignalClient.ts
```diff
@@ -267,189 +270,201 @@ export class SignalClient {
...
+              } else if (reason instanceof ConnectionError) {
+                throw reason;
               } else {
-                reject(
-                  new ConnectionError(
-                    `Encountered unknown websocket error during connection: ${ev.toString()}`,
-                    ConnectionErrorReason.InternalError,
-                    resp.status,
-                  ),
+                console.error(reason);
```
1egoman: Another instance where I suspect `console` should be some form of `log`?

src/api/SignalClient.ts
```diff
@@ -475,28 +490,16 @@ export class SignalClient {
...
-
-        if (this.ws.readyState < this.ws.CLOSING) {
-          this.ws.close();
-          // 250ms grace period for ws to close gracefully
-          await Promise.race([closePromise, sleep(250)]);
-        }
+        const closePromise = this.ws.closed;
         this.ws = undefined;
+        this.streamWriter = undefined;
+        await Promise.race([closePromise, sleep(250)]);
```
1egoman: I'm curious, why the 250ms timeout? Could make that delay in ms some sort of constant with a descriptive name to demystify it a bit?

src/room/defaults.ts
```diff
@@ -48,5 +48,5 @@ export const roomConnectOptionDefaults: InternalRoomConnectOptions = {
   autoSubscribe: true,
   maxRetries: 1,
   peerConnectionTimeout: 15_000,
-  websocketTimeout: 15_000,
```
1egoman, line 51: Just to confirm, this was purposeful and not something done locally for testing + accidentally committed, correct? If this was purposeful, I'm curious the background on why it was decreased so substantially given I'd think behind the scenes all the same network characteristics are present.

src/api/SignalClient.ts
lukasIO: the method is only called after `this.ws` is set. For this handler we expect the `ws` to always be present

src/api/SignalClient.ts
lukasIO: ugh, thanks, these were only meant as debug helpers, removing all console logs

src/api/SignalClient.ts
lukasIO: It's a rather arbitrary number. Just to make sure the WS had some time to close successfully but give an upper bound. 
Agree, this is nicer when named explicitly

src/room/defaults.ts
lukasIO: thanks for catching that!

src/api/SignalClient.ts
```diff
@@ -94,6 +95,9 @@ export enum SignalConnectionState {
   DISCONNECTED,
 }
 
+/** specifies how much time (in ms) we allow for the ws to close its connection gracefully before continuing */
+const MAX_WS_CLOSE_TIME = 250;
```
1egoman, line 99: Nitpick: I know it's in the comment above, but it would be good to have a unit somewhere in this variable name as well.

PR #1645, opened 2025-09-09
Add `TokenSource` token fetching abstraction

1egoman:
### Background
Previously, when connecting to a room, this library required passing an explicit `url` / `token` - `room.connect(url, token)`. However, this has some downsides:
- Just one literal set of credentials is being passed in, so when performing reconnects those credentials could be expired.
- There are some opportunities for some potential performance improvements  - in particular, credential caching (ie, use the existing credentials as long as their are valid, and get new ones once they are expired) and token refreshing on room disconnect to ensure fresh credentials are present are two that come to mind.

This change introduces a new concept ~`ConnectionCredentials`~ `TokenSource` ~which is now accepted by `room.connect` and `room.prepareConnection` (importantly, in a backwards compatible way!), and addresses both points~ which initially operates in isolation of `Room`, but will be more integrated with `Room` in a coming change.

There are multiple implementations (`TokenSourceLiteral` / `TokenSourceCustom` / `TokenSourceSandboxTokenServer`) which adapt to different token generation mechanisms as well, which should mean that for most simple cases users can use one of these options. If something more fancy is desired, there are a few abstract base classes (`TokenSourceFixed` and `TokenSourceConfigurable`) which can be extended based on desired behavior to allow users to create their own fully custom credential fetching strategies.

### Usage example
```typescript
// Literal matches the current behavior today:
const tokenSource = TokenSource.literal({
  serverUrl: "wss://foo.livekit.cloud",
  participantToken: "xxxxxxxxxxxxx",
});

// NOTE: this seems kinda pointless but this is phase one - `Literal`s will become more important once you can pass a `TokenSource` directly into `Room`
const response = await credentials.fetch();
await room.connect(response.serverUrl, response.participantToken);

// Literal can also be used in an async fashion:
const credentials = TokenSource.literal(async () => ({
  serverUrl: "wss://foo.livekit.cloud",
  participantToken: "xxxxxxxxxxxxx",
}));

const response = await credentials.fetch();
await room.connect(response.serverUrl, response.participantToken);

// Custom allows for async token generation - credentials are cached as well, so the passed function is
// only called when the old token expires:
const credentials = new TokenSource.Custom((request) => {
  // Wait some time
  await new Promise((resolve) => setTimeout(resolve, 2000));

  return {
    serverUrl: "wss://foo.livekit.cloud",
    participantToken: "xxxxxx request data encoded in here xxxxxxx",
  };
});

const response = await credentials.fetch();
await room.connect(response.serverUrl, response.participantToken);

// SandboxTokenServer uses a sandbox token server:
const credentials = TokenSource.sandboxTokenServer("sandbox id here");
const response = await credentials.fetch();
await room.connect(response.serverUrl, response.participantToken);
```

### DIFF
--- a/src/room/token-source/TokenSource.ts
+++ b/src/room/token-source/TokenSource.ts
@@ -279,7 +279,7 @@ export const TokenSource = {
    * For more info:
    * @see https://cloud.livekit.io/projects/p_/sandbox/templates/token-server
    */
-  sandboxTokenServer(sandboxId: string, options: SandboxTokenServerOptions) {
+  sandboxTokenServer(sandboxId: string, options: SandboxTokenServerOptions = {}) {
     return new TokenSourceSandboxTokenServer(sandboxId, options);
   },
 };


src/room/ConnectionCredentials.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  /** The name of the room to join. If omitted, a random new room name will be generated instead. */
+  roomName?: string;
+
+  /** The identity of the participant the token should connect as connect as. If omitted, a random
+   * identity will be used instead. */
+  participantName?: string;
+  participantToken: string;
+};
+
```
1egoman: Note that I think before merging this, it would be worth having a discussion and ensuring this object schema is something that we're willing to commit to across all sdks for this use case. 

FWIW, this is the same interface as the sandbox token server api response body, so my suspicion is that it's probably a decent / already familiar standard. But I'm not sure.

src/room/ConnectionCredentials.ts
```diff
@@ -0,0 +1,146 @@
...
+      super();
+      this.payload = payload;
+    }
+
+    async fetch() {
+      if (this.isCachedConnectionDetailsExpired()) {
+        // FIXME: figure out a better logging solution?
+        console.warn(
+          'The credentials within ConnectionCredentials.Literal have expired, so any upcoming uses of them will likely fail.',
+        );
```
1egoman: I have a few `console.warn`s in this `ConnectionCredentials` logic - I could swap them out with some sort of logger implementation, but both are things that I would think should be loudly shown to the user even if the user has all logs turned off, so I'm not sure. I see other instances of `console.warn` in the app so it's not clear to me if this is a good or bad pattern.

src/room/ConnectionCredentials.ts
```diff
@@ -0,0 +1,146 @@
...
+      super();
+      this.options = options;
+
+      if (process.env.NODE_ENV === 'production' && !this.options.disableSecurityWarning) {
+        // FIXME: figure out a better logging solution?
+        console.warn(
+          'ConnectionCredentials.SandboxTokenServer is meant for development, and is not security hardened. In production, implement your own token generation solution.',
+        );
+      }
+    }
```
1egoman: I didn't see any other cases in this project where `process.env.NODE_ENV` was being checked.

It's not clear to me if this is a bad pattern or not - I really want to show this warning only in an end application production build and this seemed like the most direct way to accomplish that. Maybe there's some other way I could check this if what I'm doing here isn't preferred?

src/room/Room.ts
```diff
@@ -575,7 +578,24 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
    */
-  async prepareConnection(url: string, token?: string) {
+  prepareConnection(connectionCredentials: ConnectionCredentials): Promise<void>;
+  prepareConnection(url: string): Promise<void>;
+  /** @deprecated Use room.prepareConnection(connectionCredentials) instead */
+  prepareConnection(url: string, token: string): Promise<void>;
+  async prepareConnection(
+    urlOrConnectionCredentials: ConnectionCredentials | string,
+    tokenOrUnknown?: string,
+  ) {
```
1egoman: I want to call attention to what I did here for the args to `prepareConnection` - there are now a few overloads breaking out all the options. The old signature still works perfectly fine via that overload, and I also broke out the `url` only case into its own overload.

So now, these overloads are what users should use:
```typescript
room.prepareConnection(connectionCredentials);
room.prepareConnection(url); // (Required because `connectionCredentials` cannot be created without a token)
```

And the old one, while it still works, is deprecated:
```typescript
room.prepareConnection(url, token); // DEPRECATED
```

src/room/Room.ts
```diff
@@ -599,7 +619,32 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     }
   }
 
-  connect = async (url: string, token: string, opts?: RoomConnectOptions): Promise<void> => {
+  connect: {
+    (connectionCredentials: ConnectionCredentials, opts?: RoomConnectOptions): Promise<void>;
+    /** @deprecated Use room.connect(connectionCredentials, opts?: RoomConnectOptions) instead */
+    (url: string, token: string, opts?: RoomConnectOptions): Promise<void>;
+  } = async (urlOrConnectionCredentials, tokenOrOpts, optsOrUnset?: unknown): Promise<void> => {
+    let opts: RoomConnectOptions = {};
```
1egoman: I want to call attention to what I did here for the args to `connect` - there are now two overloads, one for the "new" version, and one for the "old" version. Note that the old / preexisting version still works perfectly fine.

So now, this overload is what users should use:
```typescript
room.connect(connectionCredentials, /* optional opts */);
```

And the old one, while it still works, is deprecated:
```typescript
room.connect(url, token, /* optional opts */); // DEPRECATED
```

src/room/ConnectionCredentials.ts
lukasIO: what's the use case for participantName and roomName being returned from the connection details endpoint? 

Should there be an agentName returned?

src/room/ConnectionCredentials.ts
lukasIO: the RPC related console.warns shouldn't be there. 

Generally trying to keep all LiveKit related logs using the logger implementation

src/room/ConnectionCredentials.ts
```diff
@@ -0,0 +1,149 @@
...
+  };
+
+  /** ConnectionCredentials.SandboxTokenServer queries a sandbox token server for credentials,
+   * which supports quick prototyping / getting started types of use cases.
+   *
+   * This token provider is INSECURE and should NOT be used in production.
+   *
+   * For more info:
+   * @see https://cloud.livekit.io/projects/p_/sandbox/templates/token-server */
+  export class SandboxTokenServer extends ConnectionCredentials {
```
lukasIO: are we sure we really want to support sandbox credentials retrieval outside of sandboxes?

src/room/Room.ts
lukasIO: nice

src/room/ConnectionCredentials.ts
lukasIO: ah, I think this currently conflates both connectionDetails request and response?

src/room/ConnectionCredentials.ts
1egoman: Just to make sure I understand - are you referring to the `baseUrl` key in `SandboxTokenServerOptions`? If so, the thought there was that it would give a generic way to point this at staging environments for LiveKit staff / etc but that it wouldn't really be an end user used option. If you can think of a better way to expose the ability to do that (or, maybe this using staging use case isn't as important as I'd think it would be), let me know.

src/room/ConnectionCredentials.ts
1egoman: This `ConnectionDetails` data structure is the "response" using that nomenclature, and is the exact response coming back from the sandbox token server currently - see the below screenshot taken from the sandbox token server root page docs:

Screenshot 2025-09-10 at 9 33 33 AM

I don't have an explicit "request" data structure as the existing `Room` isn't built to need this, but I think in theory it'd look something like the sandbox request body. So something like:
```typescript
type ConnectionDetailsRequest = {
  roomName?: string;
  participantName?: string;
  roomConfig: /* some more complicated type here */,
};

type ConnectionDetailsResponse = {
  serverUrl: string;
  roomName?: string;
  participantName?: string;
  participantToken: string;
};
```

> what's the use case for participantName and roomName being returned from the connection details endpoint?

The existing sandbox endpoint was returning them - I'm guessing it is so if the names are autogenerated at token generation time, a client could get access to their values later on? But I'm not sure.

> Should there be an agentName returned?

I opted to ignore the complexities of agent dispatch at least initially / just port what I had since I figured you and others would have thoughts. I think to handle this though I'd prefer to include `roomConfig` in the `ConnectionDetailsRequest`, and then decode the room config data from the token and make that accessible for agent sdk abstractions to query (ie, via some sort of method on a `ConnectionCredentials` instance).

A few advantages to that approach:
- It's more generic than using agent specific parameters which make no sense outside of the agent use case
- It looks like the room configuration value already has inherent support for multiple agents so we won't have to figure out how to get that in here later
- This works well with `ConnectionCredentials.Literal` - there's no "request" object in that context, so the `roomConfig` value from the token being decoded / made available downstream back transparently handles that case too.

1egoman: Ok, did the above. Example of what this ends up looking like:

### Custom token generation example
```typescript
const connectionCredentials = new ConnectionCredentials.Custom(async (request) => {
  // Somehow use the metadata in `request` and generate a token
  // Note that `request.roomConfig.agents[0].agentName` is `"example"`

  return { url: "...", participantToken: "token encoding request data here" };
});


// Everything above is end user facing
// --------------------------
// Everything below would be within the agent sdk internals

connectionCredentials.setRequest({
  roomConfig: new RoomConfiguration(
    agents: [new RoomAgentDispatch({ agentName: "example", metadata: "" })],
  ),
});

room.connect(connectionCredentials);

// Gets the roomconfig value decoded from the `participantToken` - this is how the agent sdk
// would know which agent(s) were dispatched
const roomConfigFromToken = connectionCredentials.getCachedResponseRoomConfig();
console.log(token.agents[0].agentName); // "example"
```

### Literal token generation example
```typescript
const connectionCredentials = new ConnectionCredentials.Literal({
  // Note there is no "request" in this context, since the credentials have already been generated
  url: "...",
  participantToken: "token with roomconfig.agents[0].agentName = example set in payload",
});

// Everything above is end user facing
// --------------------------
// Everything below would be within the agent sdk internals

// (The below could be included but is effectively a no-op since with connectioncredentials.literal, credentials
// have already been generated)
// connectionCredentials.setRequest({
//   roomConfig: new RoomConfiguration(
//     agents: [new RoomAgentDispatch({ agentName: "example", metadata: "" })],
//   ),
// });

room.connect(connectionCredentials);

// Gets the roomconfig value decoded from the `participantToken` - this is how the agent sdk would know what was dispatched
const roomConfigFromToken = connectionCredentials.getCachedResponseRoomConfig();
console.log(token.agents[0].agentName); // "example"
```

lukasIO: > I'm guessing it is so if the names are autogenerated at token generation time, a client could get access to their values later on? But I'm not sure.

the name for both room and local participant are also available once signal connected. I'm not against including it in the response, but was wondering if there is a use case for having it available before connection

1egoman: Per our discussion in slack: Updated the code + above example to store the request on `connectionCredentials` with a method rather than pass it in to `connect` / `prepareConnection`.

src/room/ConnectionCredentials.ts
1egoman: Discussed in slack - lukas thinks what I have makes sense here, other than the `SandboxTokenServer` name, which is a bit confusing, but it's what users know this functionality as currently so it probably makes sense to stick with it.

src/room/ConnectionCredentials.ts
```diff
@@ -0,0 +1,245 @@
...
+import { decodeJwt } from 'jose';
+import log, { LoggerNames, getLogger } from '../logger';
+
+const ONE_SECOND_IN_MILLISECONDS = 1000;
+const ONE_MINUTE_IN_MILLISECONDS = 60 * ONE_SECOND_IN_MILLISECONDS;
+
+/**
+ * ConnectionCredentials handles getting credentials for connecting to a new Room, caching
+ * the last result and using it until it expires. */
+export abstract class ConnectionCredentials {
```
lukasIO: not sure why it's currently not already causing a conflict that both the abstract class and the namespace are exported under the same name?

src/room/ConnectionCredentials.ts
```diff
@@ -0,0 +1,245 @@
...
+     * @see https://docs.livekit.io/home/get-started/authentication/#room-configuration
+     */
+    roomConfig?: RoomConfiguration;
+  };
+  export type Response = {
+    serverUrl: string;
+    participantToken: string;
+  };
+
+  export type LiteralOptions = { loggerName?: string };
```
lukasIO: I think we can just hardcode it without an option to customise?

src/room/ConnectionCredentials.ts
1egoman: It's a built in typescript feature called declaration merging: https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces-with-classes.

I could move the `ConnectionCredentials` class to be within the namespace and rename it `Base` or something (ie, then externally it would be known as `ConnectionCredentials.Base`) but that reads a little less well IMO in type definitions `function takesAnyCredentialsSubtype(credentials: ConnectionCredentials) {}` vs `function takesAnyCredentialsSubtype(credentials: ConnectionCredentials.Base) {}`. I don't really have a strong preference either way but lean towards what I have currently.

src/room/ConnectionCredentials.ts
```diff
@@ -0,0 +1,245 @@
...
+        );
+      }
+      return this.cachedResponse!;
+    }
+  }
+
+  /** ConnectionCredentials.Custom allows a user to define a manual function which generates new
+   * {@link Response} values on demand. Use this to get credentials from custom backends / etc.
+   * */
+  export class Custom extends ConnectionCredentials {
```
lukasIO: wondering if we should allow request to be user extensible here? 

```suggestion
  export class Custom<T extends Request = Request> extends ConnectionCredentials {
```

src/room/ConnectionCredentials.ts
1egoman: Good point, other classes did this so I just followed the same pattern. Removed.

src/room/ConnectionCredentials.ts
1egoman: Talked with lukas, we decided to hold off on this for later given a user wouldn't really be able to control the data coming in - `setRequest` will be called in livekit controlled code.

src/room/ConnectionCredentials.ts
bcherry: We should think a lot about what we want to be able to configure at this level and adjust the sandbox server to match. I think for an agent use case the #1 most important detail, and maybe the only detail that matters, is the `agent_name` of the agent you'd like to connect to. it would be ideal if you could specify only that and everything else would be automated.

We can freely redesign the schema of the token endpoint (we'll find a way to support backwards compatibility) so lets just ensure we get it right and think through all of it.

bcherry: to be more specific, i think the user in the frontend should be able to specify all of these values, which should all be optional in the request.

partiticipant
- identity (default: random/auto-generated)
- name (default: same as identity)
- metadata (default: empty)
- attributes (default: empty)

room
- name (default: random/auto-generated)
- room_config, incl. agent dispatches (default: empty)

per agent dispatch...
- agent_name (default: empty)
- metadata (this becomes job metadata) (default: empty)


We should also have an optional simplified API for dispatching agents without having to bury them in room config.

The response should probably have

- ws server url
- token
- room name
- participant identity
- participant

the last three make sense on the response because they can be auto-generated by the token endpoint so you'll want to know the values you got.

We should also probably explore making participant name and identity configurable on agent dispatch, so that you can await a specific agent participant identity rather than just their agent_name. But that doesn't need to be a blocker, it would require updates across many things.

1egoman: @bcherry 

> I think for an agent use case the https://github.com/livekit/client-sdk-js/issues/1 most important detail, and maybe the only detail that matters, is the agent_name of the agent you'd like to connect to.

Keep in mind, this abstraction is at a similar level as the `Room` and there will be another layer on top that uses the apis this provides (namely, `ConnectionCredentials.setRequest`) to inject the agent name to dispatch. Given this, I'd prefer to keep `roomConfig` in there and let the react layer deal with the complexity of setting the right value within rather than handle it at this layer explictly.

If it's helpful in framing this change, I have been proposing something conceptually like the below, which maybe could give you an idea of how this could look at the agent sdk level:
```typescript
// NOTE: conversation is a room facade
const conversation = useConversationWith("name of agent to dispatch", {
  credentials: /* connectioncredentials instance goes here */
});

useEffect(() => {
  conversation.begin();
  return () => {
    conversation.end();
  };
}, []);

// Then agent state can be accessed within a mixture of `conversation` fields and
// other new / preexisting hooks
```

> to be more specific, i think the user in the frontend should be able to specify all of these values, which should all be optional in the request.
> 
> partiticipant
> 
> identity (default: random/auto-generated)
> name (default: same as identity)
> metadata (default: empty)
> attributes (default: empty)
> room
> 
> name (default: random/auto-generated)
> room_config, incl. agent dispatches (default: empty)
> per agent dispatch...
> 
> agent_name (default: empty)
> metadata (this becomes job metadata) (default: empty)

Going to leave some notes on each of these fields:
- Participant `name` - this exists currently, but is called `participantName`.
- Participant `identity` - I don't know if I understand fully the difference between name and identity but I have no concerns with adding it if there are good reasons to do so.
- Participant `metadata` / `attributes` - Good idea, I can add these.
- Room `name` / `roomConfig` - these exist currently but maybe could be renamed to match what you are proposing?
- `agent_name` / `metadata` - Mentioned above, my preference would be for this to not be duplicated here and be handled at a higher abstraction layer.

> The response should probably have
> 
> ws server url
> token
> room name
> participant identity
> participant

Notes also on these fields:
- `ws server url` / `token` - these exist currently
- `room name` / `participant identity` - this could be decoded from the token payload, just like I'm currently doing with `roomConfig`. My preference would be to do that for this value (ie, add another getter to `ConnectionCredentials`) rather than duplicate it in the response again.
- `participant` - I'm not sure what this represents, can you elaborate on what this would be? Maybe you meant `participant name`? I'd need to check but that might be able to come from the token payload as well.

bcherry: sorry yes i meant participant name - and also i'm not meaning to be prescriptive about how they're named or what's already there vs not, just listing it all out

Once we agree on the ideal spec for the endpoint, we need to update the one used in sandbox to match and then we can build sample implementations for the other languages (or build it all into the server sdks so it's very simple to deploy a compliant endpoint). 

i agree it's probably fine to do the agent abstraction in the react layer for JS at least. I like that proposed hook! (useConversationWith)

identity is the unique identifier for a participant within a room. if another client tries to connect with the same identity, the original participant is booted from the room. there's also a participant ID generated by the server which identifies each actual connection, but the identity field is the meaningful one to developers.  meanwhile the participant name is meant to be user-facing. So you'd imagine that in a conferencing app, the user could set their name to whatever they want (e.g. "Ben") and multiple users could have the same name. But the developer would control the identities used, which would probably map to their own user_id field from their database (or another unique identifier like an email address or username)

src/room/Room.ts
bcherry: are we ready to actually deprecate token-based connection? I think this may cause a lot of concern from existing customers who have their own token server implementations already and all get told they're being deprecated when they upgrade SDKs.

1egoman: @bcherry It's hard for me to know in a vaccum how much of a concern this is in practice- two things worth mentioning:
- It's purely a typescript level deprecation - so you'll get a typescript warning when you build your end app and a message in your IDE if you use an IDE that supports that. While obviously the goal is that folks would migrate to the new api, the old signature will always exist for this current major version. So there's zero impact to existing products already built on top of it, it's purely an aesthetic concern for folks who want to keep warnings from showing up.
- There is a very straightforward migration path to the new interface: `room.connect("server url", "participant token")` becomes `room.connect(new ConnectionCredentials.Literal({ serverUrl: "server url", participantToken: "participant token" }))` (and I won't type it out, but the same thing for `prepareConnection` too). Maybe even this equivalence could be included in the deprecation warning to make the migration even easier. But since there's a one to one conversion that is so trivial, it seems like to me that even if somebody wanted to clean up this warning, it would take like 10 seconds to do so.

All this being said, this fundamentally is a product decision, and while I have thoughts, it is not up to me. So, if adding the deprecation here is indeed too burdensome for users, I'd be fine with removing it and just supporting both signatures. That won't effect the agents sdk work since the agents sdk react level logic will just use the new form and not the old. Let me know what you think here and if I need to remove the deprecation, let me know!

One other thing - I sent you a slack message about docs updates being required for this change, and this was largely what I was talking about. I just want to surface that again - no matter what is decided here around the deprecation, I still think the docs need to be updated to use this new signature rather than the existing one. I'd really like to ensure that new users are pushed to the new stuff and not using the old stuff going forwards if at all possible.

src/room/ConnectionCredentials.ts
1egoman: Re identity - got it, makes sense.

Logging the changes made coming out of @bcherry's comment above:

- Added `participantIdentity`, `participantMetadata`, `participantAttributes` fields to `ConnectionCredentials.Request`
- Added explicit entries for room name, participant identity, and participant name to `ConnectionCredentials.getCachedResponseJwtPayload` return type
- Added `ConnectionCredentials.getCachedResponseParticipantName` / `ConnectionCredentials.getCachedResponseParticipantIdentity` / `ConnectionCredentials.getCachedResponseParticipantMetadata` / `ConnectionCredentials.getCachedResponseParticipantAttributes` methods

src/room/Room.ts
bcherry: Oh that makes sense - I didn't realize there's the literal provider. Can you note that more explicitly in the deprecation notice?

Yeah we'll do all the docs updates when we're ready!

src/room/ConnectionCredentials.ts
bcherry: We'll also need to update https://github.com/livekit-examples/token-server-node (sample token server) and https://github.com/livekit/cloud-api-server/blob/main/pkg/service/sandbox.go#L680 (actual token server used in sandbox production to match (cc @rektdeckard)

src/room/ConnectionCredentials.ts
rektdeckard: Some cloud providers don't use fully standards-compliant JS engines for their web workers, and may not populate `process.env` or may not do so without additional configuration. For example, Cloudflare [makes you use a special node compatibility mode](https://developers.cloudflare.com/workers/configuration/compatibility-flags/#enable-auto-populating-processenv).  Not sure how safe this is.

Also, it may be useful to show the warning *in development* too, so devs see it and can take action.

src/room/ConnectionCredentials.ts
rektdeckard: `DevelopmentTokenServer`?

src/room/ConnectionCredentials.ts
```diff
@@ -0,0 +1,149 @@
...
+      const response = await fetch(`${baseUrl}/api/sandbox/connection-details`, {
+        method: 'POST',
+        headers: {
+          'X-Sandbox-ID': this.options.sandboxId,
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          roomName: this.options.roomName,
+          participantName: this.options.participantName,
+        }),
```
rektdeckard: Do you plan on adding support for additional `roomConfig`, etc. to this token client too? If you do, a little FYI we are moving to `snake_case` property names to better support arbitrary config properties and have a more language-agnostic API.

Ref: https://github.com/livekit/cloud-api-server/blob/b11d25cbd193c260bfd12b85ef73e1f8f657b92d/pkg/service/sandbox.go#L56-L88

1egoman: Yes, I was going to do this as part of [this chunk of work](https://github.com/livekit/client-sdk-js/pull/1645#discussion_r2349328292) once the sandbox token server endpoints support the additional fields.

Sounds good on the snake case - I think I'll probably opt to keep the `ConnectionCredentials.Request` interface camelcase and do the conversion as part of the `ConnectionCredentials.SandboxTokenServer` implementation here unless there's a reason that comes to mind why that would not be ideal.

src/room/Room.ts
1egoman: Sounds good - made the deprecation notices a bit more verbose with an example!

src/room/ConnectionCredentials.ts
1egoman: @rektdeckard 

> Some cloud providers don't use fully standards-compliant JS engines for their web workers

Interesting point - I hadn't considered that. I wonder if something like `globalThis?.process?.env?.NODE_ENV` would be a better approach here, assuming what I'm doing is even a good idea? Then in non node-compliant environments the log wouldn't show which is probably better than the inverse.

> Also, it may be useful to show the warning in development too, so devs see it and can take action.

I'm curious other's perspectives on this - yea, it could make sense, and actually it's where I started when I first wrote this. But it would mean that you'd always see a warning log in your js console as a developer, which would be annoying and would mean that most people would end up setting that `disableSecurityWarning` flag, which would mean that when they eventually deploy their app to production the log would be hidden and they would forget to swap to a more secure token generation mechanism 🤦 .

Maybe I could separate it out so that there'd be `disableSecurityWarningInDev` and `disableSecurityWarningInProduction` or something similar? There's also the argument that maybe this is all just too heavy handed and sternly worded docs messages could be good enough. I don't have a strong opinion but if you / anyone else does I'd be interested in hearing it.

bcherry: i wouldn't put this warning in here at all. it's fine to use in a production _environment_, so long as it's in a product that's in development/prototype stage. we can't know that from the frontend itself

just ensure there's a good docstring and we'll also ensure the docs on the site and in the sandbox feature are good (i'm planning to rewrite a lot of token docs after all this ships).  

we will implement a warning at some point but it's going to be in the cloud dashboard and over email, based on the amount of raw traffic your project is getting

1egoman: Sounds good - removed.

1egoman:
Renamed from `ConnectionCredentials` to `TokenSource` after a decision in a meeting, and rebase to fix merge conflicts

src/room/TokenSource.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  getCachedResponseParticipantIdentity() {
+    return this.getCachedResponseJwtPayload()?.sub ?? null;
+  }
+
+  getCachedResponseParticipantMetadata() {
+    return this.getCachedResponseJwtPayload()?.metadata ?? null;
+  }
+
+  getCachedResponseParticipantAttributes() {
```
lukasIO: I think you wanted to remove all of these granular methods?

src/room/TokenSource.ts
```diff
@@ -0,0 +1,267 @@
...
+    /**
+     * A RoomConfiguration object can be passed to request extra parameters should be included when
+     * generating connection credentials - dispatching agents, defining egress settings, etc
+     * @see https://docs.livekit.io/home/get-started/authentication/#room-configuration
+     */
+    roomConfig?: RoomConfiguration;
+  };
+  export type Response = {
+    serverUrl: string;
+    participantToken: string;
```
lukasIO: should we mirror snake case here and in the request for it to align more closely with what an endpoint is expected to actually consume and return?

src/logger.ts
```diff
@@ -12,6 +12,7 @@ export enum LogLevel {
 export enum LoggerNames {
   Default = 'livekit',
   Room = 'livekit-room',
+  TokenSource = 'livekit-connection-credentials',
```
lukasIO: maybe update the name here to avoid confusion

src/room/ConnectionCredentials.ts
1egoman: `token-server-node` change: https://github.com/livekit-examples/token-server-node/pull/59 (merged)
`cloud-api-server` change: https://github.com/livekit/cloud-api-server/pull/1293 (merged)

src/logger.ts
1egoman: 🤦 missed this, thanks!

src/room/TokenSource.ts
1egoman: I mentioned this here: https://github.com/livekit/client-sdk-js/pull/1645#discussion_r2349554822.

My preference would be to keep it camel case since those are the language conventions and do the conversion within `TokenSource.SandboxTokenServer`. At least to me there doesn't seem like there's a strong reason to break those language conventions given the meaning of each field is easily understood to be the same thing in either form. But maybe there's something I'm not taking into account there?

src/room/TokenSource.ts
1egoman: Didn't get around to this yet, but thanks for the reminder - just removed.

src/room/TokenSource.ts
lukasIO: What I was imagining is using the same type (Response and Request) also in the endpoint (and token server) examples. 

That's the main reason why I was thinking about it to match more closely what's actually in the payload of the request. 
edit: ah, and maybe also the fact that if user construct custom requests that conform to our proposed "standard" that what they set is what they get kind of

1egoman: I see, yea that does make a certain sort of sense.

Doing a bit more pondering on it, I think my concern primarily is a naming one. I think if I renamed `Request` => `RequestPayload` and `Response` => `ResponsePayload` (or maybe `-Body`?) my concerns go away (the suffix to me implies something more raw while leaving the un suffixed names available leaves the door open for a parallel camelcase version in the future if that is useful to have an internal only representation).

I'll convert them over to snake case and also at the same time add the name suffixes.

.changeset/mean-mirrors-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+add ConnectionCredentials token fetching abstraction
```
davidzhao: reminder to update naming changes here as well

examples/demo/demo.ts
```diff
@@ -164,8 +165,13 @@ const appActions = {
   ): Promise<Room | undefined> => {
     const room = new Room(roomOptions);
 
+    const tokenSource = new TokenSource.Literal({
```
davidzhao: seems cleaner to have a factory method instead.

also we should use camelCase in all JS types. that's more native here.

```suggestion
    const tokenSource = TokenSource.staticToken({
       url,
       token,
    })
```

src/room/Room.ts
```diff
@@ -588,7 +591,31 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
    * With LiveKit Cloud, it will also determine the best edge data center for
    * the current client to connect to if a token is provided.
    */
-  async prepareConnection(url: string, token?: string) {
+  prepareConnection(connectionCredentials: TokenSource): Promise<void>;
```
davidzhao:
```suggestion
  prepareConnection(tokenSource: TokenSource): Promise<void>;
```

src/room/TokenSource.ts
```diff
@@ -0,0 +1,312 @@
...
+    baseUrl?: string;
+  };
+
+  /**
+   * A temporary v1 sandbox token server adaptor for backwards compatibility while the v2 endpoints
+   * are getting deployed.
+   *
+   * FIXME: get rid of this before merging the TokenSource pull request!!
+   * */
+  export class SandboxTokenServerV1 extends TokenSource.Refreshable {
```
davidzhao: could the server just be built in a backwards compatible way so we could hard flip?

davidzhao (changes requested):
looks good! just a few nitpicks

examples/demo/demo.ts
lukasIO: Discussed with Ryan that ideally we'd want to use protobuf's `toJson` internally here which gives us
- single source of truth for clients
- language idiomatic casing styles
- consistent on the wire format (snake case'd JSON as defined by Tobias)

.changeset/mean-mirrors-change.md
1egoman: Oops! Thanks

src/room/Room.ts
1egoman: I must have missed this file when renaming, there were a few lagging old name references. Should all now be fixed!

src/room/TokenSource.ts
1egoman: This is very temporary, and should be deleted imminently.

The only reason this exists right now is [there is an in flight cloud-api-server change here](https://github.com/livekit/cloud-api-server/pull/1293) (that I worked with @rektdeckard on) which hasn't made its way to production yet. The `SandboxTokenServer` implementation uses these new changes behind the scenes, but for local development I needed something that continued to work in production so I kept the old version (`SandboxTokenServerV1`) around.

Once that cloud-api-server change ends up in production, this will be deleted and [this change](https://github.com/livekit-examples/token-server-status/pull/33) will also be promoted to production at the same time.

examples/demo/demo.ts
1egoman: > seems cleaner to have a factory method instead.

The reason why I did it this way over an explicit factory constructor type pattern was that fundamentally it's challenging to model all the `TokenSource` functionality as one class. There's too much breadth in functionality for a single monolithic class to work in all cases. Given this, it makes sense to split it up across a few levels of abstraction like I have done now. That inevitably means you end up with `TokenSourceLiteral`, `TokenSourceCustom`, `TokenSourceSandboxTokenServer`, etc. 

The question becomes then how these should be exposed to a user. What I did here with a namespace I thought was fairly elegant, and seemed reminiscent to me of [dot notation in react component names](https://legacy.reactjs.org/docs/jsx-in-depth.html#using-dot-notation-for-jsx-type) (like `<MyContext.Provider value={...}>`), but I totally understand it could be seen as maybe a little weird / unconventional. 

Assuming the existing pattern I have isn't preferable, I see two other realistic options:
- **Option 1**: Make the existing `TokenSource` implementations all private, and add factory static methods similar to what you have proposed. In other words, at a 10k foot view, something like this:
```tsx
export class TokenSource {
  /* ... implementation code here ... */
  
  /** "literal" doc string here */
  static literal(request: TokenSourceRequest) {
    return new TokenSourceLiteral(request);
  }
  /** "custom" doc string here */
  static custom(fn: (request: TokenSourceRequest) => Promise<TokenSourceResponse>) {
    return new TokenSourceCustom(fn);
  }
  /* ... etc ... */
}

/** copy of "literal" doc string here (ideally) */
class TokenSourceLiteral extends TokenSource { /* ... */ }
/** copy of "custom" doc string here (ideally) */
class TokenSourceCustom extends TokenSource { /* ... */ }
// ... etc ...

// Then, to use it later on:
import { TokenSource } from 'livekit-client';
const tokenSource = TokenSource.literal({ /* ... */ });
```

The biggest con to be of the above is just all the duplication / boilerplate through all the levels of indirection (especially the doc strings... I guess it's not _necessarily_ required to duplicate them, but it sucks that we wouldn't have those on the internal abstraction levels), but maybe that's worth for the user it if this is really the best interface.

- **Option 2**: Get rid of the namespace and expose all classes as separate imports all starting with the same prefix. This gets rid of the leading namespace with the dot and embraces fully everything just being a class. So for example:
```tsx
export class TokenSource {
  /* ... implementation code here ... */
}

/** "literal" doc string here */
export class TokenSourceLiteral extends TokenSource { /* ... */ }
/** "custom" doc string here */
export class TokenSourceCustom extends TokenSource { /* ... */ }
// ... etc ...

// Then, to use it later on:
import { TokenSourceLiteral } from 'livekit-client';
const tokenSource = new TokenSourceLiteral({ /* ... */ });
```

While it is verbose, the nice thing about this approach is the lack of duplication - no multiple sets of function args to keep up to date, only one doc string, etc. Also another nice thing is anybody else who wanted to make their own token source could just make a class starting with a `TokenSource` prefix and it matches with the stock implementations.

-----------

I generally am a fan of what I have now, but I'm curious where you / others like @lukasIO weigh in between what I have currently / option 1 / option 2 given all the pros and cons I outlined.

bcherry: I think either what you have already or your option 1 is correct, I wouldn't go for option 2. I don't have much preference between factory vs constructor here, but in swift for instance the factory option is better because you can omit the namespace, something like this:

```
room.connect(.literal("my-token"))
```

src/room/TokenSource.ts
```diff
@@ -0,0 +1,312 @@
+import { RoomConfiguration } from '@livekit/protocol';
+import { decodeJwt } from 'jose';
+import log, { LoggerNames, getLogger } from '../logger';
+
+const ONE_SECOND_IN_MILLISECONDS = 1000;
+const ONE_MINUTE_IN_MILLISECONDS = 60 * ONE_SECOND_IN_MILLISECONDS;
```
bcherry: i always find these clearer when expressed as conversion factors more directly, like this:

```suggestion
const MSEC_PER_SEC = 1000;
const MSEC_PER_MIN = 60 * ONE_SECOND_IN_MILLISECONDS;
```

bcherry (approved):

examples/demo/demo.ts
davidzhao: I prefer option 1.. it's verbose to copy the docstring once.. but for a end user it's so much nicer.. I don't have to think about these types to import. For most this could also be typing `TokenSource.` and seeing the completion on the options on how to create it.. that in itself makes the system a lot more intuitive.

davidzhao: I still don't like snake case.. this feels very foreign to someone who's programming in JS.

why are we making it harder for end users (forcing snake_case) in order to satisfy the wire protocol? to me that's something we should be handling internally.. all of the protobuf libs can also parse & generate in both casing.. if that's what we are concerned about

src/room/Room.ts
```diff
@@ -958,6 +1007,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       this.handleDisconnect(stopTracks, DisconnectReason.CLIENT_INITIATED);
       /* @ts-ignore */
       this.engine = undefined;
+      this.tokenSource?.generate();
```
davidliu: What's this extra token generation upon disconnecting for? Preparing a new token for a subsequent connection?

1egoman: > Preparing a new token for a subsequent connection?

Yep! It's to make connect -> disconnect -> connect cycles faster.

examples/demo/demo.ts
1egoman: > all of the protobuf libs can also parse & generate in both casing.. if that's what we are concerned about

A few comments up the thread, this was what @lukasIO and I decided as well. Using a protobuf for these means we get the preferred field name casing on the js end without having to write manual mapping logic, plus it's a source of truth for all sdks which seems like a nice property.

Pull request to introduce the protobuf messages: https://github.com/livekit/protocol/pull/1224

Once that gets merged I'll integrate it in here.

1egoman: > I prefer option 1.

I still generally prefer what I originally wrote (namespace with classes inside), which has all the same benefits of option 1 you outlined (only one type to import, autocomplete behavior, etc).

That being said, I don't know if I have a strong enough opinion to keep going around and around here and given @lukasIO also prefers the static method approach over the class namespace approach, I'll migrate to what I had outlined in option 1 above.

It's worth noting that the existing classes will still likely need to be exported since they will be needed for typescript typing (and also maybe `instanceof` checks? Though that seems like it would be less important to me), but I'll move over to using the static methods as the preferred constructors.

1egoman: Updated to use protobuf like @lukasIO and I had discussed - in effect this means now that all external interfaces are camelcase, except for the parameters passed to a `Custom` handler function since a user needs to be given the snake case form to pass to a server.

So:
```tsx
// Camelcase fields used as part of the external interface...
const tokenSourceLiteral = TokenSource.literal({ serverUrl: "...", participantToken: "..." });

const tokenSourceSandboxTokenServer = TokenSource.sandboxTokenServer({
  baseUrl: 'https://xxx.livekit.app',
  sandboxId: '...',
});

// ... except for here, where the snake case forms need to be exposed to meet the standard specification
const tokenSource = TokenSource.custom(async (req) => {
  const requestedRoomConfig = req.room_config;
  // ... etc ...
  return {
    server_url: url,
    participant_token: token,
  };
});
```

src/utils/camelToSnakeCase.ts
```diff
@@ -0,0 +1,16 @@
...
+type ObjectKeysToSnakeCase<Obj> = {
+  [Key in keyof Obj as CamelToSnakeCase<string & Key>]: NonNullable<ValueToSnakeCase<Obj[Key]>>;
+};
+
+export type ValueToSnakeCase<Value> =
+  Value extends Array<infer Item>
+    ? ArrayValuesToSnakeCase<Item>
+    : Value extends object
+      ? ObjectKeysToSnakeCase<Value>
+      : Value;
```
1egoman, line 16: Some context on this: As far as I can tell, there's no way to get a type of a snake case version of a protobuf type with the current protobuf infrastructure / library. Unfortunately, the protobuf library seems to return an opaque `JsonValue` type which is used everywhere that a snake case form could have maybe been found.

This is relevant because @lukasIO had brought up a [really good point](https://github.com/livekit/client-sdk-js/pull/1645#discussion_r2359938543) earlier on that it would be great to expose the snake case version so it can be used in the context of a server written in typescript for applying a type to the server request / response body. I agree and would like to be able to do this.

So this leaves us with a few options:
1. Forward along the opaque `JsonValue` type and don't expose the snake case types here at all.
2. Maintain a duplicate copy of the request / response types here in snake case
3. Somehow convert from the camelcase to snake case forms dynamically

I opted for 3 given it ended up being less complicated than I thought, but I dislike this kind of thing - it adds a fair bit of complexity. But I think out of the options it's the least bad one in my eyes at least.

src/index.ts
```diff
@@ -66,6 +66,7 @@ export * from './room/errors';
 export * from './room/events';
 export * from './room/track/Track';
 export * from './room/track/create';
+export { TokenSource } from './room/token-source/TokenSource';
```
lukasIO: should we also export some types for convenience?

1egoman: Oops, I thought this was an `export *`! Fixed, thanks.

examples/demo/demo.ts
1egoman: Updating this thread with some context on some internal discussions over the past week or so:

## Add "fixed" and "configurable" subtypes
A new notion of "fixed" and "configurable" token sources now exists [(related code here)](https://github.com/livekit/client-sdk-js/pull/1645/files#diff-404ba88626acaa8a13b0dc42130e407427e3192142d7dea09d0615a586ff42caR48-R80) - "fixed" doesn't take parameters and returns what it is going to return, and "configurable" takes parameters which end up encoded in the token.

Current fixed token source types: `literal`.
Current configurable token source types: `endpoint`, `custom`, `sandboxTokenServer`.

```tsx
const fixed = TokenSource.literal({ serverUrl: "...", participantToken: "..." });
fixed.fetch(); // <-- typescript won't let you put an options object in here

const configurable = TokenSource.sandboxTokenServer("sandbox id");
configurable.fetch({}); // <-- typescript REQUIRES that you put an options object in here
```

## Add `TokenSourceOptions`
A new `TokenSourceOptions` object has been introduced to abstract `TokenSourceRequest`, which is now used by "configurable" token sources. This contains a less verbose form of the `TokenSourceRequest` object fully in camelcase.

We have gone back and forth a bit on this, but I (Ryan) am still not a super big fan of doing this because of the extra layer of abstraction that doesn't (to me) seem to add much value given `useSession` will be the primary user of this interface and it's an extra layer to thread any schema changes through. But it sounds like I'm in the minority on this view so I think I'm just going to go along with it for now.

```tsx
// For example: TokenSourceOptions contains agentName as a more abstract form of the
// `roomConfig` token payload. So:
const tokenSource = TokenSource.sandboxTokenServer("sandbox id");
const result = await tokenSource.fetch({ agentName: "my agent to dispatch" });
```
  
## Change a few subtype function signatures  
The `TokenSource.custom` (a "configurable" token source) handler function contract has changed to `(options: TokenSourceOptions) => Promise<TokenSourceResponse>`, and the `TokenSource.literal` (a "fixed" token source) now can take a function and return a computed value in addition to a literal value.

## Make `TokenSource` stateless again
Earlier in their evolution, `TokenSource`'s were fully stateless, but at one point it was decided to make them stateful with a `setRequest` method rather than passing in the request/options into the credentials fetching function. Due to some comments we've now decided to invert that decision and make them stateless again which resolves some of the challenges of the `useSession` abstraction trying to declaratively call an imperative api.

```tsx
const tokenSource = TokenSource.sandboxTokenServer("sandbox id");
// No more setOptions / setRequest call here
const result = await tokenSource.fetch({
  agentName: "my agent to dispatch"
} /* <-- the options get passed in here instead */);
```

## Room integration
We've decided for the time being to NOT integrate `TokenSource` handling directly into the `room.connect` call, and to make this a follow up change so we can think a little more about how it should work once some miles have been put on the `useSession` abstraction.

```tsx
// Current state, given this decision. 
const tokenSource = TokenSource.literal(async () => {
  // ... make a http request here ...
  return {
    serverUrl: "...",
    participantToken: "...",
  };
});

const results = await tokenSource.fetch();
room.connect(results.serverUrl, results.participantToken);

// Revisit this later - maybe try to pass in `tokenSource` into the `new Room()` constructor in
// a backwards compatible way
```

pblazej (approved):

Issue #1655, opened 2025-09-18
Audio Context gets closed and cannot be resumed


indexds:
### Describe the bug

Hello, 

I consistently get an issue that only occurs when using a headset. Everything will work perfectly for a few minutes after a call was started, until a certain point where the AudioContext gets closed by the browser(?) at which point I am screwed because it cannot be resumed. I assume livekit is trying to do that automatically because we get the following error in the logs:

Error trying to auto-resume audio context DOMException: Can't resume if the control thread state is "closed".

I tried setting webAudioMix to both true and false and the error remained.

My application is running on Firefox 142.0.1 and was also tested on Chromium with the same results.

Any insight is appreciated as this makes the application completely unusable.

Related issues: #1116 #1445 

### Reproduction

- Plug in a headset
- Wait for many(?) minutes
- Alt-tab out of the application(?)
- Eventually the AudioContext will be force closed by the browser at which point the only way to fix it is to close and reopen the browser.

### Logs

```shell
Error trying to auto-resume audio context DOMException: Can't resume if the control thread state is "closed"
    handleResume http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:10333
    __awaiter http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:5765
    __awaiter http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:5747
    handleResume http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:10329
    getNewAudioContext http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:10340
    detectSilence http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:10300
    __awaiter http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:5765
    __awaiter http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:5747
    detectSilence http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:10297
    checkForSilence http://localhost:3000/node_modules/.vite/deps/chunk-673D7ERD.js?v=3e4507ff:14375
[...]

PR #1660, opened 2025-09-22
Properly clean up event listeners in getNewAudioContext()

indexds:
#1655 related

### DIFF
--- a/src/room/track/LocalAudioTrack.ts
+++ b/src/room/track/LocalAudioTrack.ts
@@ -244,7 +244,7 @@ export default class LocalAudioTrack extends LocalTrack<Track.Kind.Audio> {
     const trackIsSilent = await detectSilence(this);
     if (trackIsSilent) {
       if (!this.isMuted) {
-        this.log.warn('silence detected on local audio track', this.logContext);
+        this.log.debug('silence detected on local audio track', this.logContext);
       }
       this.emit(TrackEvent.AudioSilenceDetected);
     }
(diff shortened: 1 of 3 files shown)


src/room/track/LocalAudioTrack.ts
```diff
@@ -244,7 +244,7 @@ export default class LocalAudioTrack extends LocalTrack<Track.Kind.Audio> {
     const trackIsSilent = await detectSilence(this);
     if (trackIsSilent) {
       if (!this.isMuted) {
-        this.log.warn('silence detected on local audio track', this.logContext);
+        this.log.debug('silence detected on local audio track', this.logContext);
```
lukasIO, line 247: I can get behind demoting this to `info` level, `debug` seems a bit too low

lukasIO (commented):
thanks for the PR. 

looks good apart from the log level on the silence detection

src/room/track/LocalAudioTrack.ts
indexds: My problem with it was that it is spamming the console for no good reason, an event is being fired either way, if someone wants to listen to it they can, no?

indexds: Your call though, will move to info

lukasIO: yeah, the event being emitted is a good point. 
in that light `debug` seems fine

indexds:
Aight this should be mergeable then? @lukasIO

lukasIO (approved):

lukasIO:
thank you @indexds !

Issue #1661, opened 2025-09-25
TypeScript error: 'Uint8Array' is not generic in LocalTrack.d.ts (Angular 18 / TS 5.4)


attapolmkt:
### What I'm expecting
TypeScript types in livekit-client to be compatible with TypeScript 5.4 and Angular 18.

### What happens instead
Type error: `Type 'Uint8Array' is not generic.`

`getPreConnectBuffer(): ReadableStream<Uint8Array<ArrayBufferLike>> | undefined;` in LocalTrack.d.ts causes TypeScript compilation to fail, because `Uint8Array` is not a generic type. Should be just `Uint8Array`.

### Reproduction

1. Install `livekit-client` (`^2.15.7`) in Angular 18 project (TypeScript 5.4.x)
2. Run `ng serve` or `ng build`
3. See error: `Type 'Uint8Array' is not generic.` from `node_modules/livekit-client/dist/src/room/track/LocalTrack.d.ts:117:43`

### Logs

```shell
Error:

Error: node_modules/livekit-client/dist/src/room/track/LocalTrack.d.ts:117:43 - error TS2315: Type 'Uint8Array' is not generic.
117     getPreConnectBuffer(): ReadableStream<Uint8Array<ArrayBufferLike>> | undefined;
                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### System Info

```shell
System:
- macOS (Apple Silicon, tested on MacBook Pro)
- Node.js v18+
- Angular 18
- TypeScript 5.4.5

Browsers:
- Chrome latest

livekit-client: 2.15.7
```

### Severity

annoyance

### Additional Information

This type error blocks compilation in strict TypeScript environments. Workaround: Patch the d.ts file manually (`Uint8Array<ArrayBufferLike>` → `Uint8Array`). Please fix for better TS 5.x compatibility.

lukasIO:
Thanks for the report! I opened a PR to address this.

PR #1668, opened 2025-10-01
Fix TS 5.7 generic Uint8Array declaration when inferred

lukasIO:
closes https://github.com/livekit/client-sdk-js/issues/1661

So far we’ve kept pretty much up to date with TS versions and used `downlevel-dts` to transpile any non compatible TS constructs down for older versions. This CLI tool hasn't been updated in a while though, so this particular issue at hand is not part of the fixes it supports.

When a `Uint8Array` like  type is inferred directly from usage within a TS >5.7 project (like this one) then the emitted `d.ts` will use something like `Uint8Array<ArrayBuffer>` which is incompatible with older TS versions. 

The options we have is 
- explicitly declare `Uint8Array` as return type (or declared type) 
- use a type helper for _all_ Uint8Array usage that looks like `ReturnType<Uint8Array.from>`
- write a babel plugin that transforms it 

I opted for the explicit declaration for now.

full context: https://github.com/microsoft/TypeScript/issues/62240

### DIFF
--- a/.changeset/fifty-shoes-act.md
+++ b/.changeset/fifty-shoes-act.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Fix TS 5.9 generic Uint8Array declaration when inferred


package.json
```diff
@@ -50,7 +50,7 @@
     "format": "prettier --write src examples/**/*.ts",
     "format:check": "prettier --check src examples/**/*.ts",
     "ci:publish": "pnpm build:clean && pnpm compat && changeset publish",
-    "downlevel-dts": "downlevel-dts ./dist/ ./dist/ts4.2 --to=4.2",
+    "downlevel-dts": "downlevel-dts ./dist/src ./dist/ts4.2 --to=4.2",
```
lukasIO, line 53: drive by fix that ensures we really only downlevel the emitted d.ts files in the dist/src folder where the original source types are emitted to.

1egoman (approved):
What's here makes sense to me!

That being said, as is this change doesn't do anything to prevent this from happening in the future, to me it seems like without some guardrails, this will inevitably regress.

I realize this might be a bit of a tedious ask, but maybe it would be worth adding to CI something like:
1. Build the library normally.
2. Install the oldest supported typescript version.
3. Use that oldest supported typescript version to build a script containing something like `import 'livekit-client'`.
4. If step 3 it returns a non-zero code, fail the CI build.

Thoughts on that?

PR #1669, opened 2025-10-01
Apply server supplied cache control settings for region url provider

lukasIO:


src/room/utils.ts
```diff
@@ -720,3 +720,14 @@ export function splitUtf8(s: string, n: number): Uint8Array[] {
...
+export function extractMaxAgeFromRequestHeaders(headers: Headers): number | undefined {
+  const cacheControl = headers.get('Cache-Control');
+  if (cacheControl) {
+    const maxAge = cacheControl.match(/max-age=(\d+)/)?.[1];
+    if (maxAge) {
+      return parseInt(maxAge, 10);
+    }
+  }
+  return undefined;
+}
```
1egoman, line 733: There probably is a good reason you wrote this yourself, but just FYI in case, it looks like there's some fairly decent packages for parsing cache control headers on npm like [this one](https://github.com/etienne-martin/cache-control-parser).

One of those could maybe be a more robust option, but on the other hand, if the input is always going to be coming from a livekit-written endpoint maybe it isn't all that important for the parser to be super robust and to depend on some questionably maintained package.

src/room/utils.test.ts
```diff
@@ -60,3 +60,106 @@ describe('splitUtf8', () => {
     expect(splitUtf8('', 5)).toEqual([]);
   });
 });
+
+describe('extractMaxAgeFromRequestHeaders', () => {
+  it('extracts max-age from valid Cache-Control header', () => {
```
1egoman, line 65: One case maybe worth adding here (if you don't switch this logic to a pre-existing package as mentioned in a previous comment) - at least by eye, I'm not sure that something like `public, some-unrelated-max-age=1000, max-age=7200` would result in `7200` and not `1000`.

Maybe update the regex with a negative lookahead on the character before to ensure it's not `[a-zA-Z0-9-]`?

src/room/RegionUrlProvider.ts
```diff
@@ -37,7 +37,7 @@ export class RegionUrlProvider {
     if (!this.isCloud()) {
       throw Error('region availability is only supported for LiveKit Cloud domains');
     }
-    if (!this.regionSettings || Date.now() - this.lastUpdateAt > this.settingsCacheTime) {
+    if (!this.regionSettings || Date.now() - this.lastUpdateAt > this.settingsCacheTimeInMs) {
       this.regionSettings = await this.fetchRegionSettings(abortSignal);
```
1egoman, line 41: Very much a nitpick so feel free to ignore if you see things differently, but this would read a lot nicer if the `Date.now() - this.lastUpdateAt > this.settingsCacheTimeInMs` logic were extracted into its own private method IMO.

src/room/utils.ts
lukasIO: > There probably is a good reason you wrote this yourself

didn't actually cross my mind to use a package, I guess the reasons are mostly for conciseness and avoiding 

> to depend on some questionably maintained package

1egoman: Yep makes sense - up to you, both approaches have their pros and cons especially for something this simple.

src/room/utils.test.ts
lukasIO: instead of the negative lookahead I added a change to make sure it's either beginning, whitespace or comma

src/room/RegionUrlProvider.ts
lukasIO: gonna leave this as is for now

Issue #1673, opened 2025-10-01
Engine may throw uncaught errors


bryfox:
### Describe the bug

If our app has issued an unresolved call to `participant.streamBytes` when calling `room.disconnect`, the library may throw uncaught errors. For example: "cannot negotiate on closed engine".

Instead, I would expect that the errors could be catchable on our call to `streamBytes`.

I see a couple of places in RTCEngine where `negotiate()` isn't awaited ([1](https://github.com/livekit/client-sdk-js/blob/e705c15af7ec3fb0195ce4657d0d97600a49a43e/src/room/RTCEngine.ts#L270); [2](https://github.com/livekit/client-sdk-js/blob/e705c15af7ec3fb0195ce4657d0d97600a49a43e/src/room/RTCEngine.ts#L1341)).

Ideally, the "cannot negotiate on closed engine" would also reject with a LivekitError instance.

### Reproduction

Using v2.15.7.

- Connect two participants
- Participant A: concurrently call localParticipant.streamBytes, and room.disconnect

Observed: uncaught error "cannot negotiate on closed engine"

I could write a minimal case for this, but it would be solved by awaiting the `negotiate()` promises noted above.

### Logs

```shell
// Browser logs; context omitted (all entries are for the local participant ID)

disconnect from room
pc state change: from CONNECTED to CLOSED
primary PC state changed 5
clearing ping interval
Uncaught (in promise) cannot negotiate on closed engineUnderstand this error
room event trackUnsubscribed
room event trackSubscriptionStatusChanged
room event trackUnsubscribed
room event trackSubscriptionStatusChanged
room event connectionStateChanged
room event disconnected
could not createOffer with closed peer connection
```

### System Info

```shell
macOS 15.6.1; Chrome 140.0.7339.214
```

### Severity

annoyance

lukasIO:
thanks for the report! 

We don't want to await the negotiate call in order to not delay further execution, but I think a `.catch` with an error log on the negotiate promises makes sense. 

Would you want to open a PR for that?

PR #1674, opened 2025-10-01
Avoid uncaught errors related to send/disconnect races

bryfox:
This adds `catch` in a few places to log instead of throwing uncaught errors.

Fixes https://github.com/livekit/client-sdk-js/issues/1673

The RTCEngine change is prompted by the issue description, and is not awaited per https://github.com/livekit/client-sdk-js/issues/1673#issuecomment-3356968383.

The `outOfBandFailureRejectingFuture` catch is prompted by a similar setup to this issue, where the promise can be rejected before anything even starts iterating the stream. In this case, I observed an uncaught "Participant ... unexpectedly disconnected in the middle of sending data" on Participant A when Participant B sends and disconnects concurrently. This probably needs a little extra review as it seems more subtle, but I think the approach is valid. I also confirmed it avoided the uncaught errors in my test app.

No tests here as I didn't find any precedent for such.

### DIFF
--- a/src/room/data-stream/incoming/IncomingDataStreamManager.ts
+++ b/src/room/data-stream/incoming/IncomingDataStreamManager.ts
@@ -124,6 +124,9 @@ export default class IncomingDataStreamManager {
 
       let streamController: ReadableStreamDefaultController<DataStream_Chunk>;
       const outOfBandFailureRejectingFuture = new Future<never>();
+      outOfBandFailureRejectingFuture.promise.catch((err) => {
+        this.log.error(err);
+      });
 
       const info: ByteStreamInfo = {
         id: streamHeader.streamId,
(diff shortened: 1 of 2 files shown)


src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -124,6 +124,9 @@ export default class IncomingDataStreamManager {
 
       let streamController: ReadableStreamDefaultController<DataStream_Chunk>;
       const outOfBandFailureRejectingFuture = new Future<never>();
+      outOfBandFailureRejectingFuture.promise.catch((err) => {
```
lukasIO, line 127: @1egoman does this look reasonable to you?

1egoman: I think this is fine, it doesn't seem like it interferes with the existing logic, just logs a copy of any errors 👍

PR #1675, opened 2025-10-02
Hide encryption field for now

lukasIO:


1egoman (approved):
Assuming actually releasing the new encryption logic will be a somewhat fast follow change (ie, this code won't stay uncommented for long), this looks good to me.

If it's going to be on the order of months before this gets released (IMO it doesn't sound like that, then it probably makes sense to just get rid of the code all together and re-add it once ready to release.

PR #1676, opened 2025-10-02
Export supportsAudioOutputSelection helper

lukasIO:


src/room/utils.ts
```diff
@@ -151,6 +151,15 @@ export function supportsSetSinkId(elm?: HTMLMediaElement): boolean {
   return 'setSinkId' in elm;
 }
 
+/**
+ * Checks whether or not setting an audio output via {@link Room#setActiveDevice}
+ * is supported for the current browser.
+ */
+export function supportsAudioOutputSelection(): boolean {
```
1egoman, line 158: Nitpick: should this be a static method on `Room` instead?

lukasIO: the other utils in a similar shape are exported in the same way, so opting for consistency

PR #1677, opened 2025-10-03
Add ability to use `TokenSource` in `Room` constructor

1egoman:
Adds the ability to use a `TokenSource` in the `Room`'s constructor to avoid passing in duplicate credentials in `connect` and `prepareConnection`.

## Fixed token source example:
```typescript
const roomA = new Room({
  tokenSource: TokenSource.literal({ serverUrl: "...", participantToken: "..." }),
  // Note: no fetch options can be passed here because TokenSource.literal is
  // not a configurable token source
});
roomA.prepareConnection(); // Note: no parameters required here, because of the token source passed in to the constructor!
roomA.connect(); // Note: no parameters required, but optionally `RoomConnectOptions` can still be passed
```

## Configurable token source example:
```typescript
const roomB = new Room({
  tokenSource: TokenSource.sandboxTokenServer("sandbox id here"),
  agentName: "agent name here", // Note: fetch options can be specified here, but are optional
});
roomB.prepareConnection(); // Note: no parameters required here, because of the token source passed in to the constructor!
roomB.connect(); // Note: no parameters required, but optionally `RoomConnectOptions` can still be passed
```

## Legacy example
This change is fully backwards compatible, so the existing call pattern still works as expected:
```typescript
const roomB = new Room();
roomB.prepareConnection("server url", "optional participant token");
roomB.connect("server url", "participant token");
```

### DIFF
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -3,7 +3,7 @@ import E2EEWorker from '../../src/e2ee/worker/e2ee.worker?worker';
 import type {
   ChatMessage,
   RoomConnectOptions,
-  RoomOptions,
+  RoomOptionsWithTokenSource,
   ScalabilityMode,
   SimulationScenario,
   VideoCaptureOptions,
@@ -109,7 +109,8 @@ const appActions = {
 
     updateSearchParams(url, token, cryptoKey);
 
-    const roomOpts: RoomOptions = {
+    const roomOpts: RoomOptionsWithTokenSource = {
+      tokenSource: TokenSource.literal({ serverUrl: url, participantToken: token }),
       adaptiveStream,
       dynacast,
       audioOutput: {
@@ -151,22 +152,17 @@ const appActions = {
         iceTransportPolicy: 'relay',
       };
     }
-    await appActions.connectToRoom(url, token, roomOpts, connectOpts, shouldPublish);
+    await appActions.connectToRoom(roomOpts, connectOpts, shouldPublish);
 
     state.bitrateInterval = setInterval(renderBitrate, 1000);
   },
 
   connectToRoom: async (
-    url: string,
-    token: string,
-    roomOptions?: RoomOptions,
+    roomOptions?: RoomOptionsWithTokenSource,
     connectOptions?: RoomConnectOptions,
     shouldPublish?: boolean,
   ): Promise<Room | undefined> => {
-    const room = new Room({
-      roomOptions,
-      tokenSource: TokenSource.literal({ serverUrl: url, participantToken: token }),
-    });
+    const room = new Room(roomOptions);
 
     startTime = Date.now();
     await room.prepareConnection();
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -130,10 +130,10 @@ const appActions = {
       videoCaptureDefaults: {
         resolution: VideoPresets.h720.resolution,
       },
-      encryption: e2eeEnabled
-        ? { keyProvider: state.e2eeKeyProvider, worker: new E2EEWorker() }
-        : undefined,
-    } satisfies RoomOptionsWithTokenSourceFixed;
+      // encryption: e2eeEnabled
+      //   ? { keyProvider: state.e2eeKeyProvider, worker: new E2EEWorker() }
+      //   : undefined,
+    };
     if (
       roomOpts.publishDefaults?.videoCodec === 'av1' ||
       roomOpts.publishDefaults?.videoCodec === 'vp9'
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -6,6 +6,9 @@ import type {
   RoomOptionsWithTokenSource,
   ScalabilityMode,
   SimulationScenario,
+  TokenSourceConfigurable,
+  TokenSourceFixed,
+  TokenSourceLiteral,
   VideoCaptureOptions,
   VideoCodec,
 } from '../../src/index';
@@ -109,7 +112,7 @@ const appActions = {
 
     updateSearchParams(url, token, cryptoKey);
 
-    const roomOpts: RoomOptionsWithTokenSource = {
+    const roomOpts = {
       tokenSource: TokenSource.literal({ serverUrl: url, participantToken: token }),
       adaptiveStream,
       dynacast,
@@ -133,7 +136,7 @@ const appActions = {
       encryption: e2eeEnabled
         ? { keyProvider: state.e2eeKeyProvider, worker: new E2EEWorker() }
         : undefined,
-    };
+    } satisfies RoomOptionsWithTokenSource;
     if (
       roomOpts.publishDefaults?.videoCodec === 'av1' ||
       roomOpts.publishDefaults?.videoCodec === 'vp9'
@@ -158,7 +161,7 @@ const appActions = {
   },
 
   connectToRoom: async (
-    roomOptions?: RoomOptionsWithTokenSource,
+    roomOptions?: RoomOptionsWithTokenSource<TokenSourceFixed>,
     connectOptions?: RoomConnectOptions,
     shouldPublish?: boolean,
   ): Promise<Room | undefined> => {
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -3,12 +3,9 @@ import E2EEWorker from '../../src/e2ee/worker/e2ee.worker?worker';
 import type {
   ChatMessage,
   RoomConnectOptions,
-  RoomOptionsWithTokenSourceFixed,
+  RoomOptionsTokenSourceFixed,
   ScalabilityMode,
   SimulationScenario,
-  TokenSourceConfigurable,
-  TokenSourceFixed,
-  TokenSourceLiteral,
   VideoCaptureOptions,
   VideoCodec,
 } from '../../src/index';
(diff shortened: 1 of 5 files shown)


examples/demo/demo.ts
```diff
@@ -128,9 +130,9 @@ const appActions = {
       videoCaptureDefaults: {
         resolution: VideoPresets.h720.resolution,
       },
-      encryption: e2eeEnabled
-        ? { keyProvider: state.e2eeKeyProvider, worker: new E2EEWorker() }
-        : undefined,
+      // encryption: e2eeEnabled
+      //   ? { keyProvider: state.e2eeKeyProvider, worker: new E2EEWorker() }
+      //   : undefined,
     };
```
1egoman, line 136: FYI @lukasIO you missed a `encryption` instance [in your recent pull request](https://github.com/livekit/client-sdk-js/pull/1675).

src/room/Room.ts
```diff
@@ -125,7 +134,9 @@ const connectionReconcileFrequency = 4 * 1000;
  *
  * @noInheritDoc
  */
-class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) {
+class Room<
+  Options extends RoomOptions = RoomOptions,
+> extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) {
   state: ConnectionState = ConnectionState.Disconnected;
```
1egoman, line 140: In order to properly implement the typescript interface for this change, I added a generic to `Room` with a default value, so it should be fully backwards compatible.

I think this potentially could serve as a foundation for more of these types of room interface changes based on option value type updates in the future so I wanted to call it out.

src/options.ts
```diff
@@ -101,10 +103,65 @@ export interface InternalRoomOptions {
...
+  };
+
 /**
  * Options for when creating a new room
  */
-export interface RoomOptions extends Partial<Omit<InternalRoomOptions, 'encryption'>> {}
+export type RoomOptions =
+  | RoomOptionsLegacy
+  | RoomOptionsTokenSourceFixed
+  | RoomOptionsTokenSourceConfigurable;
```
1egoman: The `RoomOptions` interface is now fairly more complex with a few different branches based off of the passed `TokenSource` value.

src/options.ts
```diff
@@ -101,10 +103,65 @@ export interface InternalRoomOptions {
...
+ * be included.
+ */
+type BlockTokenSourceFetchOptions = { [P in keyof TokenSourceFetchOptions]: never };
+
+export type RoomOptionsLegacy = RoomOptionsBase & { tokenSource?: never };
+export type RoomOptionsTokenSourceFixed = RoomOptionsBase &
+  BlockTokenSourceFetchOptions & {
+    tokenSource: TokenSourceFixed;
+  };
+
```
1egoman: I got stuck on this for a while so I wanted to call it out and see what others thought / if there was maybe a better way to approach this.

See the below small example:
```tsx
class DemoRoom<Options extends { a: true }> {
  constructor(options: Options) {}
}
new DemoRoom({ a: true, b: true }); // No error here
```

It probably makes sense why that extra `b: true` key is allowed in the above example, but in our case, this behavior is actually not optimal, because then it means extra keys can be specified for fixed token sources:
```typescript
type Base = { /* ... */ };
type Fixed = Base & {
  tokenSource: TokenSourceFixed;
};
type Configurable = Base & TokenSourceFetchOptions & {
  tokenSource: TokenSourceConfigurable;
};

class DemoRoom<Options extends Base | Fixed | Configurable> {
  constructor(options: Options) {}
}

const room = new DemoRoom({
  tokenSource: TokenSource.literal({ serverUrl: "...", participantToken: "..." }),
  // No error on the next line, because it goes down the `Base`
  // branch of `Options`, not the `Fixed` branch :(
  agentName: "agent to dispatch",
});
```

The only way I could think to fix it is by doing something like this:
```typescript
class DemoRoom<Options extends { a: true, b: never }> {
  constructor(options: Options) {}
}
new DemoRoom({ a: true, b: true }); // Now there is an error here
```

And that's what this `BlockTokenSourceFetchOptions` type is doing effectively, forcibly setting `never` on all the fetch option values so that it won't go down the `RoomOptionsLegacy` branch when they are set and shouldn't be there, and instead give a hard error.

src/room/Room.ts
```diff
@@ -610,9 +664,38 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
+    ? {
+        (opts?: RoomConnectOptions): Promise<void>;
+      }
+    : {
+        (url: string, token: string, opts?: RoomConnectOptions): Promise<void>;
+      } = async (
+    urlOrOptsOrUnset: string | RoomConnectOptions | undefined,
+    tokenOrUnset?: string,
+    optsOrUnset?: RoomConnectOptions,
+  ) => {
```
1egoman: For now I opted not to deprecate the legacy `Room.connect` / `Room.prepareConnection` paths. I think we should continue to discuss when this should happen though - maybe after a few months after it being fully released and in the docs we can reevaluate it.

src/room/Room.ts
1egoman: (Out of date, will make new comment on top of new diff)

src/options.ts
1egoman: (Out of date, will make new comment on top of new diff)

src/options.ts
1egoman: (Out of date, will make new comment on top of new diff)

src/room/Room.ts
1egoman: Decision: avoid the deprecation for now and merge this as is, then work with the docs team to update the docs, and then once that is done, do a follow up change to add the deprecation officially in.

src/room/Room.ts
```diff
@@ -125,7 +131,12 @@ const connectionReconcileFrequency = 4 * 1000;
...
  * @noInheritDoc
  */
-class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) {
+class Room<
+  RoomTokenSource extends TokenSourceFixed | TokenSourceConfigurable | null =
+    | TokenSourceFixed
+    | TokenSourceConfigurable
+    | null,
+> extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) {
   state: ConnectionState = ConnectionState.Disconnected;
```
1egoman, line 140: In order to properly implement the typescript interface for this change, I added a `RoomTokenSource` generic to `Room` with a default value, so it should be fully backwards compatible.

src/options.ts
```diff
@@ -101,10 +103,51 @@ export interface InternalRoomOptions {
...
+ * This type isn't necesarily required but helps guide users in an editor against accidentally
+ * including token source fetch options in the room constructor params for "fixed" token sources.
+ * Otherwise since the Room constructor `Options` generic "extends" RoomOptions, extra fields could
+ * be included.
+ */
+type BlockTokenSourceFetchOptions = { [P in keyof TokenSourceFetchOptions]: never };
+
+export type RoomOptionsLegacyOrTokenSourceFixed = RoomOptionsBase & BlockTokenSourceFetchOptions;
+export type RoomOptionsTokenSourceConfigurable = RoomOptionsBase & Partial<TokenSourceFetchOptions>;
+
```
1egoman, line 142: I got stuck on this for a while so I wanted to call it out and see what others thought / if there was maybe a better way to approach this.

See the below small example:
```tsx
class DemoRoom<Options extends { a: true }> {
  constructor(options: Options) {}
}
new DemoRoom({ a: true, b: true }); // No error here
```

It probably makes sense why that extra `b: true` key is allowed in the above example, but in our case, this behavior is actually not optimal, because then it means extra keys can be specified for fixed token sources:
```typescript
type Base = RoomOptions;
type LegacyOrFixed = Base;
type Configurable = Base & TokenSourceFetchOptions;

class DemoRoom<Options extends LegacyOrFixed> {
  constructor(_ignoredTokenSource: TokenSourceFixed, options: Options) {}
}

const room = new DemoRoom(
  TokenSource.literal({ serverUrl: "...", participantToken: "..." }),
  {
    // No error on the next line, because it "agentName" isn't
    // specified in "LegacyOrFixed", so a superobject of that
    // can include the key...  :(
    agentName: "agent to dispatch",
  }
);
```

The only way I could think to fix it is by doing something like this:
```typescript
class DemoRoom<Options extends { a: true, b: never }> {
  constructor(options: Options) {}
}
new DemoRoom({ a: true, b: true }); // Now there is an error here
```

And that's what this `BlockTokenSourceFetchOptions` type is doing effectively, forcibly setting `never` on all the fetch option values so that it won't go down the `RoomOptionsLegacyOrTokenSourceFixed` branch when they are set and shouldn't be there, and instead give a hard error.

pblazej (approved):
Nothing to add on the hi-level

src/options.ts
lukasIO: oh, I didn't realise we would now also pass token fetch options in the options. 

this wouldn't be necessary if we didn't do this, right?

1egoman: It wouldn't correct, but if we didn't then there wouldn't be a way to inject them from within the `useSession` level. Did you have some other mechanism in mind?

lukasIO: I imagined `useSession` would just call `fetch` on the token source

PR #1682, opened 2025-10-09
Single peer connection support

lukasIO:


src/api/SignalClient.ts
```diff
@@ -1063,3 +1077,31 @@ function createConnectionParams(
...
   return params;
 }
+
+function createJoinRequestConnectionParams(
+  token: string,
+  info: ClientInfo,
+  opts: ConnectOpts,
+): URLSearchParams {
+  const params = new URLSearchParams();
+  params.set('access_token', token);
```
boks1971, line 1087: Should we also move to `Bearer` token for this? I believe that is a pending item for clients.

lukasIO: unfortunately that's not supported from within the browser for WS connections.

src/api/SignalClient.ts
```diff
@@ -1063,3 +1077,31 @@ function createConnectionParams(
...
+  });
+  if (opts.reconnectReason) {
+    joinRequest.reconnectReason = opts.reconnectReason;
+  }
+  const wrappedJoinRequest = new WrappedJoinRequest({
+    joinRequest: joinRequest.toBinary(),
+  });
+  params.set('join_request', btoa(new TextDecoder('utf-8').decode(wrappedJoinRequest.toBinary())));
+
+  return params;
```
boks1971, line 1106: some potential optimisations for later
1. Can send add track requests and publisher offer in the initial connect request
2. Can also add a few m-lines each for audio and video gratuitously so that subscribes are faster

src/room/PCTransportManager.ts
```diff
@@ -117,11 +120,6 @@ export class PCTransportManager {
     this.updateState();
   }
 
-  requireSubscriber(require = true) {
```
boks1971, line 120: this is not needed anymore?

lukasIO: it was unused in the SDK, so I removed it

PR #1683, opened 2025-10-10
Add some token source utility function tests + fix some bugs

1egoman:
I happened to notice a bug in the token source token expiry checking logic that slipped through [the original pull request](https://github.com/livekit/client-sdk-js/pull/1645), so I fixed it and added some tests (which I really should have done in the first place 😬 ) to make sure this doesn't regress again.

### DIFF
--- a/src/room/token-source/utils.ts
+++ b/src/room/token-source/utils.ts
@@ -5,7 +5,7 @@ import type { RoomConfigurationObject, TokenPayload } from './types';
 const ONE_SECOND_IN_MILLISECONDS = 1000;
 const ONE_MINUTE_IN_MILLISECONDS = 60 * ONE_SECOND_IN_MILLISECONDS;
 
-export function isResponseExpired(response: TokenSourceResponse) {
+export function isResponseTokenValid(response: TokenSourceResponse) {
   const jwtPayload = decodeTokenPayload(response.participantToken);
   if (!jwtPayload?.nbf || !jwtPayload?.exp) {
     return true;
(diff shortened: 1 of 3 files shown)


src/room/token-source/utils.ts
```diff
@@ -7,14 +7,20 @@ const ONE_MINUTE_IN_MILLISECONDS = 60 * ONE_SECOND_IN_MILLISECONDS;
...
   const now = new Date();
-  return expiresAt >= now;
+
+  const nbfInMilliseconds = jwtPayload.nbf * ONE_SECOND_IN_MILLISECONDS;
+  const nbfDate = new Date(nbfInMilliseconds);
+
+  const expInMilliseconds = jwtPayload.exp * ONE_SECOND_IN_MILLISECONDS;
+  const expDate = new Date(expInMilliseconds - ONE_MINUTE_IN_MILLISECONDS);
+
+  const isValid = nbfDate <= now && expDate > now;
```
1egoman: The main bug was this `return expiresAt >= now;` line was backwards. It should have actually been `return expiresAt < now;`.

So, I fixed this and also at the same time added some logic to check the `nbf` field in the jwt to make sure that it's valid in the other direction.

src/room/token-source/utils.ts
```diff
@@ -5,16 +5,21 @@ import type { RoomConfigurationObject, TokenPayload } from './types';
...
   const now = new Date();
-  return expiresAt >= now;
+
+  const nbfInMilliseconds = jwtPayload.nbf * ONE_SECOND_IN_MILLISECONDS;
+  const nbfDate = new Date(nbfInMilliseconds);
+
+  const expInMilliseconds = jwtPayload.exp * ONE_SECOND_IN_MILLISECONDS;
+  const expDate = new Date(expInMilliseconds - ONE_MINUTE_IN_MILLISECONDS);
+
+  return nbfDate <= now && expDate > now;
```
xianshijing-lk, line 22: nit, I think it will be slightly cleaner if we use sec rather than ms, for instance: 

```
const nowSec = Date.now() / 1000;
const validFrom = jwtPayload.nbf;
const validUntil = jwtPayload.exp - 60; // 1 minute buffer
return nowSec >= validFrom && nowSec < validUntil;
```

PR #1687, opened 2025-10-14
Ensure leave requests can be sent before join response is received

lukasIO:
this PR ensures that on aborted connections we still send a LeaveRequest to the server (we could consider adding a new reason for this to not conflate `ClientInitiated` too much?). 

It also ensures that we don't have to wait for `joinResponse` to be present for a leave request to be initiated by assigning the streamWriter immediately after the ws connection is opened

### DIFF
--- a/src/api/SignalClient.test.ts
+++ b/src/api/SignalClient.test.ts
@@ -272,9 +272,6 @@ describe('SignalClient.connect', () => {
       // Now abort the connection (after WS opens, before join response)
       abortController.abort(new Error('User aborted connection'));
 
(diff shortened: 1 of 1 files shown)


src/api/SignalClient.test.ts
```diff
@@ -448,17 +539,6 @@ describe('SignalClient.handleSignalConnected', () => {
     }
   });
 
-  it('should set up stream writer from connection writable', () => {
```
lukasIO, line 451: deleting this as we're setting up the connection writable immediately on ws connection opened now

PR #1694, opened 2025-10-15
Fix the RPC comment and clamp the timeout if the provided value is less than 8s

xianshijing-lk:
So the problem is that the performRpc is a public function, and we never sanity check if users provide values larger than the connection_timeout, which might cause unexpected behavior. 

This PR fixes the comment and also clamp the connection response timeout if needed.

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -1792,13 +1792,7 @@ export default class LocalParticipant extends Participant {
 
       const effectiveTimeout = Math.max(responseTimeout, minEffectiveTimeout);
       const id = crypto.randomUUID();
-      await this.publishRpcRequest(
-        destinationIdentity,
-        id,
-        method,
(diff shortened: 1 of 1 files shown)


src/room/participant/LocalParticipant.ts
hiroshihorie, line 1814: Does this need to be `effectiveTimeout` instead of `responseTimeout` ? just making sure I could be wrong 😅  @lukasIO

bcherry: no - this one should be responseTimeout. effectiveTimeout may be poorly named - it's the time given to the _other side_ to run its handler. this side should still account for network roundtrips, but that side doesn't need to think at all about the network within the context of its handler

hiroshihorie: Cool, thanks for the clarification, I'm not sure why I got confused with this one 😅

PR #1696, opened 2025-10-15
Fix track mapping when single peer connectionis used

lukasIO:


src/room/Room.ts
```diff
@@ -1416,6 +1421,29 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
+        this.log.error(
+          `Tried to add a track whose 'sid' could not be found for a participant, that's not present. Sid: ${participantSid}`,
+          this.logContext,
+        );
+        return;
+      }
+
+      trackId = id;
+    }
+    if (!trackId.startsWith('TR')) {
```
lukasIO, line 1439: @boks1971  `addSubscribedMediaTrack` has its own logic for dealing with "unknown" trackIds. 

I think we should either remove the handling there altogether if we're comfortable with that (Firefox is mentioned as the motivation behind this best-effort-mapping, but I'll need to test) or let execution continue here and only log a warning. 

Any preference on your side?

boks1971: Okay either way. 

A warning here might be good as the code is all in one place.

lukasIO: sounds good, demoted to warning and removed the return

src/api/SignalClient.ts
```diff
@@ -797,6 +804,11 @@ export class SignalClient {
       if (this.onMediaSectionsRequirement) {
         this.onMediaSectionsRequirement(msg.value);
       }
+    } else if (msg.case === 'mappedAnswer') {
+      const sd = fromProtoSessionDescription(msg.value.sessionDescription!);
```
boks1971: I did the force `!` here as typescript was complaining. Is it better to check if it is undefined instead. I put the `!` in the `id` below also.

src/room/Room.ts
```diff
@@ -2541,6 +2568,13 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     if (event !== RoomEvent.ActiveSpeakersChanged && event !== RoomEvent.TranscriptionReceived) {
       // only extract logContext from arguments in order to avoid logging the whole object tree
       const minimizedArgs = mapArgs(args).filter((arg: unknown) => arg !== undefined);
+      if (event === RoomEvent.TrackSubscribed || event === RoomEvent.TrackUnsubscribed) {
+        this.log.trace(`subscribe trace: ${event}`, {
+          ...this.logContext,
+          event,
+          args: minimizedArgs,
+        });
+      }
```
boks1971, line 2577: do we still need this and the other `deferring on track for later` log ☝️

lukasIO: removed 👍

Issue #1699, opened 2025-10-19
Chrome 109 compatibility for Win7-8


thepacketloss:
### Describe the bug

The AbortSignal.any static method that I think added in this PR :

https://github.com/livekit/client-sdk-js/pull/1695/files

is not compatible with Chrome 109 (latest chrome version for Win 7-8) and only compatible with 116+

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#browser_compatibility

### Reproduction

Test entering meeting room on Win 7 or 8 with Chrome 109

### Logs

```shell

```

### System Info

```shell
livekit-client: ^2.15.11 => 2.15.11
```

### Severity

blocking an upgrade

PR #1700, opened 2025-10-20
Remove AbortSignal.any usage

lukasIO:
closes https://github.com/livekit/client-sdk-js/issues/1699
closes https://github.com/livekit/livekit/issues/4014

### DIFF
--- a/.changeset/twenty-spiders-joke.md
+++ b/.changeset/twenty-spiders-joke.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Remove AbortSignal.any usage


src/api/SignalClient.ts
```diff
@@ -325,7 +324,8 @@ export class SignalClient {
           reject(target instanceof AbortSignal ? target.reason : target);
         };
 
-        combinedAbort.addEventListener('abort', abortHandler);
+        timeoutAbortController.signal.addEventListener('abort', abortHandler);
+        abortSignal?.addEventListener('abort', abortHandler);
```
xianshijing-lk, line 325: [question](https://pullpo.io/cc?l=question) 

curiously, do we need to clean up the eventListeners at certain points ?

lukasIO: it's a good point, I will add this in

src/api/SignalClient.ts
```diff
@@ -301,13 +301,12 @@ export class SignalClient {
...
-        const signals = abortSignal
-          ? [timeoutAbortController.signal, abortSignal]
-          : [timeoutAbortController.signal];
-
-        const combinedAbort = AbortSignal.any(signals);
-
+        let alreadyAborted = false;
         const abortHandler = async (event: Event) => {
+          if (alreadyAborted) {
+            return;
```
1egoman, line 305: thought(if-minor): I wonder if there's a benefit to implementing more of a "ponyfill" type approach where there's a `function signalAny(...signals: Array<AbortSignal>): AbortSignal` type function that can be used here instead and also used in the future if we need similar type of functionality?

lukasIO: In my latest commit to address the listener cleanup I opted to remove one layer of indirection of the timeoutcontroller so that we end up with only one abortcontroller anyways. 
If we run into this again, I agree, a helper would be nice.

PR #1701, opened 2025-10-20
Update supported browser versions based on compatible API usage

lukasIO:


.browserslistrc
```diff
@@ -1,12 +1,12 @@
...
+safari >= 14.1
+ios_saf >= 14.5
+chrome >= 100
+and_chr >= 100
+android >= 100
+firefox >= 102
+and_ff >= 100
+edge >= 100
+Opera >= 100
+Samsung >= 15
```
1egoman, line 10: chore: I know you had said you made this change based on the apis that the client sdk is currently using in slack, but maybe it's worth digging a bit into the analytics and seeing what browsers / environments users are using this sdk in? I have had a longstanding todo list item to do this but it's been fairly low priority. But it worries me a bit to just wholesale update these without any data to back it up.

PR #1707, opened 2025-10-23
Only export types for TokenSource variants to try to push users to use the TokenSource.foo static constructors instead

1egoman:
Lukas suggested this and I think it's probably a good idea.


lukasIO (approved):
nice, thank you!

PR #1709, opened 2025-10-24
Cache region settings per project

lukasIO:


src/room/RegionUrlProvider.ts
```diff
@@ -3,16 +3,40 @@ import log from '../logger';
...
+
+type CachedRegionSettings = {
+  regionSettings: RegionSettings;
+  updatedAtInMs: number;
+  maxAgeInMs: number;
+};
+
 export class RegionUrlProvider {
-  private serverUrl: URL;
+  private static readonly cache: Map<string, CachedRegionSettings> = new Map();
```
lukasIO, line 15: Note: the cache currently never gets cleared. We could think about introducing some cleanup after x amount of time, but I don't expect this to be significant

src/room/RegionUrlProvider.ts
```diff
@@ -3,16 +3,45 @@ import log from '../logger';
...
+  private static scheduleRefetch(url: URL, token: string, maxAgeInMs: number) {
+    clearTimeout(this.settingsTimeout);
+    RegionUrlProvider.settingsTimeout = setTimeout(async () => {
+      try {
+        const newSettings = await fetchRegionSettings(url, token);
+        RegionUrlProvider.updateCachedRegionSettings(url, token, newSettings);
+      } catch (error: unknown) {
+        log.debug('auto refetching of region settings failed', { error });
+        // continue retrying with the same max age
+        RegionUrlProvider.scheduleRefetch(url, token, maxAgeInMs);
```
lukasIO: callout: on failed attempts we'll try again indefinitely with the same max age interval. Is there a good signal that we could use to stop trying?

src/room/RegionUrlProvider.ts
```diff
@@ -3,16 +3,45 @@ import log from '../logger';
...
+  updatedAtInMs: number;
+  maxAgeInMs: number;
+};
+
 export class RegionUrlProvider {
-  private serverUrl: URL;
+  private static readonly cache: Map<string, CachedRegionSettings> = new Map();
 
-  private token: string;
+  private static settingsTimeout: ReturnType<typeof setTimeout>;
```
xianshijing-lk: [question](https://pullpo.io/cc?l=question) 

I definitely don't know enough about the use case, so it is just one question, is it intentional to have a global settingsTimeout ? and allow scheduleRefetch to clear all the timeouts ? 

Or should it that we will need to have a map of host and timeout, and only clear the corresponding timeout ?

src/room/RegionUrlProvider.ts
```diff
@@ -3,16 +3,45 @@ import log from '../logger';
...
 
-  private regionSettings: RegionSettings | undefined;
+  private static scheduleRefetch(url: URL, token: string, maxAgeInMs: number) {
+    clearTimeout(this.settingsTimeout);
+    RegionUrlProvider.settingsTimeout = setTimeout(async () => {
```
1egoman: Nitpick: IMO pick a style (`this.` vs `RegionUrlProvider.`) and stick with it throughout the whole implementation rather than using both.

src/room/RegionUrlProvider.ts
1egoman: thought: I am not sure how fancy you would be willing to get here or how likely this is to fail in practice, but one option could be to do an exponential backoff (maybe `scheduleRefresh` could take an optional parameter like `previousErrorCount` to determine the current round's delay either via a lookup table or some computation?), which would mean at least you wouldn't have failures over and over rapidly.

src/room/RegionUrlProvider.ts
```diff
@@ -3,16 +3,45 @@ import log from '../logger';
...
 
+const DEFAULT_MAX_AGE_MS = 5_000;
+
+type CachedRegionSettings = {
+  regionSettings: RegionSettings;
+  updatedAtInMs: number;
+  maxAgeInMs: number;
+};
+
 export class RegionUrlProvider {
```
1egoman, line 14: question: Is there a situation where the old non caching implementation would be preferable over this new implementation? I don't know enough about the context around this change to know for sure. If there are pros and cons to each approach then maybe it should be configurable in some way?

src/room/RTCEngine.ts
```diff
@@ -603,7 +603,11 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       this.log.debug('client leave request', { ...this.logContext, reason: leave?.reason });
       if (leave.regions && this.regionUrlProvider) {
         this.log.debug('updating regions', this.logContext);
-        this.regionUrlProvider.setServerReportedRegions(leave.regions);
+        this.regionUrlProvider.setServerReportedRegions({
+          updatedAtInMs: Date.now(),
+          maxAgeInMs: 5_000,
```
1egoman: suggestion: Is it worth using the `DEFAULT_MAX_AGE_MS` constant here?


```suggestion
          maxAgeInMs: DEFAULT_MAX_AGE_MS,
```

src/room/RegionUrlProvider.ts
```diff
@@ -57,38 +96,37 @@ export class RegionUrlProvider {
...
+    RegionUrlProvider.updateCachedRegionSettings(this.serverUrl, this.token, settings);
   }
 }
 
 function getCloudConfigUrl(serverUrl: URL) {
   return `${serverUrl.protocol.replace('ws', 'http')}//${serverUrl.host}/settings`;
 }
+
+async function fetchRegionSettings(
+  serverUrl: URL,
```
1egoman, line 113: question: Is there a reason that this is not a private static method of `RegionUrlProvider`? Maybe because there's a need to call it elsewhere that I am missing?

src/room/RegionUrlProvider.ts
```diff
@@ -57,38 +96,37 @@ export class RegionUrlProvider {
     this.attemptedRegions = [];
   }
```
1egoman, line 101: In regards to:

> Note: the cache currently never gets cleared. We could think about introducing some cleanup after x amount of time, but I don't expect this to be significant

suggestion: This might be a good place to clear the cache if you wanted to add it somewhere (and maybe rename to `reset`? Or the current name kinda fits too), because looking at some of the rtc signaling code, it looks like this is called when a connection is explicitly restarted. I'm guessing that when that is done you'd probably want to force refetching if region settings at the same time?

1egoman (approved):
This generally looks good to me, and none of my notes are blocking - I've not yet had the chance to dig into this part of the code significantly yet so it's likely some of them don't really make sense.

src/room/RegionUrlProvider.ts
lukasIO: I actually had an exponential backoff in a previous commit, but removed it again as I wasn't sure whether it's the right strategy for this kind of endpoint. 
 
> failures over and over rapidly.

this shouldn't be the case anyways, we always keep at least the max-age time between attempts

src/room/RegionUrlProvider.ts
lukasIO: I thought I had addressed this, but yeah, makes perfect sense

src/room/RegionUrlProvider.ts
lukasIO: For restarts the cache might still be useful though?

I was more concerned with the cache staying in there even after all room connections have been disconnected for a long amount of time.

src/room/RegionUrlProvider.ts
lukasIO: I was thinking about this as well, we could have a timeout per project also. 
I think multiple projects in parallel would be a very rare edge case, but your suggestion is certainly the cleaner approach. I will add it in.

src/room/RegionUrlProvider.ts
lukasIO: Not a good reason, it's entirely self contained and not really related to the RegionUrlProvider class (it doesn't access any internals), so I left it outside in accordance with the getCloudConfigUrl method above.

src/room/RTCEngine.ts
lukasIO: thanks for the suggestion

src/room/RegionUrlProvider.ts
lukasIO: I cannot think of a situation. It's somewhat controllable by the server setting the max-age though

PR #1712, opened 2025-10-27
Add adaptive stream check

1egoman:
A customer reported that `RemoteVideoTrack.prototype.updateVisibility` was being errantly called when an associated room was not configured in `adaptiveStream` mode, causing videos to pause occasionally when stream state updates were being sent by the SFU. This change attempts to fix this problem.

Associated change where this functionality was added, for context: https://github.com/livekit/client-sdk-js/pull/1625/files#diff-9c68d93756d9f83946fa4be7c2ab17e138699737ff7d3752dab2f4619063bfcbR47

### DIFF
--- a/.changeset/rotten-rivers-own.md
+++ b/.changeset/rotten-rivers-own.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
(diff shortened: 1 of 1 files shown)


lukasIO (approved):
this makes sense, and thanks for the drive by logging fix!

PR #1714, opened 2025-10-29
Add exponential backoff strategy in case of connection failures

lukasIO:


src/room/RegionUrlProvider.ts
```diff
@@ -45,6 +45,18 @@ export class RegionUrlProvider {
           regionSettingsResponse.status,
         );
       }
+    } catch (e: unknown) {
+      if (e instanceof ConnectionError) {
+        // rethrow connection errors
+        throw e;
+      } else {
+        // wrap other errors as connection errors (e.g. timeouts)
+        throw new ConnectionError(
```
lukasIO, line 54: I initallially wanted to inspect error stati from these as well in order to only increase failedAttempts on specific status codes, but we don't reliably have status codes for the connection errors in the normal connection path. 

I left this in because it feels like a general improvement in terms of consistency.

lukasIO:
closing in favor of #1715

PR #1715, opened 2025-10-29
Use backoff strategy across projects and in between region attempts

lukasIO:
supersedes https://github.com/livekit/client-sdk-js/pull/1714

Key differences: 
- backs off also in between region attempts
- uses a project level map to store failed attempts across room instances
- starts the backoff timer at failure time so that attempts that start at a later time aren't penalised by the back off

### DIFF
--- a/src/room/BackOffStrategy.test.ts
+++ b/src/room/BackOffStrategy.test.ts
@@ -9,11 +9,8 @@ vi.mock('./utils', async () => {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     sleep: vi.fn((ms: number) => Promise.resolve()),
     extractProjectFromUrl: vi.fn((url: URL) => {
-      // Mock to extract project name from cloud URLs
-      if (url.hostname.includes('livekit.cloud')) {
-        return url.hostname.split('.')[0];
-      }
-      return null;
+      // @ts-ignore
+      return actual.extractProjectFromUrl(url);
     }),
   };
 });


src/room/BackOffStrategy.ts
```diff
@@ -0,0 +1,56 @@
...
+      projectName,
+      sleep(Math.min(CONNECTION_BACKOFF_MIN * Math.pow(2, failureCount), CONNECTION_BACKOFF_MAX)),
+    );
+  }
+
+  getBackOffPromise(urlString: string): Promise<void> {
+    const url = new URL(urlString);
+    const projectName = url && extractProjectFromUrl(url);
+    return (projectName && this.backOffPromises.get(projectName)) || Promise.resolve();
+  }
```
1egoman, line 59: suggestion: That whole `|| Promise.resolve()` bit really threw me for a loop and I had to read that through a second time to really get what that part was doing.

IMO, it would be much more readable to drop that and rework the method into something that abstracts the notion of promises a little better to downstream callers like the below:

```suggestion
  async sleepForBackOffDuration(urlString: string) {
    const url = new URL(urlString);
    const projectName = url && extractProjectFromUrl(url);
    const backoffPromise = projectName && this.backOffPromises.get(projectName);
    if (backoffPromise) {
      await backoffPromise;
    }
  }
```

src/room/BackOffStrategy.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export class BackOffStrategy {
+  private static _instance: BackOffStrategy | null = null;
+
+  private failedConnectionAttempts = new Map<string, number>();
+
+  private backOffPromises = new Map<string, Promise<void>>();
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  private constructor() {}
```
1egoman, line 29: question: Is there some reason why there needs an explicit constructor here? At least by eye it looks like you could probaby drop this and it would compile the same.

src/room/Room.ts
```diff
@@ -115,7 +116,7 @@ export enum ConnectionState {
   SignalReconnecting = 'signalReconnecting',
 }
 
-const connectionReconcileFrequency = 4 * 1000;
+const CONNECTION_RECONCILE_FREQUENCY = 4 * 1000;
```
1egoman: nitpick: It would be good to have a unit in this constant - I'm assuming this is in milliseconds?


```suggestion

const CONNECTION_RECONCILE_FREQUENCY_MS = 4 * 1000;
```

src/room/BackOffStrategy.ts
```diff
@@ -0,0 +1,56 @@
+import { extractProjectFromUrl, sleep } from './utils';
+
+const CONNECTION_BACKOFF_MIN = 500;
+const CONNECTION_BACKOFF_MAX = 15_000;
```
1egoman: nitpick: It would be good to have a unit in this constant - I'm assuming this is in milliseconds?


```suggestion
const CONNECTION_BACKOFF_MIN_MS = 500;
const CONNECTION_BACKOFF_MAX_MS = 15_000;
```

src/room/BackOffStrategy.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export class BackOffStrategy {
+  private static _instance: BackOffStrategy | null = null;
+
+  private failedConnectionAttempts = new Map<string, number>();
+
```
1egoman, line 25: suggestion: I think I understand why you moved this state into an external class (so you can share it across room attempts) but if you are going to manage this with a singleton, it might be worth copying the existing pattern in use in other singleton's in the project (ie, `DeviceManager`) - so something like:

```ts
export default class DeviceManager {
  private static instance?: DeviceManager;

  static mediaDeviceKinds: MediaDeviceKind[] = ['audioinput', 'audiooutput', 'videoinput'];

  static getInstance(): DeviceManager {
    if (this.instance === undefined) {
      this.instance = new DeviceManager();
    }
    return this.instance;
  }

  /* ... etc ... */
}
```

Alternatively, I wonder if a better way to do it could be to extract all the instance management logic out and just store an instance as a constant, so you are effectively just exposing an object containing functions to downstream consumers:
```ts
class BackOffStrategy { /* ... */ }

const backoffStrategy = new BackOffStrategy();
export default backoffStrategy;
```

src/room/BackOffStrategy.ts
```diff
@@ -0,0 +1,56 @@
+import { extractProjectFromUrl, sleep } from './utils';
+
+const CONNECTION_BACKOFF_MIN = 500;
+const CONNECTION_BACKOFF_MAX = 15_000;
+
+export class BackOffStrategy {
```
1egoman, line 21: suggestion: I think it would be helpful to add a docs comment to this class explaining where it is used and also making it clear that this only is active when connected to the cloud environment and not self hosted environments.

src/room/BackOffStrategy.ts
lukasIO: it's declaring the constructor as private so that instantiating the class by itself is not possible

src/room/BackOffStrategy.ts
lukasIO: Good call, I adopted the pattern used in `DeviceManager`

src/room/BackOffStrategy.ts
lukasIO: I'll make it a bit clearer by declaring the `backoffPromise` separately in the function body like in your example.

PR #1725, opened 2025-11-05
Ensure publication isn't attempted after timeout rejected the promise

lukasIO:
With deferred publishing (starting publishing before a room has fully connected) there was a case where the publish promise would get rejected (as expected) however internally the track would still get published, which seems unexpected.

also the underlying track wasn't stopped previously on failed publishing attempts

### DIFF
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -166,7 +166,7 @@ const appActions = {
     const room = new Room(roomOptions);
 
     startTime = Date.now();
-    // await room.prepareConnection(url, token);
+    await room.prepareConnection(url, token);
     const prewarmTime = Date.now() - startTime;
     appendLog(`prewarmed connection in ${prewarmTime}ms`);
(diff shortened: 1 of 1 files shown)


examples/demo/demo.ts
```diff
@@ -166,7 +166,7 @@ const appActions = {
     const room = new Room(roomOptions);
 
     startTime = Date.now();
-    await room.prepareConnection(url, token);
+    // await room.prepareConnection(url, token);
```
xianshijing-lk: curiously, is it intentionally to comment out this prepareConnection ? 

If yes, please add a comment to explain why you can't remove this line of code.

xianshijing-lk:
One question, lgtm

1egoman (approved):
Makes sense to me!

examples/demo/demo.ts
lukasIO: wasn't intentional, pushed the fix apparently while you were reviewing, sorry for that!

PR #1729, opened 2025-11-06
fix(e2ee): propagate worker decryption errors and reject corresponding promises

CSantosM:
### Summary

Fix unhandled promise rejections caused by missing error propagation in async encrypt/decrypt requests within the E2EE worker. Now, decryption failures correctly reject their promises, letting callers handle them (e.g. show fallback UI).

### Problem

When `DataCryptor.decrypt()` threw inside the worker, the exception was not caught. The main-thread promise from `E2EEManager.handleEncryptedData()` stayed pending and never rejected. This caused unhandled promise rejections (`"Uncaught (in promise)"`) and made it impossible for consumers to respond to decryption errors.

### Fix

This PR ensures proper rejection and error propagation for async encryption/decryption flows:

* Added an optional `uuid` field to `ErrorMessage` to tie worker errors to specific async requests.
* Caught errors in the worker’s `decryptDataRequest` and posted structured error messages (with `uuid`).
* Updated `E2EEManager` to reject the corresponding pending promise when receiving a matching error message.
* Emitted a general `EncryptionError` event only for errors not tied to a request.

### Result

Unhandled promise rejections are eliminated. Async request failures now reject cleanly and can be caught by consumers, restoring expected error handling behavior.

### DIFF
--- a/.changeset/gorgeous-llamas-whisper.md
+++ b/.changeset/gorgeous-llamas-whisper.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+fix(e2ee): propagate worker data decryption errors and reject corresponding promises


src/e2ee/E2eeManager.ts
```diff
@@ -144,7 +144,25 @@ export class E2EEManager
...
+
+          const encryptFuture = this.encryptDataRequests.get(data.uuid);
+          if (encryptFuture?.reject) {
+            encryptFuture.reject(data.error);
+            break; // Don't emit general error if it's handled by future
+          }
+        }
+
+        // Emit general error event for unhandled errors
+        this.emit(EncryptionEvent.EncryptionError, data.error);
```
lukasIO: this is missing the participant identity field now

src/e2ee/worker/e2ee.worker.ts
```diff
@@ -111,21 +111,37 @@ onmessage = (ev) => {
...
-          data: { payload: decryptedPayload, uuid: data.uuid },
-        } satisfies DecryptDataResponseMessage);
+        try {
+          const { payload: decryptedPayload } = await DataCryptor.decrypt(
+            data.payload,
+            data.iv,
+            getParticipantKeyHandler(data.participantIdentity),
+            data.keyIndex,
+          );
+          console.log('decrypted payload', {
```
lukasIO: I know you didn't introduce this, but the console log looks like an oversight, would be good to remove this log as well, thanks!

lukasIO (commented):
thank you for catching this! 
this makes sense, two small things

src/e2ee/E2eeManager.ts
CSantosM: My fault. Fixed!

PR #1731, opened 2025-11-07
Update release workflow to support trusted publishing

lukasIO:
see https://docs.npmjs.com/trusted-publishers


1egoman (approved):
Looks good by eye to me, is there a way to test it other than merging?

lukasIO:
Unfortunately not, it's going to be trial and error afaik

PR #1732, opened 2025-11-08
Use github_token for for changesets step

lukasIO:
the changesets command needs the github_token to be present,

related https://github.com/changesets/action/issues/525


1egoman (approved):
Assuming this a second change related to https://github.com/livekit/client-sdk-js/pull/1731, makes sense

PR #1744, opened 2025-11-13
Fixed an issue where createLocalAudioTrack would overwrite existing audio constraint flags

itamayo:
If deviceId is not specified in the constraints, other audio constraint flags (e.g. echoCancellation) get overwritten unexpectedly.

### DIFF
--- a/src/room/track/create.ts
+++ b/src/room/track/create.ts
@@ -78,16 +78,12 @@ export async function createLocalTracks(
       deviceId: { ideal: deviceId },
     };
   }
-  if (
-    internalOptions.audio === true ||
-    (typeof internalOptions.audio === 'object' && !internalOptions.audio.deviceId)
-  ) {
+  if (internalOptions.audio === true) {
+    internalOptions.audio = { deviceId: 'default' };
+  } else if (typeof internalOptions.audio === 'object' && internalOptions.audio !== null) {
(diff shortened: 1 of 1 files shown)


itamayo:
Take into account that deviceId is optional; the code should handle cases where other flags are defined but deviceId is not.

lukasIO (commented):
thanks for the PR! 

this makes sense, just one ask: 
we're already checking for `typeof internalOption.audio` in the conditional clause above, I think it would be a bit cleaner if we did something like 


```ts

 if (internalOptions.audio === true) {
    internalOptions.audio = { deviceId: 'default' };
} else if (typeof internalOptions.audio === 'object' && !internalOptions.audio.deviceId) {
 ...

```

itamayo:
Yes, I think what you’re proposing is cleaner. What do you think of the following code? If it looks good to you, I can update the PR accordingly. 
```ts
if (internalOptions.audio === true) {
  internalOptions.audio = { deviceId: 'default' };
} else if (typeof internalOptions.audio === 'object' && internalOptions.audio !== null) {
  internalOptions.audio = {
    ...internalOptions.audio,
    deviceId: internalOptions.audio.deviceId || 'default',
  };
}
```

lukasIO:
yeah, this looks good :)

PR #1745, opened 2025-11-14
Speed up network switch recovery

lukasIO:
Previously the SDK detected temporary disconnects via either
- WS connection state
- ping timeout
- peer connection state

All of these take a while to trigger, so in cases of network switching (e.g. WIFI -> tethering) the ping timeout was usually the first indication to trigger a resume.

By leveraging the browser's built in online listener, we can recover from these kind of network interruptions a lot quicker, making network changes feel a lot more seamless to users.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -53,7 +53,6 @@ import log, { LoggerNames, getLogger } from '../logger';
 import type { InternalRoomOptions } from '../options';
 import { DataPacketBuffer } from '../utils/dataPacketBuffer';
 import { TTLMap } from '../utils/ttlmap';
-import { BackOffStrategy } from './BackOffStrategy';
 import PCTransport, { PCEvents } from './PCTransport';
 import { PCTransportManager, PCTransportState } from './PCTransportManager';
 import type { ReconnectContext, ReconnectPolicy } from './ReconnectPolicy';


src/room/RTCEngine.ts
```diff
@@ -204,6 +204,9 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private midToTrackId: { [key: string]: string } = {};
 
+  /** used to indicate whether the browser is currently waiting to reconnect */
+  private isWaitingForNetworkReconnect: boolean = false;
```
boks1971, line 209: should we make this a connect option? My experience with this is very old. At some point, the browser detection was a bit flaky and probably triggered offline/online unnecessarily.

lukasIO: that's a good point. Do you happen to remember some of the flakier scenarios? What kind of situation would trigger this?


To make it more robust we could fire an arbitrary `HEAD` request to LK infrastructure when `offline` is triggered and only if that fails/doesn't resolve within a timeout, we set `isWaitingForNetworkReconnect` to true? 
That has the caveat that at that point in time the network interruption might have resolved already

lukasIO: there's also a newer (chromium only) API for [detecting network switches](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/change_event). I was hoping we could get broader support with the online listener

boks1971: I don't remember unfortunately. I think it might have been browser saying online when it was not really which is not bad, ie we attempt a reconnection unnecessarily. I do not remember if the other way happened where offline was triggered incorrectly.

We can probably keep this the default and provide a flag to disable if somebody does not want it.

Would be good to not rely on LK infra for OSS use cases.

lukasIO: I'm a tiny bit reluctant about adding it as a configurable option mainly because the config is such a beast already. 

I've added it in [c47d878](https://github.com/livekit/client-sdk-js/pull/1745/commits/c47d878a49f598c100f82ccee4ed0fd06bdb3eef) to `RoomOptions` as we don't have access to connect options directly within the RTCEngine and it felt a bit wrong to have it on `SignalOptions` instead.

boks1971: Agreed that the config is getting big. If you feel confident that it is working (and like I said my information is at least 6 - 7 years old and I am guessing browsers do much better nowadays), it is okay to not have the flag I think.

src/room/RTCEngine.ts
```diff
@@ -1619,23 +1623,62 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+
+  private handleBrowserOffline = async () => {
+    if (!this.url) {
+      return;
+    }
+    try {
+      await Promise.race([
+        fetch(toHttpUrl(this.url), { method: 'HEAD' }),
+        // if there's no internet connection the fetch rejects immediately, so we only use a short timeout here
+        sleep(4_000).then(() => Promise.reject()),
```
xianshijing-lk, line 1660: [nitpick](https://pullpo.io/cc?l=nitpick) 
nit, put in an error message in the reject function like 
Promise.reject(new Error('offline fetch timeout'));

lukasIO: the error message isn't used in the catch clause so, this wouldn't have any impact

PR #1747, opened 2025-11-14
Typesafe error propagation in signal connection path

lukasIO:


src/api/WebSocketStream.ts
```diff
@@ -26,11 +30,11 @@ export interface WebSocketStreamOptions {
...
   readonly url: string;
 
-  readonly opened: Promise<WebSocketConnection<T>>;
+  readonly opened: ResultAsync<WebSocketConnection<T>, WebSocketError>;
 
-  readonly closed: Promise<WebSocketCloseInfo>;
+  readonly closed: ResultAsync<WebSocketCloseInfo, WebSocketError>;
 
-  readonly close: (closeInfo?: WebSocketCloseInfo) => void;
+  readonly close!: (closeInfo?: WebSocketCloseInfo) => void;
```
1egoman, line 37: suggestion: if the plan is to eventually migrate to the in-built browser `WebsocketStream` (and from what I can see, it is on [some sort of](https://github.com/mozilla/standards-positions/issues/970) [standards track](https://github.com/whatwg/websockets/pull/48)), it might be worth treating this like vendored code that shouldn't be modified and doing these `ResultAsync` patches where this is being used in the signal client code.

I can think of pros and cons to both approaches though so I'd be curious to hear your rationale if there is a reason why doing this would be overly burdensome.

src/room/errors.ts
```diff
@@ -10,37 +10,148 @@ export class LivekitError extends Error {
...
+  reason: ConnectionErrorReason.ServerUnreachable;
+  status?: number;
+  context?: unknown;
+};
+
+type WebSocket = {
+  reason: ConnectionErrorReason.WebSocket;
   status?: number;
+  context?: string;
+};
```
1egoman, line 71: question: Just to make sure I am understanding this properly - in what case should `context` be `unknown`? I understand the `never` case (no value) and explicit types like `DisconnectReason`, but is there a reason the `unknown`s can't be more defined types?

src/room/errors.ts
```diff
@@ -111,6 +222,24 @@ export class SignalRequestError extends LivekitError {
...
+
+// export class AbortError extends LivekitError {
+//   readonly type = 'abort';
+
+//   constructor(message: string = 'AbortError') {
+//     super(18, message);
+//     this.name = 'AbortError';
+//   }
+// }
+
```
1egoman: nitpick: FYI that there is some dead code here

src/room/Room.ts
```diff
@@ -1915,18 +1917,13 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
       }
     } catch (error) {
       if (error instanceof RpcError) {
         responseError = error;
       } else {
-        console.warn(
-          `Uncaught error returned by RPC handler for ${method}. Returning APPLICATION_ERROR instead.`,
-          error,
-        );
         responseError = RpcError.builtIn('APPLICATION_ERROR');
```
1egoman, line 1940: issue: instead of getting rid of these warnings (which I am guessing was done because they use `console` and not the logger, maybe it would be better to swap them out for `log.warn`s? Unless there's something specific to the error handling change around why these were removed, I think they serve some value and would be good to keep around.

src/room/RTCEngine.ts
```diff
@@ -254,38 +256,21 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     token: string,
     opts: SignalOptions,
     abortSignal?: AbortSignal,
-  ): Promise<JoinResponse> {
+    // TODO: be more explicit about error types
+  ): Promise<Result<JoinResponse, Error>> {
```
1egoman: question: What is keeping you from replacing the `Error` in here with a union of errors? Are there places that are not immediately obvious from looking at the diff where non `LivekitError`-descended errors are being thrown that this must cover?

src/api/WebSocketStream.ts
lukasIO: Fair point. Was thinking about this too. I have currently less confidence into adapting the standard proposal any time soon. Given that, I think starting with the error propagation at the lowest level sounded like the right choice to me

src/room/errors.ts
lukasIO: We probably should be more specific, you're right. 
I think it's mostly the `internal` error that currently has different things set.
FWIW we don't use the context outside of logging currently

src/room/Room.ts
lukasIO: yeah, that was overly ambitious regex deletion of my debug logs 😬

src/room/RTCEngine.ts
lukasIO: Updated this to `ConnectionError`

src/room/Room.ts
lukasIO: brought the logs back with proper logging

src/room/errors.ts
lukasIO: done

src/api/utils.ts
```diff
@@ -49,3 +52,130 @@ export function getAbortReasonAsString(
...
+      try {
+        const result = await ra;
+        return result.match(
+          (value) => value,
+          (error) => {
+            throw error as Error;
+          },
+        );
+      } catch (error) {
+        await onFinally();
```
xianshijing-lk: Does it mean that onFinally() will be called twice if the initial ResultAsync ra fails ? 
One in catch() {}, another one in finally ?

src/api/utils.ts
```diff
@@ -49,3 +52,130 @@ export function getAbortReasonAsString(
...
+    return errAsync(ConnectionError.cancelled('AbortSignal invoked'));
+  }
+
+  const abortResult = ResultAsync.fromPromise(
+    new Promise<never>((_, reject) => {
+      const onAbortHandler = () => {
+        signal?.removeEventListener('abort', onAbortHandler);
+        reject(ConnectionError.cancelled('AbortSignal invoked'));
+      };
+      signal?.addEventListener('abort', onAbortHandler);
```
xianshijing-lk: will it leak this listener if  ra completes first ?

src/api/utils.ts
```diff
@@ -49,3 +52,130 @@ export function getAbortReasonAsString(
...
+  signal: AbortSignal | undefined,
+): ResultAsync<T, E | ReturnType<typeof ConnectionError.cancelled>> {
+  if (signal?.aborted) {
+    return errAsync(ConnectionError.cancelled('AbortSignal invoked'));
+  }
+
+  const abortResult = ResultAsync.fromPromise(
+    new Promise<never>((_, reject) => {
+      const onAbortHandler = () => {
+        signal?.removeEventListener('abort', onAbortHandler);
```
xianshijing-lk: nit, signal.removeEventListener('abort', onAbortHandler);

as it gets to this handler, so the eventListener must be valid

src/api/utils.ts
lukasIO: yes 🤦 
nice catch!

src/api/utils.ts
lukasIO: thanks, changed it so that it only listens to one event being emitted

src/api/utils.ts
lukasIO: addressed in the lower comment

src/api/SignalClient.ts
```diff
@@ -241,229 +242,174 @@ export class SignalClient {
...
+    const self = this;
 
-        const handleSignalConnected = (
-          connection: WebSocketConnection,
-          firstMessage?: SignalResponse,
-        ) => {
-          this.handleSignalConnected(connection, wsTimeout, firstMessage);
-        };
+    return withMutex(
+      safeTry<U, ConnectionError>(async function* () {
```
1egoman, line 277: suggestion(non-blocking): When I was first trying to parse through this I saw you swapped all `this` -> `self` and it took me a second to figure out exactly where `self` was being assigned and why you had done this. It looks like (from what I can tell) it's due to the generator function creating a new explicit `this` scope.

I wanted to propose another option instead - something like `async function* () {}.bind(this)`. There's some discussion here on the pros/cons: https://github.com/supermacro/neverthrow/issues/632. At least as somebody coming into reading this code for the first time, I think it would help but it's minor and something I don't have a strong opinion on.

src/api/SignalClient.ts
```diff
@@ -241,229 +242,174 @@ export class SignalClient {
...
+          }),
+          abortSignal,
+        ).orTee((error) => {
+          self.close(undefined, error.message);
+        });
+
+        const firstMessageOrClose = raceResults([
+          self.processInitialSignalMessage(wsConnection),
+          // Return the close promise as error if it resolves first
+          ws!.closed
```
1egoman: nitpick: It is possible I am missing something, but scanning through it looks like you might be able to get rid of the `!` because `const ws = ...` is assigned above and this should always exist.


```suggestion
          ws.closed
```

src/api/SignalClient.ts
```diff
@@ -1002,31 +964,34 @@ export class SignalClient {
    * @returns A ConnectionError with appropriate reason and status
    * @internal
    */
-  private async handleConnectionError(
+  private async fetchErrorInfo(
     reason: unknown,
     validateUrl: string,
-  ): Promise<ConnectionError> {
+  ): Promise<Result<never, ConnectionError>> {
```
1egoman, line 970: nitpick: I think this should be `ResultAsync` instead and the returns from this function updated to be `errAsync`?

UPDATE: reading a little further, this might actually be intended. I'm still leaving this comment though because I think it's relevant to a future comment.

src/api/utils.ts
```diff
@@ -49,3 +52,128 @@ export function getAbortReasonAsString(
...
+          (v) => Promise.resolve(v),
+          (err) => Promise.reject(err),
+        ),
+      ),
+  );
+
+  return ResultAsync.fromPromise(Promise.race(settledPromises), (e) => e as E);
+}
+
+export type ResultAsyncLike<T, E> = ResultAsync<T, E> | Promise<Result<T, E>>;
```
1egoman, line 179: question: Interesting - in my [last comment](https://github.com/livekit/client-sdk-js/pull/1747/files#r2546112423) I had assumed that `Promise<Result<T, E>>` was a mistake, but this line implies this is very much intentional. Is this a pattern because of async functions always returning promises, ie, the same type of topic as [this issue](https://github.com/supermacro/neverthrow/issues/42)?

Just thinking out loud but I wonder if handling this difference at this level of abstraction is the right place to do it (naively as somebody who hasn't really dug too much into `neverthrow` in depth yet, this semantic difference is surprising), or if it would be better to write a utility function which can convert any `Promise<Result<T, E>>` to a `ResultAsync<T, E>`.

I think I'd need to play around with it a bit on my own to develop a stronger opinion so feel free to keep it how it is for now and we can discuss it in the future once I've been able to dig in further.

src/room/RTCEngine.ts
```diff
@@ -1033,116 +1038,122 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
   }
 
-  private async restartConnection(regionUrl?: string) {
-    try {
-      if (!this.url || !this.token) {
+  private async restartConnection(
+    regionUrl?: string,
+  ): Promise<Result<void, UnexpectedConnectionState | SignalReconnectError>> {
+    const self = this;
+    const restartResultAsync = safeTry(async function* () {
```
1egoman, line 1082: suggestion(non-blocking): another instance of https://github.com/livekit/client-sdk-js/pull/1747/files#r2546082926 - consider maybe also using `.bind(this)` here.

1egoman (approved):
Looks good to me! Just generally, I think I'm realizing there's some fairly nuanced behavior with neverthrow that doesn't carry over 1:1 with how rust handles errors, so I think I'll need to spend some time digging into some of those nuances.

src/api/SignalClient.ts
lukasIO: Generally 50:50 on both, but reading through the issue you linked it mentions two steps being necessary for typescript today to properly bind it, i.e.: `return safeTry(function* (this: CurrentClass) {}).bind(this)` which would sway me to leave it as is.

src/api/utils.ts
lukasIO: yeah, it's on purpose but mostly as an intermediate step until we have ported all async internals to the result type. 

The helper function is a nice idea, we'd then have to have it include a generic `Error` though as the `Promise` could still through and would have to be unwrapped safely.

src/api/SignalClient.ts
lukasIO: nice catch!

src/api/utils.ts
lukasIO: > Is this a pattern because of async functions always returning promises, ie, the same type of topic as https://github.com/supermacro/neverthrow/issues/42?

yes, this was the reason and for the helpers I thought having a helper type to accept both would be nice

PR #1753, opened 2025-11-17
Add explicit error types for future helper

lukasIO:
This PR adds an error type to the `Future` type arguments that gets enforced when `future.reject` is called. 

For now all errors are just of type `Error` for the least amount of impact, but the plan is to further narrow down the possible error types with which a future can be rejected in follow ups and eventually replace it with a `ResultAsync<T, E>`

### DIFF
--- a/.changeset/shaggy-donuts-hammer.md
+++ b/.changeset/shaggy-donuts-hammer.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Add explicit error types for future helper


src/room/utils.ts
```diff
@@ -455,12 +455,12 @@ export function getStereoAudioStreamTrack() {
   return stereoTrack;
 }
 
-export class Future<T> {
+export class Future<T, E extends Error> {
```
1egoman, line 458: nitpick: is it worth making this `E extends Error = Error`? Or is the idea you want to ensure that all `Future`s have an `Error` specified in the future so adding the default like this would be counterproductive?

lukasIO: thought about this, too. I left it without the default in the end to force a conscious decision about which errors should be possible.

PR #1754, opened 2025-11-18
Control latency of lossy data channel

cnderrauber:
This PR introduces client-side changes to control latency for the lossy data channel (server change: https://github.com/livekit/livekit/pull/4088). The lossy data channel is affected by SCTP congestion control and can exhibit high latency when bandwidth is overused. For a lossy data channel, users typically care more about message latency than reliability. This change allows the sender (client or SFU edge) to proactively drop messages when the buffered queue delay (buffered bytes / current send rate) becomes too high, in order to keep latency under control.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -211,6 +211,8 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private lossyDataStatInterval: ReturnType<typeof setInterval> | undefined;
 
+  private lossyDataDropCount: number = 0;
+
   private midToTrackId: { [key: string]: string } = {};
 
   /** used to indicate whether the browser is currently waiting to reconnect */
@@ -363,6 +365,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       clearInterval(this.lossyDataStatInterval);
       this.lossyDataStatInterval = undefined;
     }
+    this.lossyDataDropCount = 0;
   }
 
   async cleanupClient() {
@@ -1324,6 +1327,13 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       } else {
         // lossy channel, drop messages to reduce latency
         if (!this.isBufferStatusLow(kind)) {
+          this.lossyDataDropCount += 1;
+          if (this.lossyDataDropCount % 100 === 0) {
+            this.log.warn(
+              `dropping lossy data channel messages, total dropped: ${this.lossyDataDropCount}`,
+              this.logContext,
+            );
+          }
           return;
         }
         this.lossyDataStatCurrentBytes += msg.byteLength;


src/room/RTCEngine.ts
```diff
@@ -1281,12 +1315,18 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-    await this.waitForBufferStatusLow(kind);
 
     const dc = this.dataChannelForKind(kind);
     if (dc) {
       if (kind === DataPacket_Kind.RELIABLE) {
+        await this.waitForBufferStatusLow(kind);
         this.reliableMessageBuffer.push({ data: msg, sequence: packet.sequence });
+      } else {
+        // lossy channel, drop messages to reduce latency
+        if (!this.isBufferStatusLow(kind)) {
```
cnderrauber, line 1329: @lukasIO I see you have change to wait for buffer status become low before send message(https://github.com/livekit/client-sdk-js/pull/1691), do you think it make sense to drop message for the lossy data channel instead of waiting the buffer is drained?

boks1971 (approved):
lgtm!

src/room/RTCEngine.ts
```diff
@@ -203,6 +205,12 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private reliableReceivedState: TTLMap<string, number> = new TTLMap(reliabeReceiveStateTTL);
 
+  private lossyDataStatCurrentBytes: number = 0;
+
+  private lossyDataStatByterate: number = 0;
+
+  private lossyDataStatTimer: ReturnType<typeof setTimeout> | undefined;
```
lukasIO:
```suggestion
  private lossyDataStatInterval: ReturnType<typeof setInterval> | undefined;
```

src/room/RTCEngine.ts
```diff
@@ -347,6 +356,15 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     this.reliableReceivedState.clear();
   }
 
+  cleanupLossyDataStats() {
+    this.lossyDataStatByterate = 0;
+    this.lossyDataStatCurrentBytes = 0;
+    if (this.lossyDataStatTimer) {
+      clearTimeout(this.lossyDataStatTimer);
```
lukasIO:
```suggestion
      clearInterval(this.lossyDataStatTimer);
```

src/room/RTCEngine.ts
lukasIO: The backpressure feels more transparent to users, but I don't have a strong opinion on which is better for lossy. 

I'd be curious to know how quickly and at what rate lossy messages would get dropped in practice on a somewhat congested network with this change.

cnderrauber: There are two changes: adaptive buffer threshold based on sender bitrate and dropping instead of waiting (backpressure).
The latency depends on the buffer threshold, with a fixed 64kB buffer threshold the demo can get over 2 sec latency if it produces message at higher rate than the bandwidth (256kb, 5%packet loss). And <500ms with the adaptive buffer threold (8kB minimum).
For waiting, I think the user need to implement a queue & dropping himself if he want to control the latency while the data source produces the message faster than network bandwidth, and we might need to describe the behavior in the document..

lukasIO: Yeah, waiting would be totally on the user but they would "know" about the latency. 

I think the changes in this PR make sense. Maybe we can add a warning log that the message rate is (too) high and messages are being dropped? 
Just to give users some sort of feedback around what's happening

PR #1757, opened 2025-11-19
Don't align buffer in isBufferStatusLow

cnderrauber:
align buffer during resuming will clear the
reliable message buffer and lost them after
resumed.


lukasIO (approved):
thank you 🎉

PR #1760, opened 2025-11-19
Upgrade eslint to v9 with newer recommended configs

lukasIO:
this PR adopts the new flat file config for eslint and uses the airbnb-typescript-extended plugin (as the old airbnb typescript one doesn't get updates anymore). 

With this change we now also get [linting with type information](https://typescript-eslint.io/getting-started/typed-linting). 
The only downside is that the linting job takes a bit longer (because TS is additionally running). 

I added a minimal set of ignored rules for now so that the lint job still passes with the current code base. We can then do smaller PRs to address individual rules.


eslint.config.mjs
```diff
@@ -0,0 +1,178 @@
...
+      'no-else-return': strictness,
+      'no-nested-ternary': strictness,
+      'no-promise-executor-return': strictness,
+      'prefer-const': strictness,
+      'prefer-exponentiation-operator': strictness,
+      'no-async-promise-executor': strictness,
+      'no-console': strictness,
+      'no-restricted-properties': strictness,
+      'no-undef-init': strictness,
+      ' no-irregular-whitespace': strictness,
```
1egoman: issue: Is this leading whitespace expected to be here?

```suggestion
      'no-irregular-whitespace': strictness,
```

1egoman (commented):
This generally makes sense to me!

Note that the tests aren't passing, at least in the most recent CI run as of this review it is because of the below (and it looks related to this change). Once the below is fixed I'll approve this!
```
> livekit-client@2.16.0 compat /home/runner/work/client-sdk-js/client-sdk-js
> eslint --no-eslintrc --config ./.eslintrc.dist.cjs ./dist/livekit-client.umd.js

Invalid option '--eslintrc' - perhaps you meant '--ext'?
You're using eslint.config.js, some command line flags are no longer available. Please see https://eslint.org/docs/latest/use/command-line-interface for details.
 ELIFECYCLE  Command failed with exit code 2.
```

eslint.config.mjs
lukasIO: there's a joke about `irregular-whitespace` here. 
Thanks!

PR #1761, opened 2025-11-20
Add check to make sure newly compiled dts versions work in downstream projects with older typescript versions

1egoman:
Runs a check to make sure that a given version of livekit-client can be used in a downstream project which is configured with an older version typescript.

In theory, `downlevel-dts` should handle this, but there has been at least one breaking typing change in the past which `downlevel-dts` has been unable to patch over (more info: https://github.com/livekit/client-sdk-js/pull/1668) so this serves as a mechanism to catch cases like this in the future before they are reported by users.

### DIFF
--- a/validate-dts-against-old-typescript.sh
+++ b/validate-dts-against-old-typescript.sh
@@ -33,7 +33,10 @@ cat <<EOF > tsconfig.json
       "DOM.Iterable",
       "ES2017",
       "ES2018.Promise",
-      "ES2021.WeakRef"
+      "ES2018.AsyncGenerator",
+      "ES2020.BigInt",
+      "ES2021.WeakRef",
+      "ESNext.AsyncIterable"
     ],
     "rootDir": "./",
     "outDir": "dist",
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-dts-against-old-typescript.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: pnpm/action-setup@v4
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: 'pnpm'
+      - name: Install dependencies
+        run: pnpm install
+      - name: Validates that livekit-client can be parsed in a typescript 4.8 project
+        run: pnpm validate-dts-against-old-typescript 4.8
```
1egoman: I'm not sure what the oldest supported typescript version actually should be - I tried `4.2` and this script failed (I think this was related to `import type` related syntax? But I am not 100% sure.). Just cherry picking more well known versions, `4.8` seemed to work so I went with it.

PR #1763, opened 2025-11-21
Typesafe errors in connection path

lukasIO:


src/room/RTCEngine.ts
```diff
@@ -1172,7 +1172,11 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     );
 
     if (reconnectResult.isErr()) {
-      return err(reconnectResult.error);
+      return err(
+        new SignalReconnectError(
```
lukasIO, line 1182: one of the errors was caused by this method returning a `ConnectionError` for some reason the return type  on the resumeConnection method isn't being type checked here it seems. 

This led to the reconnect flow triggering a "full reconnect" instead of a "resume" as "resumes" can only happen for SignalReconnectErrors

src/room/RTCEngine.ts
```diff
@@ -1119,7 +1119,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       self.client.setReconnected();
       self.emit(EngineEvent.SignalRestarted, joinResult.value);
 
-      await self.waitForPCReconnected();
+      yield* await self.waitForPCReconnected();
```
lukasIO, line 1128: the other error was caused by self.waitForPCReconnected throwing. 

This is the pitfall of doing a gradual adoption of this pattern. Extra care needed for ensuring that if a method is expected to not throw anymore it actually won't

lukasIO:
will try to break this up in smaller PRs, starting with #1764

PR #1766, opened 2025-11-24
Use result type for websocket stream

lukasIO:


src/api/SignalClient.ts
```diff
@@ -372,22 +372,30 @@ export class SignalClient {
...
-            .catch((reason) => {
               if (this.isEstablishingConnection) {
                 reject(
-                  ConnectionError.internal(
-                    `Websocket error during a (re)connection attempt: ${reason}`,
+                  ConnectionError.websocket(
+                    `Websocket got closed during a (re)connection attempt: ${closeInfo.reason}`,
+                  ),
+                );
+              } else if (this.state === SignalConnectionState.CONNECTED) {
```
lukasIO, line 382: This is the only expected behavioural change (a bug fix) that ensures the reconnect path is also initiated even if the ws closes with a `1000` code. 

This is needed to make some simulated scenarios work. The bug is also present currently on main.

src/api/WebSocketStream.test.ts
```diff
@@ -267,8 +297,11 @@ describe('WebSocketStream', () => {
       const wsStream = new WebSocketStream<ArrayBuffer | string>('wss://test.example.com');
 
       mockWebSocket.triggerOpen();
-      const connection = await wsStream.opened;
+      const result = await wsStream.opened;
+      expect(result.isOk()).toBe(true);
+      if (!result.isOk()) return;
 
+      const connection = result.value;
```
1egoman, line 304: nitpick: can this use `getConnectionOrFail`? It looks like right now if there were an error here line 302 would return without throwing it so it would get dropped.

src/api/WebSocketStream.ts
```diff
@@ -26,11 +30,11 @@ export interface WebSocketStreamOptions {
 export class WebSocketStream<T extends ArrayBuffer | string = ArrayBuffer | string> {
```
1egoman, line 30: suggestion(if-minor): since this is starting to diverge from the `WebSocketStream` spec, it might be worth making it clear in the docs that this is "inspired by" the spec. At least right now this reads like it's an implementation of the spec (like it used to be) which is a bit confusing given it's throwing application specific custom error types / using `neverthrow` / etc.

PR #1767, opened 2025-11-25
evaluate byethrow

lukasIO:


Karibash:
@lukasIO 
Hi! I noticed that you took the time to try byethrow in this PR — thank you very much for that 🙏
I also saw that the PR was eventually closed without being merged.

If you don’t mind sharing, I’d really appreciate any feedback on why it was closed (for example, API design, ergonomics, type complexity, or general fit for the project). Even brief thoughts would be very helpful for improving the library.

Thanks again for giving it a try, and I appreciate your time!

PR #1768, opened 2025-11-25
fix: remove listener from correct object

wuhkuh:
Stumbled upon this issue while debugging active Rooms retained in memory.
They were added to the `document`, but removed from the `window`.


lukasIO (approved):
great catch! 

Thanks for digging into this and fixing it

PR #1769, opened 2025-11-26
fix: add secure context checks for navigator.mediaDevices

yrendaoai:
# Fix: Support LiveKit Client in Non-Secure Context (HTTP Environment)

##  Problem

The original code directly calls `navigator.mediaDevices` APIs without checking for secure context. In HTTP environments (non-HTTPS), these APIs are restricted by browser security policies:

- `navigator.mediaDevices` may be `undefined`
- `addEventListener('devicechange')` fails even if `mediaDevices` exists
- `getUserMedia()` calls throw errors

This causes the WebRTC video player to fail in HTTP environments, breaking the live video streaming functionality.

##  Solution

- **Added secure context checks** before accessing `navigator.mediaDevices` APIs

### 1. Code Changes in Fixed PR
The fixed PR adds security context checks in `src/room/Room.ts`:

**Before:**
```typescript
navigator.mediaDevices?.addEventListener('devicechange', this.handleDeviceChange, {
  signal: abortController.signal,
});
```

**After:**
```typescript
if (window.isSecureContext && navigator.mediaDevices?.addEventListener) {
  navigator.mediaDevices.addEventListener('devicechange', this.handleDeviceChange, {
    signal: abortController.signal,
  });
}
```

Fixed locations:
-  Device change event listener registration (constructor)
-  Device change event listener removal (cleanup)
-  `getUserMedia()` calls for video tracks
-  `getUserMedia()` calls for audio tracks

##  Testing

-  Tested in HTTP environment - no errors when accessing `navigator.mediaDevices`
-  Tested in HTTPS environment - functionality works as expected
-  WebRTC video player works correctly in both secure and non-secure contexts
-  Device change events are properly handled when available

##  Related

This fix ensures compatibility with HTTP environments while maintaining full functionality in HTTPS environments. The changes are backward compatible and do not affect existing HTTPS deployments.

##  Notes

- The fixed PR maintains the same API as the original `livekit-client`
- All features work identically in secure contexts
- In non-secure contexts, device change events are gracefully skipped
- `getUserMedia()` calls fall back to dummy tracks when not available

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -2498,10 +2498,9 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
           name: 'video-dummy',
         }),
         new LocalVideoTrack(
-          publishOptions.useRealTracks
+          publishOptions.useRealTracks && window.navigator.mediaDevices?.getUserMedia
             ? (
-                (await window.navigator.mediaDevices?.getUserMedia?.({ video: true })) ??
-                new MediaStream()
+                await window.navigator.mediaDevices.getUserMedia({ video: true })
               ).getVideoTracks()[0]
             : createDummyVideoStreamTrack(
                 160 * (participantOptions.aspectRatios[0] ?? 1),
@@ -2528,10 +2527,8 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
           type: TrackType.AUDIO,
         }),
         new LocalAudioTrack(
-          publishOptions.useRealTracks
-            ? (
-                (await navigator.mediaDevices?.getUserMedia?.({ audio: true })) ?? new MediaStream()
-              ).getAudioTracks()[0]
+          publishOptions.useRealTracks && navigator.mediaDevices?.getUserMedia
+            ? (await navigator.mediaDevices.getUserMedia({ audio: true })).getAudioTracks()[0]
             : getEmptyAudioStreamTrack(),
           undefined,
           false,


src/room/Room.ts
```diff
@@ -276,9 +276,11 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       const abortController = new AbortController();
 
       // in order to catch device changes prior to room connection we need to register the event in the constructor
-      navigator.mediaDevices?.addEventListener('devicechange', this.handleDeviceChange, {
-        signal: abortController.signal,
-      });
+      if (window.isSecureContext && navigator.mediaDevices?.addEventListener) {
+        navigator.mediaDevices.addEventListener('devicechange', this.handleDeviceChange, {
```
lukasIO: we use optional chaining for this in other places already, I think this could be used here as well with the same effect as explicitly checking for secureContext?
```suggestion
        navigator.mediaDevices?.addEventListener('devicechange', this.handleDeviceChange, {
```

you mention

> addEventListener('devicechange') fails even if mediaDevices exists

under which circumstances is this the case?

lukasIO (commented):
thanks for the PR and the detailed description!

src/room/Room.ts
yrendaoai: Hi! Thanks for the question — here’s the context in which addEventListener('devicechange') fails even when navigator.mediaDevices exists.

In my case:

The page was running over HTTP (insecure context).

In this environment, navigator.mediaDevices does exist, but it only exposes methods like getUserMedia.
image

However, mediaDevices does not implement addEventListener, so calling it results in:

TypeError: navigator.mediaDevices.addEventListener is not a function


This matches the spec behavior: devicechange and event handling on mediaDevices are only guaranteed in secure contexts (HTTPS). On HTTP, the object may be present but not fully functional.

So the failure happens specifically when:
- navigator.mediaDevices exists
- but the browser/environment doesn’t provide the event interface (e.g., insecure context)

I tested this in:

Chrome Version 142.0.7444.176 (Official Build) (64-bit)

React 18.2.0 (though the framework isn't relevant but who knows; it's a browser API behavior)

Let me know if you'd like me to test in other environments.

yrendaoai: For the change suggestion, I got a safer approach would be:

navigator.mediaDevices?.addEventListener?.('devicechange', this.handleDeviceChange, 

This ensures we only call the handler when both the object and the method are available, which matches Chrome browser behavior in non-secure contexts.

Happy to make a PR if this approach sounds reasonable!

lukasIO: that approach sounds great, thanks!

src/room/Room.ts
```diff
@@ -2499,8 +2499,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
         }),
         new LocalVideoTrack(
           publishOptions.useRealTracks
-            ? (
-                await window.navigator.mediaDevices.getUserMedia({ video: true })
+            ? (await window.navigator.mediaDevices?.getUserMedia?.({ video: true }) ?? new MediaStream()
               ).getVideoTracks()[0]
```
yrendaoai, line 2504: Note: The `getUserMedia` calls with `?? new MediaStream()` fallback (lines 2502, 2529) 
ensure we don't get undefined when calling `.getVideoTracks()[0]` in non-secure contexts 
where `mediaDevices?.getUserMedia?.()` returns undefined due to optional chaining.

lukasIO: instead of falling back to `new MediaStream()` we could simply check for `publishOptions.useRealTracks && window.navigator.mediaDevices?.getUserMedia` above

src/room/Room.ts
```diff
@@ -2533,9 +2527,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
         }),
         new LocalAudioTrack(
           publishOptions.useRealTracks
-            ? (window.isSecureContext && navigator.mediaDevices
-                ? await navigator.mediaDevices.getUserMedia({ audio: true })
-                : new MediaStream()
+            ? (await navigator.mediaDevices?.getUserMedia?.({ audio: true }) ?? new MediaStream()
```
lukasIO: same as above

yrendaoai:
@lukasIO Thanks for the feedback! I've updated the PR to use optional chaining as suggested and moved the condition checks upfront for getUserMedia calls. Please take a look when you have a chance.

PR #1775, opened 2025-12-02
Fix connection check issues with pinned regions

lukasIO:


src/connectionHelper/checks/turn.ts
```diff
@@ -7,6 +9,10 @@ export class TURNCheck extends Checker {
   }
 
   async perform(): Promise<void> {
+    if (isCloud(new URL(this.url))) {
+      this.url =
+        (await new RegionUrlProvider(this.url, this.token).getNextBestRegionUrl()) ?? this.url;
+    }
```
1egoman, line 16: question: Is there value in trying to move this type of logic up to the `src/connectionHelper/ConnectionCheck.ts` level, so that it will apply to all checks? Or is there a reason this logic just needs to only apply to the `TURNCheck` and `WebsocketCheck`?

src/connectionHelper/checks/turn.ts
```diff
@@ -7,6 +9,10 @@ export class TURNCheck extends Checker {
   }
 
   async perform(): Promise<void> {
+    if (isCloud(new URL(this.url))) {
+      this.url =
```
1egoman, line 14: suggestion(non-blocking): Just like the other modification you made to the `WebsocketCheck`, is there value in logging here to make it clear that a different url is being used?

ie, something like:
```suggestion
    if (isCloud(new URL(this.url))) {
      this.appendMessage('Using region specific url');
      this.url =
```

src/connectionHelper/checks/turn.ts
lukasIO: those are the only two that construct their own `SignalClient` instance. 

it would be possible to override the url on the test itself, I opted against that to give some more transparency around what is happening in the websocket test

src/connectionHelper/checks/turn.ts
lukasIO: makes sense

src/connectionHelper/checks/websocket.ts
```diff
@@ -13,16 +15,43 @@ export class WebSocketCheck extends Checker {
...
+        const regionUrl = await regionProvider.getNextBestRegionUrl();
+        if (regionUrl) {
+          joinRes = await signalClient.join(regionUrl, this.token, {
+            autoSubscribe: true,
+            maxRetries: 0,
+            e2eeEnabled: false,
+            websocketTimeout: 15_000,
+            singlePeerConnection: false,
+          });
+          this.appendMessage(
```
xianshijing-lk, line 42: [thought](https://pullpo.io/cc?l=thought) 
just one comment, 
if the above joinRes = await signalClient.join(regionUrl, this.token {}) throws again inside the regionUrl, you will not get the message to append since the code is inside a catch, but the caller will still gets an exception. 

I would think it is OK, but just double check with you.

src/connectionHelper/checks/websocket.ts
```diff
@@ -13,16 +15,43 @@ export class WebSocketCheck extends Checker {
...
+            `Fallback to region worked. To avoid initial connections failing, ensure you're calling room.prepareConnection() ahead of time`,
+          );
+          return;
```
xianshijing-lk: is it intentional to have the return here ? 

Note, with this return, the following code won't run:
    if (joinRes) {
      this.appendMessage(`Connected to server, version ${joinRes.serverVersion}.`);
      if (joinRes.serverInfo?.edition === ServerInfo_Edition.Cloud && joinRes.serverInfo?.region) {
        this.appendMessage(`LiveKit Cloud: ${joinRes.serverInfo?.region}`);
      }
    }
    await signalClient.close();

src/connectionHelper/checks/websocket.ts
lukasIO: the `Checker` class wraps this and outputs the error, so this should be fine

src/connectionHelper/checks/websocket.ts
lukasIO: thanks! fixed

PR #1781, opened 2025-12-22
Add new rtc path and handle legacy server handling

lukasIO:


src/api/SignalClient.ts
```diff
@@ -995,10 +997,22 @@ export class SignalClient {
...
+          return ConnectionError.serviceNotFound(
+            'v0 RTC path not found. Consider upgrading your LiveKit server version',
+            'v0-rtc',
+          );
+        case 401:
+        case 403:
+          const msg = await resp.text();
+          return ConnectionError.notAllowed(msg, resp.status);
+        default:
+          break;
```
1egoman, line 1012: question: The old code looks like it was checking any 4xx error for this case. Is it possible that the validation url could return a 4xx other than 401/403 and should anything happen in that case to keep the preexisting behavior?

lukasIO: the current codes returned by the validation logic can be seen here https://github.com/livekit/livekit/blob/4104b8270ba8c653f80a27737bd125f06ef0e06d/pkg/service/utils.go#L245-L246

Today there's 400 and 401. 
Returning `notAllowed` for a 400 error doesn't feel quite right. Surfacing the underlying status code instead feels a bit more obvious.

xianshijing-lk (approved):

PR #1785, opened 2026-01-05
Add new rtc path with implicit single peer connection mode

lukasIO:


src/api/SignalClient.ts
```diff
@@ -995,10 +997,22 @@ export class SignalClient {
...
       const resp = await fetch(validateUrl);
-      if (resp.status.toFixed(0).startsWith('4')) {
-        const msg = await resp.text();
-        return ConnectionError.notAllowed(msg, resp.status);
-      } else if (reason instanceof ConnectionError) {
+
+      switch (resp.status) {
+        case 404:
+          return ConnectionError.serviceNotFound(
+            'v0 RTC path not found. Consider upgrading your LiveKit server version',
```
boks1971: should this be `v1` path not found?

lukasIO: thanks!

PR #1789, opened 2026-01-09
e2ee: ensure frame cryptor transform setup works for rapid subscription changes

lukasIO:


src/e2ee/worker/FrameCryptor.race.test.ts
```diff
@@ -0,0 +1,546 @@
...
+    keyProviderOptions,
+    sifTrailer: new Uint8Array(),
+  });
+
+  return { cryptor, keys };
+}
+
+describe('FrameCryptor Race Conditions', () => {
+  beforeEach(() => {
+    vitest.useFakeTimers();
```
xianshijing-lk, line 63: [nitpick](https://pullpo.io/cc?l=nitpick) 

I think it is better to call vitest.setSystemTime(new Date('2026-01-01T00:00:00Z')); here

xianshijing-lk (approved):
some nits. lgtm

src/e2ee/worker/FrameCryptor.ts
```diff
@@ -185,39 +195,66 @@ export class FrameCryptor extends BaseFrameCryptor {
...
+    // Always update trackId, even on reuse
+    this.trackId = trackId;
+
+    // If we're reusing and have an active transform skip setup
+    if (isReuse && this.currentTransform) {
+      workerLogger.debug('reusing existing transform', {
         ...this.logContext,
+        trackId,
       });
       return;
```
xianshijing-lk, line 217: [question](https://pullpo.io/cc?l=question) 
question, does it matter if the underlying streams match ? 
I wonder if you should check things like 
if (this.currentTransform.readable === readable && this.currentTransform.writable === writable) { return; }

lukasIO: that's a good idea as an additional safety check! I'll add that

src/e2ee/worker/FrameCryptor.race.test.ts
lukasIO: Interesting, can you elaborate on why?
I'll merge it in as is but happy to follow up with that change

PR #1792, opened 2026-01-22
Data tracks - Packet serialization + deserialialization, plus some error handling changes

1egoman:
Implements binary serialization and deserialization for data track packets. For now, this is being introduced in isolation to reduce the review burden, and further data tracks work will be done in follow up pull requests which build on top of this.

Included in this change also is some relatively early stage error handling infrastructure to attempt to further iterate on [the](https://github.com/livekit/client-sdk-js/pull/1747) [typesafe](https://github.com/livekit/client-sdk-js/pull/1758) [error](https://github.com/livekit/client-sdk-js/pull/1763) [handling](https://github.com/livekit/client-sdk-js/pull/1766) [pull requests](https://github.com/livekit/client-sdk-js/pull/1770) that lukas has been working on. I think we both still aren't really satisfied with the current state of things (namely, `TypedPromise` doesn't really support synchronous results, and introducing a new separate synchronous`Result` type of some sort makes interop between these two types a challenge). A LLM suggested we explore this `Throws` branded type approach (see `./throws-transformer/README.md` for more info about how it works) which seems to have some fairly early promise, so for now I've opted to build the data tracks error handling logic on top. That is part of what is resulting in the diff size of this change being so large 😬 .

## Summary by CodeRabbit

* **New Features**
  * Full data-track packet system: header/packet (de)serialization, extensions (E2EE, user timestamp), timestamps, and JSON output.

* **Tools**
  * Throws-check tooling: CLI, language-service plugin, transformer, README, and CI validation step added.

* **Errors & Utilities**
  * Typed error framework, serialization errors, wrap-around integers, timestamp utilities, and serialization base class.

* **Tests**
  * Expanded unit tests covering packets, extensions, handles, and wrap-around behavior.

* **Chores**
  * New package script and added tooling dependency; tsconfig plugin registration updated.

### DIFF
--- a/src/room/data-track/packet/extensions.ts
+++ b/src/room/data-track/packet/extensions.ts
@@ -41,7 +41,8 @@ export class DataTrackUserTimestampExtension extends DataTrackExtension {
     dataView.setUint16(byteIndex, DataTrackUserTimestampExtension.tag);
     byteIndex += U16_LENGTH_BYTES;
 
-    dataView.setUint16(byteIndex, DataTrackUserTimestampExtension.lengthBytes - 1);
+    const rtpOrientedLength = DataTrackUserTimestampExtension.lengthBytes - 1;
+    dataView.setUint16(byteIndex, rtpOrientedLength);
     byteIndex += U16_LENGTH_BYTES;
 
     dataView.setBigUint64(byteIndex, this.timestamp);
@@ -97,7 +98,8 @@ export class DataTrackE2eeExtension extends DataTrackExtension {
     dataView.setUint16(byteIndex, DataTrackE2eeExtension.tag);
     byteIndex += U16_LENGTH_BYTES;
 
-    dataView.setUint16(byteIndex, DataTrackE2eeExtension.lengthBytes - 1);
+    const rtpOrientedLength = DataTrackE2eeExtension.lengthBytes - 1;
+    dataView.setUint16(byteIndex, rtpOrientedLength);
     byteIndex += U16_LENGTH_BYTES;
 
     dataView.setUint8(byteIndex, this.keyIndex);
@@ -196,7 +198,8 @@ export class DataTrackExtensions extends Serializable {
       const tag = dataView.getUint16(byteIndex);
       byteIndex += U16_LENGTH_BYTES;
 
-      const lengthBytes = dataView.getUint16(byteIndex);
+      const rtpOrientedLength = dataView.getUint16(byteIndex);
+      const lengthBytes = rtpOrientedLength + 1;
       byteIndex += U16_LENGTH_BYTES;
 
       if (tag === EXT_TAG_PADDING) {
@@ -210,7 +213,7 @@ export class DataTrackExtensions extends Serializable {
             throw DataTrackDeserializeError.malformedExt(tag);
           }
           userTimestamp = new DataTrackUserTimestampExtension(dataView.getBigUint64(byteIndex));
-          byteIndex += lengthBytes + U8_LENGTH_BYTES;
+          byteIndex += lengthBytes;
           break;
 
         case DataTrackExtensionTag.E2ee:


src/room/data-track/packet/extensions.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  static lengthBytes: number;
+}
+
+export class DataTrackUserTimestampExtension extends DataTrackExtension {
+  tag = DataTrackExtensionTag.UserTimestamp;
+
+  lengthBytes = 8;
+
+  private timestamp: bigint;
```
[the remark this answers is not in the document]
1egoman: This is out of date, static properties are in use everywhere for `DataTrackUserTimestampExtension` and `DataTrackE2eeExtension`.

src/room/data-track/packet/extensions.ts
```diff
@@ -0,0 +1,244 @@
...
+
+        default:
+          // Skip over unknown extensions (forward compatible).
+          if (dataView.byteLength - byteIndex < lengthBytes) {
+            throw DataTrackDeserializeError.malformedExt(tag);
+          }
+          byteIndex += lengthBytes + U8_LENGTH_BYTES;
+          break;
+      }
+    }
```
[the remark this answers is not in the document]
1egoman: This is fine, the padding bytes are bogus and can be skipped over.

src/room/data-track/packet/index.ts
```diff
@@ -0,0 +1,333 @@
...
+      input instanceof DataView
+        ? input
+        : new DataView(input instanceof ArrayBuffer ? input : input.buffer);
+
+    const [header, headerByteLength] = DataTrackPacketHeader.fromBinary(dataView);
+
+    const payload = dataView.buffer.slice(
+      dataView.byteOffset + headerByteLength + U8_LENGTH_BYTES,
+      dataView.byteLength,
+    );
```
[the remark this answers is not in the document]
1egoman: No this is expected, there needs to be 1 byte added to make sure the payload starts after the header and doesn't include the last header byte as part of the payload.

tsconfig.json
```diff
@@ -24,7 +25,8 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "verbatimModuleSyntax": true,
-    "ignoreDeprecations": "5.0"
+    "ignoreDeprecations": "5.0",
+    "plugins": [{ "name": "./throws-transformer/plugin.ts" }],
```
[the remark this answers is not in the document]
1egoman: I made the change in the other direction, so this should be addressed.

throws-transformer/engine.ts
```diff
@@ -0,0 +1,818 @@
...
+      const rejectResult = checkPromiseReject(node, sourceFile, checker);
+      if (rejectResult) results.push(rejectResult);
+    }
+
+    // Check await expressions
+    if (ts.isAwaitExpression(node) && ts.isCallExpression(node.expression)) {
+      const result = checkCallExpression(node.expression, sourceFile, checker);
+      if (result) results.push(result);
+    }
+
```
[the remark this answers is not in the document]
1egoman: No this isn't correct, in one place `checkCallExpression` is being called on `node`, and the other `checkCallExpression` is being called on `node.expression`.

throws-transformer/engine.ts
```diff
@@ -0,0 +1,818 @@
...
+      line: line + 1,
+      column: character + 1,
+      start,
+      length,
+
+      functionName: "Promise.reject",
+      unhandledErrors: [rejectedTypeName],
+      message: `Promise.reject('${rejectedTypeName}') but it's not declared. Declared: ${declaredNames || "never"}. Add it to Throws<> in your return type.`,
+    };
+  }
```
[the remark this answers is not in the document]
1egoman: There are a lot of edge cases this script doesn't handle right now, `new Promise` constructors being another one. If we're going to invest into this `Throws` branded type thing long term some work needs to be done to make this more comprehensive.

src/room/data-track/utils.test.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { WrapAroundUnsignedInt } from './utils';
+
+describe('WrapAroundUnsignedInt', () => {
+  it('should test initialization + edge cases', () => {
+    expect(WrapAroundUnsignedInt.u16(1).value).toEqual(1);
+    expect(WrapAroundUnsignedInt.u16(196_606 /* (3 * 65535) + 1 */).value).toEqual(65534);
+    expect(() => WrapAroundUnsignedInt.u16(-1).value).toThrow(
```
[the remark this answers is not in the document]
1egoman: IMO, the suggestion here is even more confusing...

src/room/data-track/packet/errors.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+export type DataTrackDeserializeErrorAll =
+  | DataTrackDeserializeError<DataTrackDeserializeErrorReason.TooShort>
+  | DataTrackDeserializeError<DataTrackDeserializeErrorReason.HeaderOverrun>
+  | DataTrackDeserializeError<DataTrackDeserializeErrorReason.MissingExtWords>
+  | DataTrackDeserializeError<DataTrackDeserializeErrorReason.UnsupportedVersion>
+  | DataTrackDeserializeError<DataTrackDeserializeErrorReason.InvalidHandle>
+  | DataTrackDeserializeError<DataTrackDeserializeErrorReason.MalformedExt>;
```
1egoman, line 82: Context on this: The `Throws` branded type script isn't smart enough to be able to realize that a `CustomError<ReasonEnum.one> | CustomError<ReasonEnum.two> | ...` across all enum reasons is equal to a `CustomError<ReasonEnum>`. So I have to enumerate all the errors in cases where I want a function to "return any `CustomError` type".

I think this could be fixed longer term if desired but I also will say I don't really mind it - explicitly calling out all the cases I think has some value and makes it harder to make the `Throws` type brand have extra errors which aren't being raised.

src/room/data-track/packet/extensions.ts
```diff
@@ -0,0 +1,253 @@
...
+      dataView.setUint8(byteIndex, this.iv[i]);
+      byteIndex += U8_LENGTH_BYTES;
+    }
+
+    const totalLengthBytes = this.toBinaryLengthBytes();
+    if (byteIndex !== totalLengthBytes) {
+      throw new Error(
+        `DataTrackE2eeExtension.toBinaryInto: Wrote ${byteIndex} bytes but expected length was ${totalLengthBytes} bytes`,
+      );
+    }
```
1egoman, line 119: Worth noting: I have adopted the pattern throughout this code (and am thinking if we keep using this `Throws` pattern it would be worth adopting everywhere) of throwing a regular error being a "panic" level error / assertion that should never happen unless some built in invariant is broken.

src/room/data-track/packet/index.ts
```diff
@@ -0,0 +1,343 @@
...
+    byteIndex += U16_LENGTH_BYTES;
+    dataView.setUint32(byteIndex, this.timestamp.asTicks());
+    byteIndex += U32_LENGTH_BYTES;
+
+    if (extensionsLengthBytes > 0) {
+      // NOTE: the below value shouldn't have one subtracted from it.
+      //
+      // There is an off by one bug in the SFU logic which to fix would require making a
+      // backwards incompatible change, so he decision was made to just subtract one here instead.
+      const hackedExtensionLengthWords = extensionsLengthWords - 1;
```
1egoman: Wanted to surface this because it is weird / unusual, this was a change copied from the [rust implementation here](https://github.com/livekit/rust-sdks/commit/02114d80dadbb8bb0e5cf82a3ffdb6aca31805ec).

src/room/data-track/packet/index.ts
```diff
@@ -0,0 +1,343 @@
...
+      if (
+        e instanceof DataTrackHandleError &&
+        (e.isReason(DataTrackHandleErrorReason.Reserved) ||
+          e.isReason(DataTrackHandleErrorReason.TooLarge))
+      ) {
+        throw DataTrackDeserializeError.invalidHandle(e);
+      } else {
+        throw e;
+      }
+    }
```
1egoman, line 221: This is an example of catching an error and rethrowing it as a higher level / more generic error.

The way this works behind the scenes is actually really neat - it takes advantage of the `Error` [cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) field to render a series of errors all stacked on top of each other. ie:
```
> a = new Error('a')
Error: a
    at <anonymous>:1:5
> b = new Error('b', { cause: a });
Error: b
    at <anonymous>:1:5
Caused by: Error: a
    at <anonymous>:1:5
```

This means that at higher levels we can purely catch / handle the higher level of error without having to care about the more specific exact situation, while also exposing the more specific situation up the call stack for debugging without any extra additional complexity.

src/room/data-track/packet/index.ts
```diff
@@ -0,0 +1,343 @@
...
+  toJSON() {
+    return {
+      marker: this.marker,
+      trackHandle: this.trackHandle.value,
+      sequence: this.sequence.value,
+      frameNumber: this.frameNumber.value,
+      timestamp: this.timestamp.timestamp,
+      extensions: this.extensions.toJSON(),
+    };
+  }
```
1egoman, line 286: Note that all data track packet related classes have a `toJSON` method for debug printing, which provides a similar set of functionality as the debug trait in the rust implementation.

src/room/errors.ts
```diff
@@ -1,15 +1,33 @@
...
+/**
+ * LiveKit specific error type representing an error with an associated set of reasons.
+ * Use this to represent an error with multiple different but contextually related variants.
+ * */
+export abstract class LivekitReasonedError<Reason> extends LivekitError {
+  abstract reason: Reason;
+
+  abstract reasonName: string;
+}
+
```
1egoman, line 30: I'm on the fence on how useful this is in its current state, but I added a `LivekitReasonedError` to try to encode the standard `reason` / `reasonName` fields that a bunch of error types all seem to be using.

What I really want to have here (and the reason I originally introduced this) is a `isReason` that uses a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) on the return value so that any "reasoned error" can be easily type narrowed by reason from `ReasonedError<unknown>` to (for example) `ReasonedError<ReasonEnum.One>`. However, this it turned out was really difficult to do in an abstract class - there isn't a good way to get the "next level up" class name so that it can be included in the type predicate. So for now, I had to fall back to adding separate `isReason` implementations to each downstream error class... 😞

src/utils/TypedPromise.ts
```diff
@@ -14,6 +14,13 @@ export default class TypedPromise<T, E extends Error> extends Promise<T> {
     return super.catch(onrejected);
   }
 
+  static resolve: {
+    (): TypedPromise<void, never>;
+    <V>(value: V): TypedPromise<Awaited<V>, never>;
+  } = <V>(value?: V): TypedPromise<Awaited<V>, never> => {
+    return super.resolve(value) as TypedPromise<Awaited<V>, never>;
+  };
+
```
1egoman, line 23: Note that in the process of digging into all the error handling logic, I ended up discovering an issue in `TypedPromise` - `TypedPromise.resolve` wasn't returning a `TypedPromise` value which was leading to incorrect result types on downstream functions containing `TypedPromise.resolve`.

If we move fully to this `Throws` type thing, this obviously will get deleted. But in the case we opt not to I wanted to get the bugfix in.

src/room/data-track/packet/index.ts
```diff
@@ -0,0 +1,343 @@
...
+    byteIndex += U16_LENGTH_BYTES;
+    dataView.setUint16(byteIndex, this.sequence.value);
+    byteIndex += U16_LENGTH_BYTES;
+    dataView.setUint16(byteIndex, this.frameNumber.value);
+    byteIndex += U16_LENGTH_BYTES;
+    dataView.setUint32(byteIndex, this.timestamp.asTicks());
+    byteIndex += U32_LENGTH_BYTES;
+
+    if (extensionsLengthBytes > 0) {
+      // NOTE: the below value shouldn't have one subtracted from it.
```
ladvoc: comment: I don't think I would call this a bug. RTP does something similar with the idea being if the extension bit is set, there should always be at least one extension word. I think either way is valid as long as it is explicit in the specification.

src/room/data-track/packet/index.ts
```diff
@@ -0,0 +1,343 @@
...
+    let totalLengthBytes = BASE_HEADER_LEN;
+    if (extLengthBytes > 0) {
+      totalLengthBytes += EXT_WORDS_INDICATOR_SIZE + extLengthBytes + extPaddingLengthBytes;
+    }
+    return totalLengthBytes;
+  }
+
+  toBinaryInto(dataView: DataView): number {
+    let initial = SUPPORTED_VERSION << VERSION_SHIFT;
+
```
ladvoc, line 95: question: Probably missing something, but is there a check to ensure the data view has enough capacity for serialization? I see such a check for deserialization (where the error is _TooShort_).

src/room/data-track/packet/serializable.ts
```diff
@@ -0,0 +1,25 @@
+/** An abstract class implementing common behavior related to data track binary serialization. */
+export default abstract class Serializable {
```
ladvoc, line 5: praise: This is clean.

src/room/data-track/packet/index.test.ts
```diff
@@ -0,0 +1,521 @@
...
+        0x0,
+      ]);
+      packetBytes[0] |= 1 << EXT_FLAG_SHIFT; // Extension flag
+
+      expect(() => DataTrackPacket.fromBinary(packetBytes)).toThrow(
+        'Header exceeds total packet length',
+      );
+    });
+  });
+});
```
ladvoc, line 615: comment: Not strictly necessary, but it might be a good idea to add a simple round trip serialization/deserialization test. I added such as test in the Rust implementation the other day (see _packet/mod.rs_).

1egoman: Ah yes, good point. Added in 31a3b58.

src/room/data-track/packet/index.ts
1egoman: Ah, I see. I've updated the comments to reflect this updated description in 50dc51e.

src/room/data-track/packet/index.ts
1egoman: Good point, I seemed to have missed this. Added the [two new serialization error cases](https://github.com/livekit/rust-sdks/blob/28a650c325418b3bb8dfee66176ba6b1bbd26b20/livekit-datatrack/src/packet/serialize.rs#L21-L25) from the rust implementation and some tests to exercise them.

src/room/data-track/packet/extensions.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  toBinaryLengthBytes(): number {
+    return (
+      U16_LENGTH_BYTES /* tag */ +
+      U16_LENGTH_BYTES /* length */ +
+      DataTrackUserTimestampExtension.lengthBytes
+    );
+  }
+
+  toBinaryInto(dataView: DataView): Throws<number, never> {
```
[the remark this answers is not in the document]
1egoman: Throws no longer has a second `E` type parameter optional default, so omitting it results in `Generic type 'Throws' requires 2 type argument(s).`. Also just generally this was a conscious choice I made, `Throws<T, never>` reads much better in that it explicitly says errors will "never" be thrown.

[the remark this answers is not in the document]
1egoman: I don't get an eslint issue in this situation locally, and the ci job run also looks to be fine. I don't think there is anything left to do here.

src/utils/TypedPromise.ts
lukasIO: doesn't this make the return type of `resolve` inherently async though?

1egoman: I'm not sure what you mean - the return type of `Promise.resolve` should always be a some sort of thenable value?

FWIW the intention was for this from a type perspective to be identical to the builtin types (included below for reference), only swapping `Promise` for `TypedPromise` everywhere.

```typescript
    /**
     * Creates a new resolved promise.
     * @returns A resolved promise.
     */
    resolve(): Promise<void>;
    /**
     * Creates a new resolved promise for the provided value.
     * @param value A promise.
     * @returns A promise whose internal state matches the provided promise.
     */
    resolve<T>(value: T): Promise<Awaited<T>>;
```

lukasIO: ah, this is the static method! makes sense

throws-transformer/README.md
lukasIO, line 1: question: is the plan to move this outside of this repo eventually?

src/room/errors.ts
lukasIO: suggestion:

I tried to allow for narrowing in other places by introducing a `readonly` tag on the error and explicitly defining that error as an extension of LiveKitError. 

I guess the same (making it readonly) could be applied for reason which should still work for the union example you have above

throws-transformer/README.md
1egoman: I think it probably makes sense, yea. For now, I kept it vendored in here just to increase iteration speed and because the repo you made with this was private / I didn't have permission to make it public. But I could see it becoming a helper package like `@livekit/mutex` / etc once it is a bit more stable.

If you have a stronger opinion about the timeline of that move though feel free to mention it here. My preference in a vaccum I think would be to wait to break it out until the static analysis checking script is a bit more robust and some other folks (at least you) have been able to give ita try and weigh in on how it works.

src/room/errors.ts
1egoman: That's a good idea - unfortunately, it doesn't seem to work for me 😞 since there's no way to type `reason` as specific enum variant. Here is a [typescript playground link](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAWwIYzACgJSIN4CwAUIolAE4Ce+xppUAFuXAO6IAiqUqAKuahADWACVRgAJgBsApgFFyzcgDoocOABlU5AObScAblqIAvoghcIDRJmm5CJOgHpHiWQA8YAZ1hhtiVAAOAcwCDABcRqQwwNaRdNKIGN5iENJwMZzcfAIiYlJyCnDkiABkJXGk0kpeAErSqJ4ImJm8-EKiEjLyinUNCEp1ntLkAG7S4thxdhVmCI0ySpJw2jbY+ojOiAAKzAHDktQARtIQcMjSnhxcrTkd+d1FADwt2e15XYXkvY1gAxfDY3EAD44qZpJIhjQHHQyExWIhpIZoSZiHFNjs4AE4ENxP4giFLBFkdFYsjKokwMlIGkMtdXrlOgVFKVymSEUpyPUfogALx8q5ZNoM+6fb79QYA8ZTKEw0inSlwBZLFa2dabADKsEkkgFNzejIe5EeYkoQMQYUwoIREIS9llsOYbERoKMpmIxlRRE9mwAtH7-QHA0HgyG-Z7PdI3FjyFAzJIGpd1DAxoIYFBDQi3FBpBJLhm7bNxNIwogwCBkMdyEijJsALJFBIYYBwEsMKBQAKeMLOItjJZ7ZTIOAALxg2tQSiK2kcOZ9AFV1Y5xHAIJ5HAB1aSHRwAKVQI1Q6og5BgASgjjqwGGOdSjgA4ktDqhJAB9ADyhwAVicoGvDY5zBAIYjEAoYAH4S3AQQwFYMBqwceVvHIEBoCKTBTiLEsywrYYABoUAuTxUF0CDECQjBtHwzFYDmUi8DMVAgOkUioJglgkGMaZkU8EAB0wc5PCI3REAAHxExAAHIxART5EAYBpEBXCAQAUcYJLWOJGC8JQwFQc5eUkpMxgAaTTQ0JKRGEtM8JQMISHlC2dL1iRiTAoEoPZ0kUs94EpMDbMYyEAEJ+Qk8Ai2ADA1K4+1rICpiDOo3zPH80CnJhD0HEyzKIyjIpY1QQ4kIEWMIHjQTECM6RUygMUwHGQ1Hjqs1I2zXNKuTaqzNkgtOVQZcwAOfwiooErED6n4Szq+DSEK4roHGrkEAAOT04syIoCikRyohI2jWMc3LXV6TuD4eiWpACwlUZxgMgAGXCjB4NRNB0ezEAARkeogduIPb8rjBNjqFU6mSeIw6szNrxEuF4QfeMGvgugy4duBHDTq76WqzHMYY6lM0zqhrPmeOl4YNUULrNXr6gGobdP0hyJNR-URUUCzPVIPq6eoCaECmi6ZsWvowFW84S3I3whcQigUNUch+MI4j1slyjhcmxA6piugeL4j6AE58IEoTVU0hhtL5pAHMtyy6Diy2xfelnhTOoo6oAbUtgBdW2dqiTw6qaqHcdhsm0Yp86RaBTBLam7AS2sxJQ8FcO2aeGpqbiTkoBUpA3PNmzLf8S5WNg4vNdwPlrcF11OY2rgYAgMgXq0XQcBlOhs9z0tpDYZ3QcNS02QkgA1Z8QASVQ4EQeM3ubxBjn8RAD0kGBcVGoQ5IRiTvvtfv0cpkWlGejRW+kXe6A0rK6+SWAm85IYbvETAV4nrDy0rbWuekHPyCQeq+5h1Zq7BWMwAAGd03AABI8CvyqKoTUJ5fCYA+gANmwKYLwS8H6SlxHApQYCL4wn3hHN2F0-iP0BEQxAV9SDZU9GgDABggA) if you have interest in trying other ideas out.

PR #1793, opened 2026-01-24
fix: prevent mute cycling when restarting tracks during unmute

mfairley:
When unmuting audio/video tracks that require reacquisition (e.g., due to
stopOnMute, ended track state, or device changes), the track restart would
cause a mute cycle (enabled=false then immediately enabled=true).

This happened because:
1. restartTrack() creates a new MediaStreamTrack
2. setMediaStreamTrack() syncs enabled state with !this.isMuted
3. But isMuted is still true at this point (super.unmute() hasn't run yet)
4. So the new track starts with enabled=false
5. Then super.unmute() sets enabled=true

On iOS, this causes audible mute/unmute sounds for the microphone and can
trigger system callbacks multiple times as the system detects the rapid
enabled state changes.

The fix adds a targetEnabled parameter that flows through:
- unmute() -> restartTrack(undefined, true)
- restartTrack() -> restart(constraints, targetEnabled)
- restart() -> setMediaStreamTrack(newTrack, false, targetEnabled)

When targetEnabled is provided, setMediaStreamTrack uses it instead of
deriving the enabled state from isMuted, ensuring the new track starts
with the correct enabled state.

## Related issues
- https://github.com/livekit/react-native-webrtc/issues/66
- https://github.com/livekit/client-sdk-react-native/issues/287

This partially addresses the above issues but does not fix the possibly unnecessary restarts caused by device ID mismatches.

## Test plan
- Tested muting and unmuting microphone on iOS
- Tested muting and unmuting video on iOS
- Verified no mute cycling occurs during unmute with track restart

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented mute–unmute cycling when re-acquiring microphones by preserving the intended enabled state during restarts.
  * Ensured camera restarts refresh simulcast encodings and rebind sender tracks so video continuity and quality are maintained.
  * Improved unmute behavior so devices come back enabled predictably after restart, avoiding unexpected muted states.

### DIFF
--- a/src/room/track/LocalTrack.ts
+++ b/src/room/track/LocalTrack.ts
@@ -150,7 +150,7 @@ export default abstract class LocalTrack<
   private async setMediaStreamTrack(
     newTrack: MediaStreamTrack,
     force?: boolean,
-    targetEnabled?: boolean,
+    isUnmuting?: boolean,
   ) {
     if (newTrack === this._mediaStreamTrack && !force) {
       return;
@@ -208,8 +208,8 @@ export default abstract class LocalTrack<
     this._mediaStreamTrack = newTrack;
     if (newTrack) {
       // sync muted state with the enabled state of the newly provided track
-      // use targetEnabled if provided (e.g. during unmute restart) to avoid mute cycling
-      this._mediaStreamTrack.enabled = targetEnabled ?? !this.isMuted;
+      // if restarting as part of an unmute, set enabled to true directly to avoid mute cycling
+      this._mediaStreamTrack.enabled = isUnmuting ? true : !this.isMuted;
       // when a valid track is replace, we'd want to start producing
       await this.resumeUpstream();
       this.attachedElements.forEach((el) => {
@@ -328,7 +328,7 @@ export default abstract class LocalTrack<
     }
   }
 
-  protected async restart(constraints?: MediaTrackConstraints, targetEnabled?: boolean) {
+  protected async restart(constraints?: MediaTrackConstraints, isUnmuting?: boolean) {
     this.manuallyStopped = false;
     const unlock = await this.trackChangeLock.lock();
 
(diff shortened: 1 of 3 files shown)


src/room/track/LocalTrack.ts
```diff
@@ -363,7 +368,7 @@ export default abstract class LocalTrack<
       newTrack.addEventListener('ended', this.handleEnded);
       this.log.debug('re-acquired MediaStreamTrack', this.logContext);
 
-      await this.setMediaStreamTrack(newTrack);
+      await this.setMediaStreamTrack(newTrack, false, targetEnabled);
```
lukasIO: restarts would always cause a track to be enabled afterwards, so I believe we can get rid of passing the `targetEnabled` prop in the other methods and only keep it here

mfairley: Could you explain a bit more? We are triggering the restart from the unmute method so I think we need to pass the argument from unmute to where it's needed, which requires going through a few methods.

lukasIO: my thinking was that for `restartTrack` it's implicit that `targetEnabled = true`. 
But looking at it again that seems like a misconception on my side, sorry about that! 

Just wanted to ensure the public API stays as lean as possible and ideally it would be preferable to not expose the `targetEnabled` param on `restartTrack`

mfairley: Yeah, I understand wanting to keep the public API as lean as possible. Some alternative approaches include introducing a class variable such as `isPendingUnmute`, but then that introduces additional state to manage and is less explicit. Another is using the current approach but having a better name e.g. targetEnabled -> isUnmuting since this is probably the only case when you don't want to immediately sync the mute state. Another is adding this parameter to VideoCaptureOptions/AudioCaptureOptions but it doesn't seem to quite fit with the intention of that. Let me know if you have any preferences.

lukasIO: looking at the code, I think it should be possible to swap out `restartTrack` with `restart` in the unmute path. Do you see any problem with that?

This would allow for the `restartTrack` signature to stay unchanged and only introduce the new param on the protected `restart` method. 

I like your suggestion of renaming it to `isUnmuting`

PR #1794, opened 2026-01-24
fix: prevent unnecessary track restarts on unmute when using ideal device constraints

mfairley:
When using `{ ideal: 'default' }` device constraints (the default), unmuting
would trigger unnecessary track restarts because the comparison between the
constraint value ('default') and actual device ID (e.g. 'audio') would always
fail.

This adds a `pendingDeviceChange` flag that is only set when `setDeviceId()`
is explicitly called while the track is muted. The unmute logic now checks
this flag instead of comparing constraint values against the actual device,
ensuring restarts only happen when the user actually requested a device change.

## Related issues
- https://github.com/livekit/react-native-webrtc/issues/66
- https://github.com/livekit/client-sdk-react-native/issues/287

## Test plan
- Tested muting and unmuting microphone on iOS with default device constraints
- Verified no unnecessary restarts occur on unmute
- Verified `setDeviceId()` while muted still triggers restart on unmute

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed unnecessary audio track restarts when unmuting with ideal device constraints; device changes made while muted now apply on unmute without forcing a restart.
* **Chores**
  * Added release changeset entry for this patch.

### DIFF
--- a/.changeset/funny-masks-strive.md
+++ b/.changeset/funny-masks-strive.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Fix unnecessary track restarts on unmute when using ideal device constraints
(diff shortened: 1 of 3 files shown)


davidliu (approved):
LGTM! @lukasIO do you see anything affecting the web side? Tested web and seems to mute/unmute fine, without reacquiring the mic track.

lukasIO:
this makes sense to me!

PR #1798, opened 2026-01-28
Data tracks packetizer and depacketizer

1egoman:
Adds the Packetizer and Depacketizer for Data Tracks.

The packetizer takes a `DataTrackFrame` (new concept, basically right now a pair of a `Uint8Array` payload and any packet extensions) and converts it into a series of `DataTrackPackets`.

The depacketizer does the inverse - it takes a series of packets and aggregates them to recreate the original `DataTrackFrame`.

## Summary by CodeRabbit

* **New Features**
  * End-to-end data-track packetizer and depacketizer, plus enhanced frame, clock, and timestamp utilities and a dedicated data-tracks logger.

* **API Changes**
  * Packet payloads now use Uint8Array; header timestamps serialize as ticks; Throws success typing broadened to allow primitive values.

* **Bug Fixes**
  * Improved timestamp monotonicity, wraparound handling, and buffer/overflow protections.

* **Tests**
  * Expanded packetizer/depacketizer and timestamp/clock test coverage.

* **Documentation**
  * Added release entry and updated type docs.

### DIFF
--- a/src/room/data-track/depacketizer.ts
+++ b/src/room/data-track/depacketizer.ts
@@ -257,14 +257,14 @@ export class DataTrackDepacketizer {
     }
     const payload = new Uint8Array(payloadLengthBytes);
 
-    let sequencePointer = partial.startSequence.value;
+    let sequencePointer = partial.startSequence.clone();
     let payloadOffsetPointerBytes = 0;
     while (true) {
-      const partialPayload = partial.payloads.get(sequencePointer);
+      const partialPayload = partial.payloads.get(sequencePointer.value);
       if (!partialPayload) {
         break;
       }
-      partial.payloads.delete(sequencePointer);
+      partial.payloads.delete(sequencePointer.value);
 
       const payloadRemainingBytes = payload.length - payloadOffsetPointerBytes;
       if (partialPayload.length > payloadRemainingBytes) {
@@ -277,8 +277,9 @@ export class DataTrackDepacketizer {
       payload.set(partialPayload, payloadOffsetPointerBytes);
       payloadOffsetPointerBytes += partialPayload.length;
 
-      if (sequencePointer < endSequence) {
-        sequencePointer += 1;
+      // NOTE: sequencePointer could wrap around, which is why this isn't a "<"
+      if (sequencePointer.value != endSequence) {
+        sequencePointer.increment();
         continue;
       }
 


src/room/data-track/packet/extensions.ts
```diff
@@ -0,0 +1,256 @@
...
+          break;
+
+        default:
+          // Skip over unknown extensions (forward compatible).
+          if (dataView.byteLength - byteIndex < lengthBytes) {
+            throw DataTrackDeserializeError.malformedExt(tag);
+          }
+          byteIndex += lengthBytes;
+          break;
+      }
```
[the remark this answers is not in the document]
1egoman: This is incorrect, `lengthBytes` takes into account the whole length of the packet, not just the `iv` section. So `lengthBytes` is `13`.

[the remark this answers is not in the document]
1egoman: Ah, I see - the variable lengthBytes is named incorrectly, this is actually a rtp length which is the length in bytes minus 1. I have fixed this [here](https://github.com/livekit/client-sdk-js/pull/1792/changes/95d52f427ee4a023c727a05c6e4a4fa6eaab7de9) and will rebase this on top of that change shortly.

src/room/data-track/packetizer.ts
```diff
@@ -0,0 +1,128 @@
...
+      const packetPayloadLengthBytes = Math.min(
+        // All but the last packet will be max length ...
+        maxPayloadSizeBytes,
+        // ... and the last packet will be as long as it needs to be to finish out the buffer.
+        frame.payload.byteLength - indexBytes,
+      );
+      const packetPayload = new Uint8Array(frame.payload, indexBytes, packetPayloadLengthBytes);
+
+      yield new DataTrackPacket(packetHeader, packetPayload);
+    }
```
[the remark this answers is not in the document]
1egoman: Based on the rust implementation, the behavior I have implemented is correct - a zero length payload should produce no packets rather than one packet with an empty payload.

src/room/data-track/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+    return new DataTrackTimestamp(this.timestamp.value, this.rateInHz);
+  }
+
+  wrappingAdd(n: number) {
+    this.timestamp.increment(n);
+  }
+
+  isBefore(other: DataTrackTimestamp<RateInHz>) {
+    return this.timestamp.value < other.timestamp.value;
+  }
```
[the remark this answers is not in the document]
1egoman: I'll add a docstring making this more clear, but the existing behavior is correct - isBefore is not resilient to wraparounds, and that matches the rust implementation I am adapting.

src/room/data-track/packetizer.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const maxPayloadSizeBytes = this.mtuSizeBytes - headerSerializedLengthBytes;
+
+    const packetCount = Math.ceil(frame.payload.byteLength / maxPayloadSizeBytes);
+
+    for (
+      let index = 0, indexBytes = 0;
+      indexBytes < frame.payload.byteLength;
+      [index, indexBytes] = [index + 1, indexBytes + maxPayloadSizeBytes]
+    ) {
```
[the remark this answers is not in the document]
1egoman: This should be done.

src/room/data-track/packet/extensions.ts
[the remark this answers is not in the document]
1egoman: I've fixed this [here](https://github.com/livekit/client-sdk-js/pull/1792/changes/95d52f427ee4a023c727a05c6e4a4fa6eaab7de9), the changes will end up in this pull request once it is rebased on top shortly.

src/room/data-track/packetizer.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      expect(packetHeaderJson.marker).toStrictEqual(
+        DataTrackPacketizer.computeFrameMarker(index, packets.length),
+      );
+      expect(packetHeaderJson.frameNumber).toStrictEqual(0);
+      expect(packetHeaderJson.trackHandle).toStrictEqual(1);
+      expect(packetHeaderJson.sequence).toStrictEqual(index);
+      expect(packetHeaderJson.extensions).toStrictEqual(frame.extensions.toJSON());
+      index += 1;
+    }
+  });
```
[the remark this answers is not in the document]
1egoman: What is here is fine and matches the rust implementation so I'm hesitant to bail early like suggested.

throws-transformer/README.md
```diff
@@ -0,0 +1,267 @@
...
+  }
+}
+```
+
+## API
+
+### Types
+
+```typescript
+type Primatives = null | undefined | string | number | bigint | boolean | symbol;
```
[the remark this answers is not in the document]
1egoman: Oops, fixed.

src/room/data-track/packet/extensions.ts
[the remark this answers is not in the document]
1egoman: Ok, this has been rebased and should be addressed in this branch now. Can you confirm that it is?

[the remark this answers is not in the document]
1egoman: Ok, this has been rebased and should be addressed in this branch now. Can you confirm that it is?

src/room/data-track/packet/index.ts
```diff
@@ -302,9 +302,9 @@ export enum FrameMarker {
 export class DataTrackPacket extends Serializable {
   header: DataTrackPacketHeader;
 
-  payload: ArrayBuffer;
+  payload: Uint8Array;
```
1egoman, line 306: A notable change - originally I had `DataTrackPacket`'s `payload` field stored as an `ArrayBuffer` because I thought it would be the source of truth of data. However as I got further into the packetizer / depacketizer work I realized that having a `DataTrackPacket` be a view (ie, with `Uint8Array`) into the array buffer being received from high level abstractions (ie, in the future a network packet) leads to much less data copying, so I switched it here.

src/room/data-track/depacketizer.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      }),
+      new Uint8Array(8),
+    );
+    expect(() => depacketizer.push(packet)).toThrowError(
+      'Frame 103 dropped: Initial packet was never received.',
+    );
+  });
+
+  it('should throw "bufferFull" when too many packets have been sent', () => {
+    const depacketizer = new DataTrackDepacketizer();
```
1egoman, line 165: FYI @ladvoc I added a few more tests in here than the [corresponding rust version](https://github.com/livekit/rust-sdks/blob/7abbf0b3495dfc4eb2bae3f2e22b5653850c1fe4/livekit-datatrack/src/remote/depacketizer.rs#L241) has. It might be worth porting the last few over - it just exercises all the error cases.

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,268 @@
...
+  push(
+    packet: DataTrackPacket,
+    options?: PushOptions,
+  ): Throws<
+    DataTrackFrame | null,
+    | DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.Interrupted>
+    | DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.BufferFull>
+    | DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.UnknownFrame>
+    | DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.Incomplete>
+  > {
```
1egoman, line 108: An important deviation from how the rust implementation works - the rust implementation returns a `DepacketizerPushResult` here which contains `frame` and `drop_error` fields (both `Option`s, so they are nullable). Note that either one can be set, or in one case (a packet being received from a different frame "interrupting" processing) both can be set.

Instead the web implementation opts to get rid of this `DepacketizerPushResult` layer and just return a `DataTrackFrame | null`. In all error cases the web implementation opts to `throw` instead, and for that one case where both can be set, the web implementation exposes an option passed to `push` that allows the caller to choose if they want to allow new packets to come in and overwrite an existing frame (`errorOnPartialFrames`). If this is `true`, an error is thrown, and if it is `false`, a warning is printed instead.

This option primarily exists for the tests right now but I could also see maybe being useful for like verifying idealized packetizing/depacketizing workflows where packet loss won't be a thing elsewhere too or something like that.

I think there could also be an argument for getting rid of the error all together and just making it a warning always if others think that is worthwhile to reduce complexity.

IMO the rust implementation should adopt this approach as well but I will leave that up to @ladvoc.

src/room/data-track/frame.ts
```diff
@@ -0,0 +1,8 @@
+import { DataTrackExtensions } from './packet/extensions';
+import { DataTrackPacketizer } from './packetizer';
+
+/** A pair of payload bytes and packet extensions which can be fed into a {@link DataTrackPacketizer}. */
+export type DataTrackFrame = {
+  payload: ArrayBuffer;
+  extensions: DataTrackExtensions;
+};
```
1egoman, line 8: FYI @ladvoc I have adopted the common data frame frame implementation as we discussed in a 1:1 rather than the [packetize / depacketize specific implementations in rust](https://github.com/livekit/rust-sdks/blob/7abbf0b3495dfc4eb2bae3f2e22b5653850c1fe4/livekit-datatrack/src/remote/depacketizer.rs#L27-L32). IMO it would be worth combining these in the rust implementation as well.

src/room/data-track/utils.ts
```diff
@@ -42,31 +47,139 @@ export class WrapAroundUnsignedInt<MaxSize extends number> {
...
   static fromRtpTicks(rtpTicks: number) {
     return new DataTrackTimestamp(rtpTicks, 90_000);
   }
 
-  asTicks() {
-    return this.timestamp;
+  static rtpRandom() {
+    // FIXME: does this need to be a higher quality PRNG?
+    const randomValue = Math.round(Math.random() * U32_MAX_SIZE);
+    return DataTrackTimestamp.fromRtpTicks(randomValue);
```
1egoman, line 99: @ladvoc I am curious your thoughts on this - when initializing the rtp clock in the packetizer with a random timestamp, how high quality does the PRNG need to be? Does it need to be cryptography grade or is `Math.random` (more info [about its pitfalls](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random)) good enough?

src/utils/throws.ts
```diff
@@ -11,7 +13,7 @@
  *
  * For more info about how this is checked, see ./throws-transformer at the root of this repo.
  */
-export type Throws<T, E extends Error> = T & { readonly __throws?: E };
+export type Throws<T, E extends Error> = (T & { readonly __throws?: E }) | Extract<T, Primitives>;
```
1egoman, line 16: A small error handling logic update - with the above, `Throws<{ key: "value" } | null, /* ... */>` now works. Previously if you put a non object into the first parameter of `Throws` the branding would get all weird (and would sometimes get dropped all together or parts of the passed union which weren't objects would get dropped depending on the exact circumstance).

src/room/data-track/utils.ts
```diff
@@ -42,31 +47,139 @@ export class WrapAroundUnsignedInt<MaxSize extends number> {
...
+
+    let result = this.base.clone();
+    result.wrappingAdd(durationTicks);
+
+    // Enforce monotonicity in RTP wraparound space
+    if (result.isBefore(this.previous)) {
+      result = this.previous;
+    }
+    this.previous = result;
+    return result;
```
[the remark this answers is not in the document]
1egoman: This should be addressed, can you confirm?

src/room/data-track/utils.ts
```diff
@@ -42,31 +47,139 @@ export class WrapAroundUnsignedInt<MaxSize extends number> {
...
+  previous: DataTrackTimestamp<RateInHz>;
+
+  rateInHz: RateInHz;
+
+  private constructor(rateInHz: RateInHz, epoch: Date, base: DataTrackTimestamp<RateInHz>) {
+    this.epoch = epoch;
+    this.base = base;
+    this.previous = base;
     this.rateInHz = rateInHz;
   }
```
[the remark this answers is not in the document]
1egoman: This should be addressed, can you confirm?

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,283 @@
...
+
+    // Note: receiving a packet with a duplicate `sequence` value is something that likely won't
+    // happen in actual use, but even if it does (maybe a low level network retransmission?) the
+    // last packet with a given sequence received should always win.
+    if (this.partial.payloads.has(packet.header.sequence.value)) {
+      log.warn(
+        `Data track frame ${this.partial.frameNumber} received duplicate packet for sequence ${packet.header.sequence.value}, so replacing with newly received packet.`,
+      );
+    }
+    this.partial.payloads.set(packet.header.sequence.value, packet.payload);
```
1egoman, line 235: Note - I think this case may not properly be handled by the rust implementation, [here is the corresponding code](https://github.com/livekit/rust-sdks/blob/7abbf0b3495dfc4eb2bae3f2e22b5653850c1fe4/livekit-datatrack/src/remote/depacketizer.rs#L123C17-L124). In the rust implementation, receiving packets with identical sequence numbers adds to `payload_len` twice, resulting in the end buffer being allocated in `finalize` being longer than it should be otherwise.

Also this could be a potential DOS issue as well - a user could craft a data track frame with a bunch of packets all with the same sequence number but with large payloads (the `MAX_BUFFER_PACKETS` check wouldn't guard against this since they alla re the same sequence number). And then when the final data track packet were to be sent, `payload_len` would be massive, attempt a massive allocation, and cause the process to run out of memory.

cc @ladvoc

src/room/data-track/packet/index.ts
ladvoc: Makes a lot of sense. No reason to move it into a new buffer when the cursor can just be moved to isolate the payload.

src/room/data-track/frame.ts
ladvoc: Yes, that makes a lot of sense. However, it is worth calling out that this type should not be exposed in the public API, since the concept of extensions is an implementation detail. For example, E2EE metadata should not be accessible through the public frame type, and if we introduce more extensions in the future that control internal processing (e.g., a frame is marked as being compressed), these should also remain internal details.

src/room/data-track/utils.ts
ladvoc: I think the main reason RTP does this is to distinguish between streams rather than for security. Additionally, even if E2EE is not enabled for a track, track packets are still encrypted in transport between the SFU and clients, so I am not too concerned about making this cryptographic grade. However, I will do a bit more research on this before giving a definitive answer.

src/room/data-track/depacketizer.ts
ladvoc: Good callout, I will revise how this is handled in Rust.

src/index.ts
```diff
@@ -155,5 +155,7 @@ export {
   type DataTrackUserTimestampExtension,
   type DataTrackE2eeExtension,
 } from './room/data-track/packet/extensions';
+export { DataTrackPacketizer } from './room/data-track/packetizer'; // FIXME: should this be exported? Maybe not?
```
ladvoc: comment: If you mean exported publicly, then no, this should be an internal implementation detail.

src/room/data-track/depacketizer.ts
ladvoc: Taking another look at this, I wanted to mention a third possibility. Push could return a complete frame, if one is available, but the drop error could be exposed through a property on the depacketizer (e.g., `depacketizer.last_push_error()`). If the consumer is interested, they can access this property after calling push.

src/room/data-track/depacketizer.test.ts
```diff
@@ -0,0 +1,334 @@
...
+      }),
+      packetPayload,
+    );
+    expect(depacketizer.push(startPacketB)).toBeNull();
+
+    const interPacketB = new DataTrackPacket(
+      new DataTrackPacketHeader({
+        ...packetHeaderParams,
+        marker: FrameMarker.Inter,
+        sequence: WrapAroundUnsignedInt.u16(1),
```
ladvoc: comment: I ported this test back to Rust. In production, sequence number does not get reset between frames, so I reflect that in the test.

ladvoc (commented):
I left some initial comments but am planning to make another pass.

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,283 @@
...
+    DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.Interrupted>
+  > {
+    if (packet.header.marker !== FrameMarker.Single) {
+      // @throws-transformer ignore - this should be treated as a "panic" and not be caught
+      throw new Error(
+        `Depacketizer.frameFromSingle: packet.header.marker was not FrameMarker.Single, found ${packet.header.marker}.`,
+      );
+    }
+
+    if (this.partial) {
```
xianshijing-lk, line 138: [question](https://pullpo.io/cc?l=question) 
will things get out of order ? 

I wonder if you should interrupt only when the frame numbers match, like
packet.header.frameNumber.value === this.partial.frameNumber

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,283 @@
...
+        `Depacketizer.beginPartial: packet.header.marker was not FrameMarker.Start, found ${packet.header.marker}.`,
+      );
+    }
+
+    if (this.partial) {
+      if (options?.errorOnPartialFrames) {
+        throw DataTrackDepacketizerDropError.interrupted(this.partial.frameNumber);
+      } else {
+        log.warn(
+          `Data track frame ${this.partial.frameNumber} was interrupted by the start of a new frame, dropping.`,
```
xianshijing-lk, line 173: same question here.

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,283 @@
...
+  private finalize(
+    partial: PartialFrame,
+    endSequence: number,
+  ): Throws<
+    DataTrackFrame,
+    DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.Incomplete>
+  > {
+    const received = partial.payloads.size;
+    const payloadLengthBytes = Array.from(partial.payloads.values()).reduce(
+      (sum, i) => sum + i.length,
```
xianshijing-lk: [nitpick](https://pullpo.io/cc?l=nitpick) 

maybe it is slight better with:
let payloadLengthBytes = 0;
for (const p of partial.payloads.values()) payloadLengthBytes += p.length;

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    if (!this.partial) {
+      throw DataTrackDepacketizerDropError.unknownFrame(packet.header.frameNumber.value);
+    }
+
+    if (packet.header.frameNumber.value !== this.partial.frameNumber) {
+      throw DataTrackDepacketizerDropError.interrupted(this.partial.frameNumber);
+    }
+    if (this.partial.payloads.size == DataTrackDepacketizer.MAX_BUFFER_PACKETS) {
```
xianshijing-lk: [nitpick](https://pullpo.io/cc?l=nitpick) 

should it be >= instead of == ? 
if (this.partial.payloads.size >= DataTrackDepacketizer.MAX_BUFFER_PACKETS) {

src/room/data-track/frame.ts
1egoman: Good point - that being said, right now `DataTrackFrame` isn't exported by the package, only the `frame.ts` module. So I think this should be fine.

src/room/data-track/depacketizer.ts
1egoman: I think I'd be fine with your suggestion too, though I think my approach is a little more js-idiomatic. Do you see there being many more of these "a warning was raised during depacketizing" types of cases to worry about?

Another option too - at least for this one "interrupted" warning, I'm not really sure how important it is to really return it as an error in the majority of cases. Assuming there aren't more of these cases that a more advanced depacketizer would encounter / this is truly a special case, it could just be logged as a warning and that's it.

src/index.ts
1egoman: Makes sense. For now I will swap them out with `import`s ([more info](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_a_module_for_its_side_effects_only)) so they will continue to get checked by static analysis tools. Once I'm importing these in code that is already exported I will get rid of these.

src/room/data-track/depacketizer.test.ts
1egoman: Ah, good to know, I have updated the test in 8363316.

src/room/data-track/depacketizer.ts
1egoman: Good point, that's probably a good idea to do for safety, updated. (Also what was there before was not quite right, it should have been a `===` not `==`, not sure how that slipped through.)

src/room/data-track/depacketizer.ts
1egoman: Sure sounds good!

src/room/data-track/depacketizer.ts
1egoman: The rust implementation was built to support out of order packet receiving, so I copied that behavior when I adapted it into typescript. I will defer to @ladvoc around how important this is but my understanding is it is important because UDP packets don't guarantee order.

I've also got a [somewhat related test here](https://github.com/livekit/client-sdk-js/blob/9ca4814e1ef379a3f607c310c45407d58090d716/src/room/data-track/depacketizer.test.ts#L284-L334) which exercises this, but reading this is making me realize it would be good to have a test directly covering this case as well. I will add one!

src/room/data-track/depacketizer.ts
1egoman: https://github.com/livekit/client-sdk-js/pull/1798#discussion_r2756178428

src/room/data-track/depacketizer.ts
1egoman: Added the additional test here: 082914f

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    // Note: receiving a packet with a duplicate `sequence` value is something that likely won't
+    // happen in actual use, but even if it does (maybe a low level network retransmission?) the
+    // last packet with a given sequence received should always win.
+    if (this.partial.payloads.has(packet.header.sequence.value)) {
+      log.warn(
+        `Data track frame ${this.partial.frameNumber} received duplicate packet for sequence ${packet.header.sequence.value}, so replacing with newly received packet.`,
+      );
+    }
+    this.partial.payloads.set(packet.header.sequence.value, packet.payload);
```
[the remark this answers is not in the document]
1egoman: This I don't think is all that big of a deal. I will add a comment calling it out and it can be addressed at a future date if it is important.

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,288 @@
...
+    if (this.partial) {
+      if (options?.errorOnPartialFrames) {
+        throw DataTrackDepacketizerDropError.interrupted(this.partial.frameNumber);
+      } else {
+        log.warn(
+          `Data track frame ${this.partial.frameNumber} was interrupted by the start of a new frame, dropping.`,
+        );
+      }
+    }
+    this.reset();
```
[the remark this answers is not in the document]
1egoman: I think this has been fixed, good point!

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,288 @@
...
```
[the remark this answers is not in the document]
1egoman: I think this has been fixed, good point!

src/room/data-track/utils.ts
```diff
@@ -42,31 +47,139 @@ export class WrapAroundUnsignedInt<MaxSize extends number> {
...
+
+  now(): DataTrackTimestamp<RateInHz> {
+    return this.at(new Date());
+  }
+
+  at(timestamp: Date) {
+    let elapsedMs = timestamp.getTime() - this.epoch.getTime();
+    let durationTicks = DataTrackClock.durationInMsToTicks(elapsedMs, this.rateInHz);
+
+    let result = this.base.clone();
```
xianshijing-lk, line 172: [question](https://pullpo.io/cc?l=question) 
curiously, any way to avoid the clone ? Note, in this function, there are quite some clones, I wonder if we can avoid some of them ?

src/room/data-track/utils.ts
xianshijing-lk: I wonder if it is better with crypto.getRandomValues ? or it is more or less the same

src/room/data-track/utils.ts
```diff
@@ -42,31 +47,139 @@ export class WrapAroundUnsignedInt<MaxSize extends number> {
...
+  clone() {
+    return new DataTrackTimestamp(this.timestamp.value, this.rateInHz);
+  }
+
+  wrappingAdd(n: number) {
+    this.timestamp.increment(n);
+  }
+
+  /** Returns true if {@link this} is before the passed other {@link DataTrackTimestamp}.
+   * Note that if the timestamp wraps around past its bounds, this function may not produce the
```
xianshijing-lk: is the wrapping a real concern ? If yes, I think we probably can easily address it, like
```
isBefore(other: DataTrackTimestamp<RateInHz>) {
  const a = this.timestamp.value >>> 0;
  const b = other.timestamp.value >>> 0;
  const diff = (b - a) >>> 0;
  return diff !== 0 && diff < 0x80000000;
}
```

xianshijing-lk (approved):
some nits, look good in general

src/room/data-track/utils.ts
1egoman: I had the same concern when writing this, FWIW. I ended up not being too worried about this in the end because in this case a "clone" is super cheap, just copying the inner `value` and wrapping it in a new class instance.

I could maybe refactor this so that every `DataTrackTimestamp` value is immutable / always returns a copy of itself, but that gets annoying with methods like `getThenIncrement` and when I was writing this initially I liked the idea much more of making them all homogeneous rather than making some methods mutable and some immutable.

Definitely open to other ideas here, but at least as of now I think I've picked the "least bad" option out of what came to mind.

src/room/data-track/utils.ts
1egoman: That was largely my question for jacob, yes, and if we needed this to be crypto-secure that's probably what I would use!

cc @ladvoc - I'm curious if you ended up digging into this any more. If you are unsure / want to be safe, I can migrate it over.

src/room/data-track/utils.ts
1egoman: Done, thanks!

src/room/data-track/utils.ts
```diff
@@ -42,31 +47,145 @@ export class WrapAroundUnsignedInt<MaxSize extends number> {
...
+  }
+
+  wrappingAdd(n: number) {
+    this.timestamp.increment(n);
+  }
+
+  /** Returns true if {@link this} is before the passed other {@link DataTrackTimestamp}. */
+  isBefore(other: DataTrackTimestamp<RateInHz>) {
+    return this.timestamp.isBefore(other.timestamp);
+  }
```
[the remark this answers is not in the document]
1egoman: No this is correct, shijing's comment resulted in the behavior of this changing, it is wraparound aware now.

src/room/data-track/utils.ts
ladvoc: Yeah, the best I can determine is randomizing is to make it more difficult to correlate streams in a packet capture (but doesn't prevent doing so), so using cryptographic grade randomness isn't necessary.

src/room/data-track/depacketizer.ts
ladvoc: I think in order to produce accurate statistics later, it is important we have a way to propagate all drop events up—even those that for now are more of edge cases. I'm also starting to think calling and treating these like errors might be misleading, since frames being dropped in a variety of ways is in inevitable in unreliable network conditions.

src/room/data-track/utils.ts
```diff
@@ -42,31 +47,145 @@ export class WrapAroundUnsignedInt<MaxSize extends number> {
...
+    return new WrapAroundUnsignedInt(previousValue, this.maxSize);
+  }
+
+  /** Returns true if {@link this} is before the passed other {@link WrapAroundUnsignedInt}. */
+  isBefore(other: WrapAroundUnsignedInt<MaxSize>) {
+    const a = this.value >>> 0;
+    const b = other.value >>> 0;
+    const diff = (b - a) >>> 0;
+    return diff !== 0 && diff < this.maxSize;
+  }
```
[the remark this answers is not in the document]
1egoman: I've added a test suite which seems to work after the modulus update, so I will keep it as is otherwise.

src/room/data-track/utils.ts
1egoman: I have added a comment to make the intent here clear.

src/room/data-track/depacketizer.ts
1egoman: > I'm also starting to think calling and treating these like errors might be misleading, since frames being dropped in a variety of ways is in inevitable in unreliable network conditions.

Yea that makes a ton of sense and I think is a good direction. I am tempted to keep it as is for now and we can readdress later when introducing the statistics stuff in a follow up change, any concerns with that?

src/room/data-track/depacketizer.ts
```diff
@@ -0,0 +1,297 @@
...
+        throw new Error(
+          `Depacketizer.finalize: Expected at least ${partialPayload.length} more bytes left in the payload buffer, only got ${payloadRemainingBytes} bytes.`,
+        );
+      }
+
+      payload.set(partialPayload, payloadOffsetPointerBytes);
+      payloadOffsetPointerBytes += partialPayload.length;
+
+      if (sequencePointer < endSequence) {
+        sequencePointer += 1;
```
ladvoc: issue: u16 wraparound needs to be handled here, and the comparison should be `!=` rather than `<` to prevent an edge case where the end sequence is less than the start sequence (e.g. start sequence is 65534, end is 1). This has been fixed in Rust.

1egoman: Thanks, I think this should be addressed in a464c28!

PR #1810, opened 2026-02-10
Data tracks - outgoing manager

1egoman:
This pull request introduces a new `OutgoingDataTrackManager` and the associated support infrastructure like `LocalDataTrack` to support sending data track packets from the web to the SFU.

### DIFF
--- a/src/room/data-track/track.ts
+++ b/src/room/data-track/track.ts
@@ -1,8 +1,6 @@
-import type { Throws } from '../../utils/throws';
 import type { DataTrackFrame } from './frame';
 import { type DataTrackHandle } from './handle';
 import type OutgoingDataTrackManager from './outgoing/OutgoingDataTrackManager';
-import type { DataTrackPushFrameError, DataTrackPushFrameErrorReason } from './outgoing/errors';
 
 export type DataTrackSid = string;
 
@@ -39,18 +37,14 @@ export class LocalDataTrack {
    *
    * - The track has been unpublished by the local participant or SFU
    * - The room is no longer connected
-   * - Frames are being pushed too fast (FIXME: this isn't the case in the js implementation?)
    */
-  tryPush(
-    payload: DataTrackFrame['payload'],
-    options?: { signal?: AbortSignal },
-  ): Throws<
-    void,
-    | DataTrackPushFrameError<DataTrackPushFrameErrorReason.Dropped>
-    | DataTrackPushFrameError<DataTrackPushFrameErrorReason.TrackUnpublished>
-  > {
-    // FIXME: rust implementation maps errors to dropped here?
-    // .map_err(|err| PushFrameError::new(err.into_inner(), PushFrameErrorReason::Dropped))
-    return this.manager.tryProcessAndSend(this.info.pubHandle, payload, options);
+  tryPush(payload: DataTrackFrame['payload']) {
+    try {
+      return this.manager.tryProcessAndSend(this.info.pubHandle, payload);
+    } catch (err) {
+      // NOTE: wrapping in the bare try/catch like this means that the Throws<...> type doesn't
+      // propegate upwards into the public interface.
+      throw err;
+    }
   }
 }


src/room/data-track/outgoing/errors.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  /** No additional data tracks can be published by the local participant. */
+  LimitReached = 3,
+
+  /** Cannot publish data track when the room is disconnected. */
+  Disconnected = 4,
+
+  // FIXME: get rid of internal error concept, this is just represented as bare throws in js
+  // Internal = 5,
+
```
1egoman: I have opted to get rid of the `DataTrackPublishErrorReason.Internal` error reason from the rust implementation, and instead just throw these errors seperately.

src/room/data-track/outgoing/errors.ts
```diff
@@ -0,0 +1,159 @@
...
+  // FIXME: this was introduced by web / there isn't a corresponding case in the rust version.
+  // Upon further reflection though I think this should exist in rust.
+  Cancelled = 6,
```
1egoman: I have added a new `DataTrackPublishErrorReason.Cancelled` reason which is used when publishing a data track is aborted via an `AbortSignal`.

I think this might be potentially a good idea to add into the rust implementation as well, as I could see allowing cancellation of publishing a data track to be a nice behavior to pass up to the caller on the associated data track creation `room` method.

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
```diff
@@ -0,0 +1,300 @@
...
+  info: DataTrackInfo;
+
+  pipeline: DataTrackOutgoingPipeline;
+};
+// FIXME: rust doesn't have this unpublishing descriptor, is it a good idea?
+export type UnpublishingDescriptor = {
+  type: 'unpublishing';
+  completionFuture: Future<void, never>;
+};
+export type Descriptor = PendingDescriptor | ActiveDescriptor | UnpublishingDescriptor;
```
1egoman: I have added another descriptor case here that the rust implementation doesn't have - `unpublishing`. This allows tracking that an unpublish is in progress and also means that the caller can get feedback on when this is complete. Also doing this means that if the SFU couldn't unpublish a data track (the rust implementation makes it seem like this isn't possible right now) it would be possible to expose that error back to a user somehow.

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
```diff
@@ -0,0 +1,300 @@
...
+      // FIXME: catch and log errors instead of rethrowing? That is what the rust implementation
+      // is doing instead.
+      // process_frame(...).inspect_err(|err| log::debug!("Process failed: {}", err))
+      // event_out_tx.try_send(...).inspect_err(|err| log::debug!("Cannot send packet to transport: {}", err));
+      //
+      // In the rust implementation this "dropped" error means something different (not enough room
+      // in the track mpsc channel)
+      throw DataTrackPushFrameError.dropped(err);
+    }
+  }
```
1egoman, line 154: The behavior of this function is a bit different from the rust implementation right now, and I'm unsure exactly how it would be best to standardize this. The closest equivilent rust code is in `DataTrack<Local>::try_push` [here](https://github.com/livekit/rust-sdks/blob/8e4f0bc9912b04725e68522832b7898aaac65998/livekit-datatrack/src/local/mod.rs#L83-L91) and `TrackTask::process_and_send` [here](https://github.com/livekit/rust-sdks/blob/8e4f0bc9912b04725e68522832b7898aaac65998/livekit-datatrack/src/local/manager.rs#L298-L312).

In rust, all the error cases here are logged (see the comments in the above code + the `TrackTask::process_and_send` link). The `dropped` error is returned when the `frame_tx.try_send(...)` in `try_push` fails.

In js, there is no `frame_tx` / `TrackTask` / etc because there isn't a need to run this across multiple processes (io is being handled in a non-blocking fashion and the computational requirements are minimal). So instead I opted to expose this `dropped` error as a wrapper around the debug logged errors.

I think the behavior should be aligned here, but if I go with the rust implemenattion's approach then I think there would never be a case where `dropped` gets thrown in the js implementation.

src/room/data-track/handle.ts
```diff
@@ -57,24 +56,20 @@ export class DataTrackHandle {
...
-    return new DataTrackHandle(raw);
-  }
-
-  constructor(raw: number) {
-    this.value = raw;
-  }
-}
+    return raw;
+  },
+};
```
1egoman, line 61: This may prove controversial, but got rid of `DataTrackHandle` as a wrapper class and instead defined a `type` shadowing a `const` ([docs about this](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)) so I could continue to get the `fromNumber` static method, and could get the documentation benefits of being able to label the type as `DataTrackHandle` rather than `number`.

The reason I did it is because I wanted to key a map by a track handle, and I couldn't look up by this given the reference of the wrapper object would change every time it was newly constructed. ie:
```js
class Wrapper { constructor(inner) { this.inner = inner; } }

const map = new Map();
map.set(new Wrapper(5), 123);
map.get(new Wrapper(5)); // -> undefined
```

src/room/data-track/track.ts
```diff
@@ -0,0 +1,56 @@
...
+  pubHandle: DataTrackHandle;
+  name: String;
+  usesE2ee: boolean;
+};
+
+export class LocalDataTrack {
+  info: DataTrackInfo;
+
+  protected manager: OutgoingDataTrackManager;
+
```
1egoman, line 19: Note that this `LocalDataTrack` implementation is just a stub at the moment, I think there's some more evolution / abstraction to be done (which will occur with the introduction of `RemoteDataTrack` in the next pull request).

Also right now this is fully outside of the existing track / publication hierarchy. Some more thinking needs to be done on what the long term plan is here and how to start putting in place the pieces to support migrating to a publication-less track world across the board eventually.

src/room/utils.ts
```diff
@@ -457,8 +458,12 @@ export function getStereoAudioStreamTrack() {
...
 
+/** An object that represents a serialized version of a `new Promise((resolve, reject) => {})`
+ * constructor. Wait for a promise resolution with `await future.promise` and explicitly resolve or
+ * reject the inner promise with `future.resolve(...)` or `future.reject(...)`.
+ */
 export class Future<T, E extends Error> {
-  promise: Promise<T>;
+  promise: Promise<Throws<T, E>>;
 
   resolve?: (arg: T) => void;
```
1egoman, line 468: Just wanted to call out explicitly since it's a fairly important change - `Future` now uses the `Throws<...>` brand type internally, and anywhere a `future.promise` is `await`ed (assuming you are in a function which returns a `Throws` branded type) will now result in a linting error.

Because of that last limitation (throws-transformer is opt in within functions that return `Throws<...>`) this is scoped to data tracks code / doesn't cause errors all over the place.

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
```diff
@@ -0,0 +1,305 @@
...
+   *
+   * If none, end-to-end encryption will be disabled for all published tracks.
+   */
+  encryptionProvider?: EncryptionProvider;
+};
+
+/** How long to wait when attempting to publish before timing out. */
+const PUBLISH_TIMEOUT_MILLISECONDS = 10_000;
+
+export default class OutgoingDataTrackManager extends (EventEmitter as new () => TypedEmitter<DataTrackOutgoingManagerCallbacks>) {
```
1egoman, line 89: I called it `OutgoingDataTrackManager` to match with the `OutgoingDataStreamManager`. Also @ladvoc discussed this in a 1:1 and he was good with this name. For reference, the rust implementation (in the `livekit-datatrack` crate) calls this `local::Manager`.

I am not sure what the end name should be, but IMO given these two entities have very close to the same behavior, it would be good to align on a name.

src/room/data-track/outgoing/errors.ts
```diff
@@ -0,0 +1,159 @@
...
+  Reason extends DataTrackPublishErrorReason,
+> extends LivekitReasonedError<Reason> {
+  readonly name = 'DataTrackPublishError';
+
+  reason: Reason;
+
+  reasonName: string;
+
+  constructor(message: string, reason: Reason, options?: { cause?: unknown }) {
+    super(21, message, options);
```
ladvoc, line 38: question: What does the constant 21 correspond to here?

src/room/data-track/outgoing/OutgoingDataTrackManager.test.ts
```diff
@@ -0,0 +1,373 @@
...
+    const unpublishEvent = await managerEvents.waitFor('sfuUnpublishRequest');
+    expect(unpublishEvent.handle).toStrictEqual(6);
+
+    // Acknowledge that the unpublish has occurred
+    manager.receivedSfuUnpublishResponse(DataTrackHandle.fromNumber(6));
+
+    await shutdownPromise;
+  });
+
+  // FIXME: add e2ee tests
```
ladvoc: praise: These test cases look great! Will plan on porting them to the Rust implementation.

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
ladvoc: This is a good point for discussion. From the user perspective, I thought fire and forget semantics made more sense for unpublishing in the general case. Also, Rust has a separate [method](https://github.com/livekit/rust-sdks/blob/8e4f0bc9912b04725e68522832b7898aaac65998/livekit-datatrack/src/track.rs#L54) that allows awaiting unpublish if that is desired (which can also be used to detect an SFU-initiated unpublish). However, I don't have a strong opinion here.

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
ladvoc: Additional context for other reviewers: the only potential downside of this naming convention is the terms "local" and "remote" are still used in the public API (e.g. LocalTrack, RemoteTrack). However, since the track managers are internal components, I don't think this necessarily needs to be consistent. Will plan on renaming in Rust if there are no objections.

src/room/data-track/track.ts
ladvoc: For your next PR, it might make sense to have a common base class `DataTrack` which local and remote inherit from. This would mirror the Rust implementation (putting aside the language specifics) by allowing common methods like `is_published` to be accessible for any data track. For v3 breaking changes, we will want to consider if it makes sense to have a single base class for all tracks (both media and data).

src/room/data-track/track.ts
```diff
@@ -0,0 +1,56 @@
...
+    return this.descriptor?.type === 'active';
+  }
+
+  /** Try pushing a frame to subscribers of the track.
+   *
+   * Pushing a frame can fail for several reasons:
+   *
+   * - The track has been unpublished by the local participant or SFU
+   * - The room is no longer connected
+   * - Frames are being pushed too fast (FIXME: this isn't the case in the js implementation?)
```
ladvoc: comment: Yes, in the Rust implementation, frames get dropped at this point if the _mpsc_ channel used to send frames to the pipeline is full. I don't think this has an equivalent in JS.

src/room/data-track/handle.ts
ladvoc: Makes sense to me. Generally, I think it is ok to diverge from the patterns in the Rust implementation if they introduce complications here.

src/room/data-track/outgoing/types.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/** Request sent to the SFU to unpublish a track. */
+export type OutputEventSfuUnpublishRequest = {
+  handle: DataTrackHandle;
+};
+
+/** Serialized packets are ready to be sent over the transport. */
+export type OutputEventPacketsAvailable = {
+  bytes: Uint8Array;
+  signal?: AbortSignal;
```
ladvoc, line 36: question: Curious how the abort signal works in this context.

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
ladvoc: I agree, I think the Rust SDK needs to have better error reporting at the call site. Let's discuss this more in our 1:1.

ladvoc (approved):
Generally looks good to me!

src/room/data-track/track.ts
1egoman: We're thinking similarly, yes! Lukas and I are going to discuss this later this week.

src/room/data-track/handle.ts
1egoman: Sounds good.

FWIW, this is more of a comment for @lukasIO, last time I used this pattern he wasn't a fan.

src/room/data-track/outgoing/types.ts
1egoman: Right now, when calling `LocalDataTrack.tryPush`, an abort signal can be passed in which allows the caller to cancel sending the frame. This is where that abort signal is emitted down to the transport layer to allow it to respond to any cancellations.

I mostly added this out of habit, but digging into this a bit more, it doesn't look like the underlying [`RTCDataChannel.send` api](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/send) supports cancellation so I will get rid of it.

1egoman: Removed in 389d49a

src/room/data-track/outgoing/errors.ts
1egoman: The js sdk at the moment has a half baked error code system for errors built into the base `LiveKitError`, and this is trying to satisfy this constraint (21 is the next one in the series).

IMO it is worth ripping this out eventually, hopefully alongside coming up with standardized error codes across all sdks.

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
lukasIO: Both sound fine to me. 
I like `Outgoing` for being very explicit about what it handles and like `Local` mainly for consistency with the public constructs

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
1egoman: From 1:1 -
- @ladvoc rename dropped -> queueFull or something similar in rust
- Consider renaming `try_push`

src/room/data-track/outgoing/errors.ts
1egoman: From 1:1 - this will be a web specific case for now, and the rust implicit cancellation behavior will be documented.

src/room/data-track/outgoing/OutgoingDataTrackManager.ts
1egoman: From 1:1 - update the web implementation to not have the extra state (and maybe add a `unpublishingFuture` type field to the active descriptor which would contain the `completionFuture` state).

In rust potentially make `wait_for_unpublish` (method linked) return a `Result<...>` to capture that errors can occur.

1egoman: Done here: 060155d

PR #1812, opened 2026-02-18
User Timestamp Parsing

chenosaurus:


src/e2ee/worker/FrameCryptor.ts
```diff
@@ -454,6 +461,32 @@ export class FrameCryptor extends BaseFrameCryptor {
...
+              rtpTimestamp: packetTrailerResult.rtpTimestamp,
+            },
+          };
+          postMessage(msg);
+        }
+      } catch {
+        // Best-effort: never break media pipeline if timestamp parsing fails.
+      }
+    }
+
```
1egoman, line 489: question: Is the `FrameCryptor` (which I believe runs as part of the e2ee worker, which may not be loaded if e2ee is not enabled) the best place for this logic to live?

Is there a reason it has to run in a web worker context (ie, maybe it is super computationally expensive / etc)? Is this feature supposed to work if e2ee is disabled?

src/e2ee/E2eeManager.ts
```diff
@@ -231,6 +241,32 @@ export class E2EEManager
...
 
+  private handleUserTimestamp(
+    trackId: string,
+    participantIdentity: string,
+    timestampUs: number,
+    frameId?: number,
+    rtpTimestamp?: number,
+  ) {
+    if (!this.room) {
+      return;
```
1egoman, line 252: question: Similar to my last comment - I don't know a ton about the user timestamp feature, but my suspicion is that putting `handleUserTimestamp` inside the `E2eeManager` is mixing concerns unnecesarily.

src/packet_trailer/packetTrailer.worker.ts
```diff
@@ -0,0 +1,106 @@
+/**
+ * Lightweight worker for stripping LKTS packet trailers from inbound
+ * encoded video frames via RTCRtpScriptTransform.
+ *
+ * When a valid trailer is found, the metadata is posted back to the main
+ * thread so the SDK can store the RTP-to-frame-metadata mapping on the
+ * corresponding RemoteVideoTrack.
+ */
+import {
```
1egoman, line 9: question: I'm a bit confused now - so this change is introducing a second webworker exclusively for packet trailer stripping, but somehow this worker ties into the existing E2eeManager / e2ee worker? Assuming this isn't in an incomplete state mid migration from one approach to another, I might need some more background context to better understand this.

As a heads up - introducing another webworker is a fairly significant change. I think it might be viable because this packet trailer stuff is fully additive, but it's something we've used sparingly, so I just want to make sure there's a good reason to use it here.

More related background context I wrote in another context where somebody was proposing introducing a new webworker, maybe you would find it useful:
> Starting a webworker requires that you pass a url to another script that you want to run to the `new Worker("https://path.to/worker/script.js")` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker)) constructor when creating the new worker, but because livekit is a library we don't have control over the end user's build process / have no way to force them to generate another file in their end application, let alone get a predictable path to where it might be hosted.
> 
> The one place that is using a webworker right now in the web sdk is end to end encryption, and the way it gets around this problem (and the way we'd have to do this for compression) is by bubbling up the web worker construction to the end user, forcing them to build / host a separate "worker script" containing the e2ee worker code, call `new Worker("...")` + pass the url to this "worker script", and then pass the fully constructed Worker object into the main Livekit SDK. For example, [this is how this is done in the client-sdk-js demo app](https://github.com/livekit/client-sdk-js/blob/f54d243a2ea3a46597c61879e42257a6fb4eca58/examples/demo/demo.ts#L2C8-L2C18). This is super cumbersome / not an easy thing to instruct a user to do.

src/room/track/RemoteVideoTrack.ts
```diff
@@ -55,6 +71,65 @@ export default class RemoteVideoTrack extends RemoteTrack<Track.Kind.Video> {
...
+   *
+   * @param timestampUs - user timestamp in microseconds since Unix epoch
+   * @param rtpTimestamp - RTP timestamp from the encoded frame (90 kHz clock).
+   *   When provided, the mapping is stored so decoded frames can look up their
+   *   frame metadata via {@link lookupFrameMetadata}.
+   * @param frameId - monotonic frame id extracted from the trailer.
+   *
+   * @internal
+   */
+  setUserTimestamp(timestampUs: number, rtpTimestamp?: number, frameId?: number) {
```
1egoman, line 87: nitpick: As a heads up, in data tracks, the `userTimestamp` field is a `bigint`. It might make sense to use that here too.

src/room/track/RemoteVideoTrack.ts
```diff
@@ -55,6 +71,65 @@ export default class RemoteVideoTrack extends RemoteTrack<Track.Kind.Video> {
...
+      }
+      this.userTimestampMap.set(rtpTimestamp, { timestampUs, frameId });
+      this.userTimestampMapOrder.push(rtpTimestamp);
+    }
+  }
+
+  /**
+   * Look up the frame metadata associated with a given RTP timestamp.
+   * The entry is consumed (removed) after lookup, matching the Rust SDK
+   * behaviour.
```
1egoman, line 103: nitpick: I suspect this was probably LLM generated with a prompt like "port the rust implementation over to the web sdk" - it would be good to strip out some of these egregious comparisons to the rust sdk before this would end up getting merged.

question: Also in this context, is actually the "consuming the entry like rust" type behavior a good idea? In rust there are benefits to avoid extra cloning, but it has the serious tradeoff of making requests non idempotent. Given the web doesn't have these same data ownership concerns it is a fairly serious rough edge for from what I can tell no real benefit, assuming there is some other way the entries in the `userTimestampMap` are cleared so it doesn't grow in an unbounded fashion.

src/room/Room.ts
```diff
@@ -123,6 +127,10 @@ export enum ConnectionState {
   SignalReconnecting = 'signalReconnecting',
 }
 
+function createPacketTrailerWorker(): Worker {
+  return new Worker(new URL('./livekit-client.packet-trailer.worker.js', import.meta.url));
```
1egoman, line 136: issue: See my comment [here](https://github.com/livekit/client-sdk-js/pull/1812/changes#r2962642625) - this type of thing isn't viable and won't work for all downstream consumers of the package transparently.

1egoman (commented):
It might be possible I'm reviewing this prematurely - Shijing had asked I take a look at this. That being said, IMO there's still quite a bit of work to be done here and I'm not sure I'm fully following how this all works right now.

chenosaurus:
closing this, moving features to new PR https://github.com/livekit/client-sdk-js/pull/1886

PR #1813, opened 2026-02-18
Prevent ongoing renegotiations from declaring the negotiation as timed out

lukasIO:


boks1971 (approved):
Great catch!

Don't fully understand the nuances of all the timing here, but overall lgtm!

PR #1819, opened 2026-02-20
Data tracks - incoming manager

1egoman:
This pull request builds on the outgoing manager [here](https://github.com/livekit/client-sdk-js/pull/1810) and introduces a new "incoming manager" which is used for ingesting events from remote participants which are publishing data tracks.

### DIFF
--- a/src/room/data-track/incoming/IncomingDataTrackManager.test.ts
+++ b/src/room/data-track/incoming/IncomingDataTrackManager.test.ts
@@ -402,7 +402,9 @@ describe('DataTrackIncomingManager', () => {
       expect(start.subscribe).toBe(true);
 
       // Make sure cancellation is immediately bubbled up
-      await expect(subscribeRequestPromise).rejects.toStrictEqual(DataTrackSubscribeError.cancelled());
+      await expect(subscribeRequestPromise).rejects.toStrictEqual(
+        DataTrackSubscribeError.cancelled(),
+      );
 
       // Make sure that there is immediately another "unsubscribe" sent
       const end = await managerEvents.waitFor('sfuUpdateSubscription');


src/room/data-track/incoming/IncomingDataTrackManager.test.ts
```diff
@@ -0,0 +1,391 @@
...
+        `PrefixingDecryptionProvider: first four bytes of payload were not 0xdeadbeef, found ${p.payload.slice(0, 4)}`,
+      );
+    }
+    return p.payload.slice(4);
+  },
+};
+
+describe('DataTrackIncomingManager', () => {
+  describe('Track publication', () => {
+    it('should test track publication additions / removals', async () => {
```
1egoman, line 35: Note to self: these tests aren't super thorough right now. I'm testing the happy paths but that's mostly about it. Look at the rust implementation and port over any of those tests. Plus also add more around non happy path in the subscription workflow.

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,483 @@
...
+   */
+  async subscribeRequest(
+    sid: DataTrackSid,
+    signal?: AbortSignal,
+  ): Promise<
+    Throws<
+      ReadableStream<DataTrackFrame /* FIXME: should this be a frame? or just a packet? */>,
+      DataTrackSubscribeError
+    >
+  > {
```
1egoman: Unlike data streams, this isn't emitting a custom object which has other metadata on it, a `readAll` method, etc. Especially given "read all" doesn't make sense for a non deterministic length subscription I think this is fine, and in the future we could always return a subclass of `ReadableStream` here worst case so I think this isn't locking us out of further extension. But maybe there are benefits of a custom object I am overlooking.

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,483 @@
...
+      // FIXME: maybe this should be a DataTrackSubscribeError.disconnected()? That's what happens
+      // here (on the caller end in the rust implementation):
+      // https://github.com/livekit/rust-sdks/blob/ccdc012e40f9b2cf6b677c07da7061216eb93a89/livekit-datatrack/src/remote/mod.rs#L81
+      throw DataTrackSubscribeError.disconnected();
+
+      // FIXME: DataTrackSubscribeError.unpublished is unused both here and in rust
+
+      // @throws-transformer ignore - this should be treated as a "panic" and not be caught
+      // throw new Error('Cannot subscribe to unknown track');
+    }
```
1egoman, line 118: Note to self: I need to clean this up, I'm not 100% sure I got the error behavior matched from rust here.

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,483 @@
...
+    for (const controller of descriptor.subscription.streamControllers) {
+      controller.close();
+    }
+
+    // FIXME: this might be wrong? Shouldn't this only occur if it is the last subscription to
+    // terminate?
+    const previousDescriptorSubscription = descriptor.subscription;
+    descriptor.subscription = { type: 'none' };
+    this.subscriptionHandles.delete(previousDescriptorSubscription.subcriptionHandle);
+
```
1egoman, line 317: This matches the behavior of the [rust implementation](https://github.com/livekit/rust-sdks/blob/45531ef1bdb8d9559ec2817381a732b1d1ae86e0/livekit-datatrack/src/remote/manager.rs#L154-L155) but by eye I think something might be off here, I would think that this code should only run on the final subscription termination, not on _any_ subscription termination.

src/room/data-track/depacketizer.ts
```diff
@@ -99,13 +99,7 @@ export default class DataTrackDepacketizer {
   push(
     packet: DataTrackPacket,
     options?: PushOptions,
-  ): Throws<
-    DataTrackFrame | null,
-    | DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.Interrupted>
-    | DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.BufferFull>
-    | DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.UnknownFrame>
-    | DataTrackDepacketizerDropError<DataTrackDepacketizerDropReason.Incomplete>
-  > {
```
1egoman, line 108: Note mostly for @lukasIO - I have made some significant enhancements to the `throws-transformer` thing being investigated for use more widely in the SDK, and in particular now it handles generic errors properly, so I can drop all of these individual error cases and just replace each with the single error type.

In addition, there's now some rudimentary handling of `new Promise` constructors and promise handling with `then` / `catch`.

src/room/data-track/e2ee.ts
```diff
@@ -10,5 +10,6 @@ export type EncryptionProvider = {
 };
 
 export type DecryptionProvider = {
-  decrypt(payload: Uint8Array, senderIdentity: string): Uint8Array;
+  // FIXME: add in explicit `Throws<..., DecryptionError>`?
+  decrypt(payload: EncryptedPayload, senderIdentity: string): Uint8Array;
 };
```
1egoman, line 15: Note to self: Leave this as is for now, and in the next / final "integration" pull request address this fixme comment and add better explicit error cases here.

For reference, what the rust implementation does: https://github.com/livekit/rust-sdks/blob/45531ef1bdb8d9559ec2817381a732b1d1ae86e0/livekit-datatrack/src/e2ee.rs#L34-L37

src/room/data-track/track-interfaces.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const TrackSymbol: symbol = Symbol.for('lk.track');
+
+export interface ITrack {
+  readonly trackSymbol: typeof TrackSymbol;
+}
+
+export function isTrack(subject: unknown): subject is ITrack {
+  return isObject(subject) && 'trackSymbol' in subject && subject.trackSymbol === TrackSymbol;
+}
```
1egoman, line 15: @lukasIO Here's a first draft of the track interfaces we had discussed last week, any thoughts are welcome!

My initial plan is not to export any of these yet outside the package yet as they will be very confusing - just right now to ensure the symbols are included in the classes so it can be "switched on" at a future date.

.changeset/heavy-geckos-wave.md
lukasIO, line 1: do we need a changeset for this? 

As these are internal changes only for now, I think a changeset at integration time for "adding data tracks" would suffice?

src/room/data-track/track-interfaces.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/** An interface representing a track (of any type) which is local and sending data to the SFU. */
+export interface ILocalTrack extends ITrack {
+  readonly isLocal: true;
+
+  isPublished(): boolean;
```
lukasIO, line 21: is a boolean sufficient here? Wondering about pending/intermittent states of "publishing" (and potentially also "unpublishing")

src/room/data-track/track-interfaces.ts
lukasIO: yeah, the naming right now would be (and potentially already is) very confusing. 

E.g. already have `isLocalTrack` helpers, which sounds like a rather easy-to-make mistake if we're checking for different track constructs altogether and the only differentiation is the import path. 

Maybe don't export the generic ones at all from this file and only export `isDataTrack` for now?

.changeset/heavy-geckos-wave.md
1egoman: Sure, I can remove it. Though worth noting, I had been including changesets in the other pull requests in the series that have been merged so far so it will be inconsistent.

src/room/data-track/track-interfaces.ts
1egoman: Good idea, yea that makes it a little better - updated!

src/room/data-track/track-interfaces.ts
1egoman: The way I was thinking about this, `isPublished` is something very minimal all local tracks should be able to fulfill but I wanted to minimize the amount of frontloaded assumptions this interface contains since we can always add to it later.

I think if you think it makes sense to expose more than just a single "published" boolean, I'd prefer to take a page out of the session api playbook and expose a few `is`-prefixed getters which each future internal track implementation could map to whatever internal state it manages (rather than exposing something like an enum with `publishing` / `published` / `unpublishing` states). And of course each track implementation could expose much richer apis based on the exact nuances / dependencies in each's publishing workflow.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: I've opted to hold off on cleaning this up until the integration pull request [here](https://github.com/livekit/client-sdk-js/pull/1820/changes/696882c5ec6d0d429f733359e15b1923cfe35b23). I did some thinking and yes I think `disconnected` is the right error here.

src/room/data-track/e2ee.ts
1egoman: Update: I've changed this quite substantially in the [integration pull request](https://github.com/livekit/client-sdk-js/pull/1820) - now it just uses `E2eeManager` directly instead of these intermediate interfaces. The rust implementation added these because e2ee logic is a lot more fragmented there but this sdk doesn't have that problem. So this comment is no longer relevant.

src/room/data-track/incoming/IncomingDataTrackManager.ts
lukasIO: I think you're right. The necessary metadata already lives on the Track itself, so from a users perspective having a "raw" readable stream returned on subscribe sounds nice

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,492 @@
...
+          return;
+        }
+        currentDescriptor.subscription.pendingRequestCount -= 1;
+        if (currentDescriptor.subscription.pendingRequestCount <= 0) {
+          // No requests are still pending, so cancel the underlying pending `sfuUpdateSubscription`
+          currentDescriptor.subscription.cancel();
+        }
+      };
+
+      abortSignal?.addEventListener('abort', onAbort);
```
lukasIO: just for safety it would make sense to check for `abortSignal.aborted` here before adding the event listener

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,492 @@
...
+                    DataTrackSubscribeError.cancelled(),
+              );
+            }
+          },
+        };
+
+        this.emit('sfuUpdateSubscription', { sid, subscribe: true });
+
+        const timeoutSignal = AbortSignal.timeout(SUBSCRIBE_TIMEOUT_MILLISECONDS);
+        const combinedSignal = signal ? AbortSignal.any([signal, timeoutSignal]) : timeoutSignal;
```
lukasIO: I believe both AbortSignal.any and AbortSignal.timeout have not-so great browser support right now, see https://github.com/livekit/client-sdk-js/pull/1700

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,492 @@
...
+
+        // Wait for the subscription to complete, or time out if it takes too long
+        const reader = await waitForCompletionFuture(descriptor, combinedSignal);
```
lukasIO: do we have any means to also react to participant's disconnection here? both the local participant disconnecting and the remote participant disconnecting should probably error out early

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,492 @@
...
+        return this.createReadableStream(sid);
+      }
+    }
+  }
+
+  /** Allocates a ReadableStream which emits when a new {@link DataTrackFrame} is received from the
+   * SFU. */
+  private createReadableStream(sid: DataTrackSid) {
+    let streamController: ReadableStreamDefaultController<DataTrackFrame> | null = null;
+    return new ReadableStream<DataTrackFrame>({
```
lukasIO: we should set an appropriate queuing strategy on the readable stream and
also potentially start to drop frames if the backpressure gets too high 🤔

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,492 @@
...
+      if (descriptor.subscription.type === 'none') {
+        continue;
+      }
+      this.emit('sfuUpdateSubscription', { sid, subscribe: true });
+    }
+  }
+
+  /** Shutdown the manager, ending any subscriptions. */
+  shutdown() {
+    for (const descriptor of this.descriptors.values()) {
```
lukasIO, line 525: should the descriptors also be cleared here afterwards?

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: Sounds good. FWIW they have been [widely available since 2024](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static) but it wouldn't be hard to use a polyfill so I'll just do that.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: Good point. I also fixed this in the `OutgoingDataTrackManager` at the same time and added tests exercising both.

Also when digging into this I found another fairly nuanced / gnarly subscription management edge case where cancelling an abortsignal on one subscription would propagate to others. I think I've addressed it though and have a test for this case as well.

Just generally, if you can I'd love a second set of eyes going through the test cases and if you can think of any other ones for subscriptions specifically (that's probably the most nuanced part of this change) feel free to suggest additional ones. I still have a bit of work to add more of these.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: Good point - I added an event in 6cbf73e which the room can send in when a remote participant disconnects and throw a new `DataTrackSubscribeError.disconnected()` error, along with an associated test case for both this when a subscription is pending and active.

also cc @ladvoc because the rust implementation probably also has this problem.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: Good point, fixed in 4c7b39c.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: Yea good point, I added something in 6bdd819. But FYI that other readable streams like [the ones that power data streams](https://github.com/livekit/client-sdk-js/blob/73932e66d75dae63a97fe1c1e6e5348042020e6e/src/room/data-stream/incoming/IncomingDataStreamManager.ts#L139-L158) don't have this, so it sounds like maybe there needs to be a pass done through everything to add it.

also cc @ladvoc, because this is similar to a discussion we had in regards to the rust implementation - the default channel buffer length is I think 4 and this should probably be configurable externally somehow.

lukasIO: Yeah, that's true. We should arguably add something for data streams as well. What makes it less significant - in theory - for data streams is that they are supposed to have a defined end. 

Thanks for adding that queuing strategy. Should we maybe use a default that's a bit higher? 4 seems very low, but not sure either about the exact use cases. 
ideally we'd want something that is a sane default that doesn't require adjustments from 90% of users

1egoman: I went with `4` because that is what the rust implementation had used for its internal mpsc channel which acts in a similar fashion, but it looks like that might have been updated to `16` and I missed that ([link](https://github.com/livekit/rust-sdks/blob/bef4a107cc28a22d872e0f4a3a831f40bfde31a8/livekit-datatrack/src/local/manager.rs#L232)), so I made that update!

If you think it should be higher, I'm open to it, but It's a hard thing to pick a non super high default for. The optimal value is heavily dependent on the sample rate of data being received, the size of each sample, and the speed at which samples are being consumed. So either we pick a super high default and users just assume it is unlimited until they hit an edge case and realize it isn't, or we pick a lower threshold and users may run into the barrier faster but it means they have to be more strategic about what it is configured to for their specific use case. Also cc @ladvoc in case you have any thoughts.

I think my preference would be to either keep it unlimited by default, pick some fairly conservative threshold, or (ideally but most complex) to somehow expose some function which can be fed in derived metrics and which could compute an optimal high water mark value.

No matter what we choose, both implementations should be updated in lock step so they have similar behavior.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: @ladvoc FYI, could you take a look at this and see if it looks right to you?

src/room/data-track/incoming/IncomingDataTrackManager.test.ts
1egoman: I've added some more since originally writing this comment - I still think it could be better but I think I'm going to call it good enough for now. All the really nuanced local subscription <-> sfu subscription reconciliation cases I can think of are covered and I think those are the most important.

src/room/data-track/incoming/IncomingDataTrackManager.ts
ladvoc: That's correct, this should only be sent on the final subscription termination (i.e., the client no longer wishes to be forwarded frames for the track). In the Rust SDK, `UnsubscribeEvent` is emitted from the track task [here](https://github.com/livekit/rust-sdks/blob/0aa941217b1c6428ce7914fc987f6c2053fff36d/livekit-datatrack/src/remote/manager.rs#L402) when the frame channel is closed, which occurs when there are no more receivers.

src/room/data-track/incoming/errors.ts
```diff
@@ -0,0 +1,57 @@
...
+  Reason extends DataTrackSubscribeErrorReason = DataTrackSubscribeErrorReason,
+> extends LivekitReasonedError<Reason> {
+  readonly name = 'DataTrackSubscribeError';
+
+  reason: Reason;
+
+  reasonName: string;
+
+  constructor(message: string, reason: Reason, options?: { cause?: unknown }) {
+    super(22, message, options);
```
ladvoc, line 24: nitpick: Maybe define `22` as a constant somewhere with a comment explaining what it is?

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,554 @@
...
+type SfuUpdateSubscription = {
+  /** Identifier of the affected track. */
+  sid: DataTrackSid;
+  /** Whether to subscribe or unsubscribe. */
+  subscribe: boolean;
+};
+
+export type DataTrackIncomingManagerCallbacks = {
+  /** Request sent to the SFU to update the subscription for a data track. */
+  sfuUpdateSubscription: (event: SfuUpdateSubscription) => void;
```
ladvoc: question: This is a TypeScript-specific concern, and I’m not sure what the usual convention is, but should the event payloads all be typed consistently—either with standalone type definitions or all inline?

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,554 @@
...
+   * requests simply attach an additional receiver to the broadcast channel, allowing them to consume
+   * frames from the existing subscription pipeline.
+   */
+  async subscribeRequest(
+    sid: DataTrackSid,
+    signal?: AbortSignal,
+    highWaterMark = READABLE_STREAM_DEFAULT_HIGH_WATER_MARK,
+  ): Promise<
+    Throws<
+      ReadableStream<DataTrackFrame /* FIXME: should this be a frame? or just a packet? */>,
```
ladvoc: Addressing your FIXME: users only ever work with frames in the public API, so this should be frame.

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,554 @@
...
+      DataTrackSubscribeError
+    >
+  > {
+    const descriptor = this.descriptors.get(sid);
+    if (!descriptor) {
+      // FIXME: maybe this should be a DataTrackSubscribeError.disconnected()? That's what happens
```
ladvoc: For context, in Rust, the disconnected reason exists because of how async channels work. For example, when subscribing to a track, it needs to get a strong reference to the channel (if it is still open) in order to send the `SubscribeRequest` to the manager. Do you think it aligns well enough semantically with the usage here to use the same error reason?

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -0,0 +1,554 @@
...
+        continue;
+      }
+      switch (descriptor.subscription.type) {
+        case 'none':
+          break;
+        case 'pending':
+          descriptor.subscription.completionFuture.reject?.(DataTrackSubscribeError.disconnected());
+          break;
+        case 'active':
+          this.unSubscribeRequest(descriptor.info.sid);
```
ladvoc, line 517: question: Handling remote participant disconnected events is not apart of the Rust implementation yet, so I'm curious about the logic here. In this case, do we need/want to explicitly unsubscribe?

ladvoc (approved):
LGTM ✅, logic looks consistent with the Rust implementation besides a few places I comment on.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: See the discussion here in a comment above: https://github.com/livekit/client-sdk-js/pull/1819#discussion_r2854396858

TL;DR - This was something @lukasIO surfaced as being important so I added it and it probably should also make its way into the rust implementation as well.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: Thank you for pointing this out! I will refactor this and move all of the payloads to explicit named types rather than being typed inline. I don't think it matters all that much (this is an internal interface) but I agree consistency would be good.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: Reading through this again - it looks like this is a more relevant error case: https://github.com/livekit/rust-sdks/blob/ccdc012e40f9b2cf6b677c07da7061216eb93a89/livekit-datatrack/src/remote/manager.rs#L118-L120. Because of that, I think I will switch this back to just throwing a bare `Error`.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: I think I will keep this as is for now, and we can discuss it further in the [integration pull request](https://github.com/livekit/client-sdk-js/pull/1820) if need be.

PR #1820, opened 2026-02-20
Data track integration

1egoman:
Integrates the data track managers (both incoming and outgoing) into the existing sdk `Room` and `RemoteParticipant` interfaces. Once this is merged, data tracks is fully implemented!

## Interface - Publishing
```tsx
const localDataTrack = await room.publishDataTrack({name: "data track name"});
await localDataTrack.tryPush(new Uint8Array(/* ... */));

// Once you are done with the local data track, it can be unpublished:
await localDataTrack.unpublish();
```

## Interface - Subscribing
```tsx
// There are two ways to get a remote data track, either a room event:
room.on(RoomEvent.RemoteDataTrackPublished, async (remoteDataTrack) => {
  // ...
});

// Or a new field on `RemoteParticipant`:
const remoteDataTrack = remoteParticipant.dataTracks.get("track name"); // Get a data track which has been published
const remoteDataTrack = await remoteParticipant.dataTracks.getDeferred("track name"); // Get a data track which _will_ be published shortly

// Either way, once you have a remoteDataTrack, you can subscribe:
const stream = await remoteDataTrack.subscribe(/* optional abort signal */);
for await (const frame of stream) {
  console.log(`Received bytes from ${remoteDataTrack.info.name}:`, frame.payload);
}
```

Todo:
- [x] Decide on how to show the example app chart vs hexdump - show both? Just one? [commit introducing chart](https://github.com/livekit/client-sdk-js/pull/1820/commits/4268beb60737be5e0852e4433acc795e8d48a22a).
- [x] Add changeset

Resolves CLT-2582

### DIFF
--- a/.changeset/metal-pugs-enjoy.md
+++ b/.changeset/metal-pugs-enjoy.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Adds data tracks support


src/utils/waitable-map.ts
```diff
@@ -0,0 +1,155 @@
...
+ * A Map-like container keyed by unique strings that supports the ability to wait
+ * for future keys to show up in the map.
+ *
+ * @example
+ * // An already existing key:
+ * const value = map.get("key");
+ * // Wait for a key which will be added soon:
+ * const value = await map.waitUntilExists("key");
+ */
+export class WaitableMap<K, V> implements Iterable<[K, V]> {
```
lukasIO: suggestion: for conciseness it appears this could also simply extend the built-in Map constructor and override the `set` method to resolve pending waits?

src/utils/waitable-map.ts
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * Returns the value for `key` immediately if it exists, otherwise returns a
+   * promise that resolves once `set(key, value)` is called.
+   *
+   * If an `AbortSignal` is provided and it is aborted before the key appears,
+   * the returned promise rejects with an {@link WaitableMapAbortError}.
+   */
+  waitUntilExists(key: K): Promise<V>;
+  waitUntilExists(key: K, signal: AbortSignal): Promise<Throws<V, WaitableMapAbortError>>;
+  async waitUntilExists(key: K, signal?: AbortSignal) {
```
lukasIO: nit: naming

alternative suggestions:
- `await(key)`
- `getAsync(key)`

My take is that `await(key)` might be the cleanest. The `getAsync` isn't really a common pattern in JS.

src/utils/waitable-map.ts
1egoman: I have a slight bias towards not extending native container types like that but I don't think I can articulate a strong advantage for why I have that bias, so sure I will make this change.

src/utils/waitable-map.ts
1egoman: I will admit, I am not a big fan of either of those naming suggestions. To me `await` on its own doesn't make it clear that there is "getting a key" involved, and I also am not a super big fan of the repetition of `await map.await("key")`.

Just to throw a few more options out there to see if we can find one we both like, here's some more potential options from a LLM:

- **`resolve(key)`** — Borrows from Promise terminology; implies the value will materialize.
- **`getDeferred(key)`** — Concise, borrows from the well-understood deferred/promise pattern.
- **`getEventually(key)`** — A little whimsical but honest about what's happening.
- **`getWhenAvailable(key)`** — Very explicit and self-documenting, though a bit long.
- **`getWhen(key)`** — Reads as "get when available," compact shortening of the above.
- **`getOnce(key)`** — Reads as "get once it's ready."

Thoughts on any of those?

src/utils/waitable-map.ts
1egoman: Done in c664504

src/room/data-track/LocalDataTrack.ts
```diff
@@ -0,0 +1,113 @@
...
+   * - The track has been unpublished by the local participant or SFU
+   * - The room is no longer connected
+   */
+  tryPush(payload: DataTrackFrame['payload']) {
+    if (!this.handle) {
+      throw DataTrackPushFrameError.trackUnpublished();
+    }
+
+    try {
+      return this.manager.tryProcessAndSend(this.handle, payload);
```
pblazej: Trying to understand different implementations now: is this async (with e2ee)?

src/room/participant/RemoteParticipant.ts
```diff
@@ -27,6 +29,14 @@ export default class RemoteParticipant extends Participant {
 
   trackPublications: Map<string, RemoteTrackPublication>;
 
+  /** A map of data track name to the corresponding {@link RemoteDataTrack}.
+   * @example
+   * // An already existing data track:
+   * const track = remoteParticipant.dataTracks.get("data track name");
+   * // Wait for a data track which will be published soon:
+   * const track = await remoteParticipant.dataTracks.waitUntilExists("data track name"); */
+  dataTracks: WaitableMap<RemoteDataTrack['info']['name'], RemoteDataTrack>;
```
pblazej: This one does not exist elsewhere, right?

1egoman: Correct, this is just a web specific interface decision to try to match with how other types of tracks work on `RemoteParticipant`. In the [pull request description](https://github.com/livekit/client-sdk-js/pull/1820#issue-3970281913) under `Interface - Subscribing` I show an example of how this works.

src/room/data-track/LocalDataTrack.ts
1egoman: Yes, this is an async call, just returning the promise outright!

-----------

For reference, if it is helpful, here is the path that data being sent outgoing from a `LocalDataTrack` takes:
- `LocalDataTrack.tryPush` - (this method you have commented on)
  - `OutgoingDataTrackManager.tryProcessAndSend` - main processing of data: https://github.com/livekit/client-sdk-js/blob/52222a114f4997bcf89aa44e8c38e3d535e13e9b/src/room/data-track/outgoing/OutgoingDataTrackManager.ts#L127
    - `DataTrackOutgoingPipeline.processFrame` - converts the `DataTrackFrame` into a series of `DataTrackPacket`s. https://github.com/livekit/client-sdk-js/blob/261981622153e6a1135a516837d7f1daf02974ea/src/room/data-track/outgoing/pipeline.ts#L36
    - `OutgoingDataTrackManager` emits a `packetsAvailable` event each time new packets are generated with these packets
  - The room is listening for a `packetsAvailable` event and when it gets packets it sends them via the (newly made more generic) `RTCEngine.sendLossyBytes` method: https://github.com/livekit/client-sdk-js/blob/261981622153e6a1135a516837d7f1daf02974ea/src/room/Room.ts#L284-L286

---------

Also for completeness, the path data takes coming in via `RemoteDataTrack`:
- A new data track packet is received from the `RTCEngine`, which emits `EngineEvent.DataTrackPacketReceived` - https://github.com/livekit/client-sdk-js/blob/261981622153e6a1135a516837d7f1daf02974ea/src/room/Room.ts#L649-L657
- This calls `IncomingDataTrackManger.packetReceived`: https://github.com/livekit/client-sdk-js/blob/261981622153e6a1135a516837d7f1daf02974ea/src/room/data-track/incoming/IncomingDataTrackManager.ts#L461
  - `IncomingDataTrackPipeline.processPacket` - consumes `DataTrackPacket`s and emits a `DataTrackFrame` once a complete frame has been formed: https://github.com/livekit/client-sdk-js/blob/261981622153e6a1135a516837d7f1daf02974ea/src/room/data-track/incoming/pipeline.ts#L53
  - Gets all `ReadableStreamDefaultController`s mapping to each subscription a user has active for a data track, and enqueues the newly received frame: https://github.com/livekit/client-sdk-js/blob/261981622153e6a1135a516837d7f1daf02974ea/src/room/data-track/incoming/IncomingDataTrackManager.ts#L501
- A user gets the frame pop out of any `remoteDataTrack.subscribe()` (returns `ReadableStream`) calls that are currently active for use in downstream code



(Sidenote - I have generally tried to architect things similarly to the rust implementation with the manger / inner pipeline / etc. I really like this manager pattern and it has made some of these interactions so much easier to test!)

src/room/events.ts
```diff
@@ -352,6 +352,16 @@ export enum RoomEvent {
    * fired when the client receives connection metrics from other participants
    */
   MetricsReceived = 'metricsReceived',
+
+  /**
+   * Emits when a new data track has been published by a downstream participant.
+   */
+  RemoteDataTrackPublished = 'remoteDataTrackPublished',
```
lukasIO:
```suggestion
DataTrackPublished = 'dataTrackPublished',
```

conforms with existing `TrackPublished` and `TrackUnpublished` APIs which are remote if not explicitly prefixed with `Local`

src/room/events.ts
```diff
@@ -352,6 +352,16 @@ export enum RoomEvent {
    * fired when the client receives connection metrics from other participants
    */
   MetricsReceived = 'metricsReceived',
+
```
lukasIO, line 355: For consistency we probably also want events for `LocalDataTrackPublished` and unpublished

src/room/events.ts
lukasIO: usually we would also proxy the events as a `ParticipantEvent` and emit it directly on the participant. 
If we want it to feel a bit closer to how existing APIs work, that might be a good idea here, too

src/utils/waitable-map.ts
lukasIO: `getDeferred` sounds good to me!

src/room/data-track/LocalDataTrack.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  /** The raw descriptor from the manager containing the internal state for this local track. */
+  protected get descriptor() {
+    return this.handle ? this.manager.getDescriptor(this.handle) : null;
+  }
+
+  /** Publish the track to the SFU. This must be done before calling {@link tryPush} for the first time. */
+  async publish(signal?: AbortSignal) {
```
lukasIO, line 69: Still not sure about the dual path of both 
```
lp.publishDataTrack(...) 
 // and
dataTrack.publish()
```

being available. Maybe let's make `publish()` an internal method for now? So still public, but not documented.

src/room/data-track/LocalDataTrack.ts
```diff
@@ -0,0 +1,113 @@
...
+    try {
```
lukasIO, line 70: I guess this should error early (or warn and return) if `isPublished` is true?

src/room/data-track/LocalDataTrack.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  /**
+   * Unpublish the track from the SFU. Once this is called, any further calls to {@link tryPush}
+   * will fail.
+   * */
+  async unpublish() {
+    if (!this.handle) {
+      throw DataTrackPushFrameError.trackUnpublished();
```
lukasIO: suggestion: might be enough to log a warning here and return early?

src/room/data-track/LocalDataTrack.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /** Try pushing a frame to subscribers of the track.
+   *
+   * Pushing a frame can fail for several reasons:
+   *
+   * - The track has been unpublished by the local participant or SFU
+   * - The room is no longer connected
+   */
+  tryPush(payload: DataTrackFrame['payload']) {
```
lukasIO: it's a bit unexpected to me that a `tryPush` method with that explicit `try*` prefix can throw

src/room/events.ts
1egoman: > conforms with existing TrackPublished and TrackUnpublished APIs which are remote if not explicitly prefixed with Local

Ah, good point - updated.

> usually we would also proxy the events as a ParticipantEvent and emit it directly on the participant.
If we want it to feel a bit closer to how existing APIs work, that might be a good idea here, too

Looking at the current participant events, it looks like they are segmented by locality rather than type. ie:
```tsx
  trackPublished: (publication: RemoteTrackPublication) => void;
  // ...
  trackUnpublished: (publication: RemoteTrackPublication) => void;
  // ...
  localTrackPublished: (publication: LocalTrackPublication) => void;
  localTrackUnpublished: (publication: LocalTrackPublication) => void;
```

Because of this, I'm sort of of the opinion to hold off on proxying these events to the participant until the larger track changes get rolled out across everything, and then at that point we can get these all going through the same interfaces rather than tacking on some additional data track specific events to this list. Any concerns with that approach?

src/room/events.ts
1egoman: Added these!

src/room/data-track/LocalDataTrack.ts
1egoman: Sure, I will keep `publish` internal for now.

src/room/data-track/LocalDataTrack.ts
1egoman: There is already an error that gets thrown within `publishRequest` if the track is already published, but I could probably make the error message a little nicer (right now it is one of those "panic" bare errors).

IMO in this case, I think an explicit throw makes sense because you really should not be publishing / unpublishing data tracks without some care. But looking at the existing tracks it looks like it just warns and that's it. So maybe it makes sense to convert this throw into a warning, not sure.

EDIT: more thoughts on this [here](https://github.com/livekit/client-sdk-js/pull/1820#discussion_r2942172721)!

src/room/data-track/LocalDataTrack.ts
1egoman: Yea I am not sure - see my previous comment about the throw vs warning [here](https://github.com/livekit/client-sdk-js/pull/1820#discussion_r2942142268). I like the hard error in that it is something that can be caught and explicitly handled though since this is checking an invariant. The existing track publication logic for audio / video tracks logs a warning so maybe it makes sense to update it to be consistent. 

I guess one downside to the hard error is it opens this code up to a [TOC-TOU issue](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) where something like the below could happen:
```tsx
if (track.isPublished) {
  // Imagine here the track is unpublished at the SFU level for some unrelated reason ...
  await track.unpublish(); // ... so this would then throw an error
}
```

I am definitely curious what others think here!

src/room/data-track/LocalDataTrack.ts
1egoman: Yea good point - this one I think should definitely be a warning. Updated!

src/room/data-track/LocalDataTrack.ts
pblazej: Thanks for a detailed explanation, I just meant why there's no `Promise<void>` and the `try/catch` thing you commented on.

1egoman: Of course! Typescript doesn't require a `Promise<void>`, it is able to infer that automatically.

For reference, the below `foo`, `bar`, and `baz` function definitions are identical:
```tsx
function returnsAPromise(): Promise<number> {
  // ...
}

function foo() {
  return returnsAPromise();
}

async function bar() {
  return returnsAPromise();
}

async function baz() {
  const result = await returnsAPromise();
  return result;
}
```

src/room/data-track/RemoteDataTrack.ts
```diff
@@ -64,9 +65,7 @@ export default class RemoteDataTrack implements IRemoteTrack, IDataTrack {
    * Note that newly created subscriptions only receive frames published after
    * the initial subscription is established.
    */
-  async subscribe(
-    options?: RemoteDataTrackSubscribeOptions,
-  ): Promise<ReadableStream<DataTrackFrame>> {
+  async subscribe(options?: RemoteDataTrackSubscribeOptions): Promise<DataTrackSubscriptionReader> {
```
1egoman: I was doing a final pass through the data tracks js interface and noticed a bit of a weird edge case when it comes to cancellation of subscriptions - ie, before, you could do this:
```tsx
const readableStream = await dataTrack.subscribe({ signal /* <- abort signal here */ });
for await (const item of readableStream) {
  // You could eventually call the below second unrelated cancellation mechanism
  // (in theory, though probably rare):
  readableStream.cancel(...);
} // (The stream is released here automatically)
```

They both sort of do different things... the signal lets you cancel the in flight subscription as it attempts to subscribe, and the .cancel() lets you cancel the subscription once it is subscribed and are done with it. Both need to exist but exposing both to a user could be quite confusing.

So I went back and forth with a LLM a bit on this and it recommended re-introducing the wrapper around the readable stream to normalize the cancellation pattern to just use the signal . Where it ended up was something like this:
```tsx
const controller = new AbortController();
const asynciterable = await dataTrack.subscribe({ signal: controller.signal });
// Note asynciterable is a custom object, not a ReadableStream (though it contains one internally)
for await (const item of asynciterable) {
  // Consume items here

  // Or, you could call `controller.abort()` if you really wanted to manually abort in
  // that edge case from before
} // (The stream is released here automatically)
```

I'm generally a fan, and there also is a pretty clean way where the data streams and data tracks versions of this can share a common interface at the bottom level if we wanted.

If anybody has any thoughts, I'd be interested in hearing them!

1egoman: Talked to Lukas a bit, he wasn't a fan, and proposed an alternative:
```tsx
const onDataTrackPublished = (track) => {
  // nothing happens on the SFU side wrt to subscriptions by default

  // when a user calls
  const stream = track.stream(); // this indicates that a subscription should happen (or an existing one being reused)

  for await (const chunk of stream) {
    // while the subscription is being wired up with the SFU this loop will simply be pending as no data has been fed to it yet
  }
 
}
```

I will revert this for now and we can discuss this a bit more before committing to a larger api change.

src/room/participant/LocalParticipant.ts
```diff
@@ -311,6 +320,38 @@ export default class LocalParticipant extends Participant {
...
+            error = DataTrackPublishError.duplicateName(response.message);
+            break;
+          case RequestResponse_Reason.INVALID_NAME:
+            error = DataTrackPublishError.invalidName(response.message);
+            break;
+          case RequestResponse_Reason.LIMIT_EXCEEDED:
+            error = DataTrackPublishError.limitReached(response.message);
+            break;
+          default:
+            this.log.error(
```
ladvoc: question: If later new error reasons are added for data track operations, would we want to log an error on older client or introduce an unknown case?

src/room/participant/LocalParticipant.ts
```diff
@@ -2235,4 +2277,16 @@ export default class LocalParticipant extends Participant {
       await sleep(20);
     }
   }
+
+  /** Publishes a data track.
+   *
+   * Returns the published data track if successful. Use {@link LocalDataTrack#tryPush}
```
ladvoc, line 2280: praise: Thank you for bringing over doc comments!

src/room/events.ts
```diff
@@ -352,6 +352,26 @@ export enum RoomEvent {
...
+  /**
+   * Emits when a new data track has been published by a downstream participant.
+   */
+  DataTrackPublished = 'dataTrackPublished',
```
ladvoc, line 359: comment: In Rust, this event is called `RemoteDataTrackPublished`.

src/room/Room.ts
```diff
@@ -243,6 +252,46 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     this.incomingDataStreamManager = new IncomingDataStreamManager();
     this.outgoingDataStreamManager = new OutgoingDataStreamManager(this.engine, this.log);
 
+    this.incomingDataTrackManager = new IncomingDataTrackManager({ e2eeManager: this.e2eeManager });
+    this.incomingDataTrackManager
+      .on('sfuUpdateSubscription', (event) => {
+        this.engine.client.sendUpdateDataSubscription(event.sid, event.subscribe);
+      })
+      .on('trackAvailable', (event) => {
```
ladvoc: question: Would it make sense to rename the manager level events for consistency with the room level ones (e.g., published instead of available)? Will do this in Rust too if you end up changing it.

src/room/data-track/RemoteDataTrack.ts
```diff
@@ -64,11 +64,9 @@ export default class RemoteDataTrack implements IRemoteTrack, IDataTrack {
    * Note that newly created subscriptions only receive frames published after
    * the initial subscription is established.
    */
-  async subscribe(
-    options?: RemoteDataTrackSubscribeOptions,
```
ladvoc, line 68: nitpick: This could probably just be called `DataTrackSubscribeOptions` since subscribing is only applicable to remote tracks.

src/room/events.ts
1egoman: This was a change lukas requested, more info here: https://github.com/livekit/client-sdk-js/pull/1820#discussion_r2941554090

src/room/data-track/RemoteDataTrack.ts
1egoman: Good idea, updated!

src/room/participant/LocalParticipant.ts
1egoman: Good point, I hadn't considered this. It looks like the rust implementation [is raising an error](https://github.com/livekit/rust-sdks/blob/00d1d2b2dfecd6a4090a2648801d52db81fc73f5/livekit-datatrack/src/local/proto.rs#L89) in this case currently - a `PublishError::Internal`.

I will do something similar here (2d58fee7) but in regards to the name: What do you think about `::Unknown` for this case in both js / rust, or maybe something else that is specific for this exact case? Internal is being used as a catch all in a few other places so it isn't unique.

src/room/Room.ts
1egoman: Good point, renamed here: e5dc9ae6

src/room/data-track/LocalDataTrack.ts
ladvoc: You make a good point; a data track unpublish can be SFU initiated (e.g., if publish data permission is revoked). Given that, I think it makes sense to treat this as a first-class error.

lukasIO: hm, from @1egoman 's description I'd be more inclined to treat it similarly to how other unpublish requests are handled, which is warning only

src/room/RTCEngine.ts
```diff
@@ -729,29 +769,40 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
     });
+    this.dataTrackDC = this.pcManager.createPublisherDataChannel(dataTrackDataChannel, {
+      ordered: false,
+      maxRetransmits: 0,
+    });
 
     // also handle messages over the pub channel, for backwards compatibility
     this.lossyDC.onmessage = this.handleDataMessage;
     this.reliableDC.onmessage = this.handleDataMessage;
+    this.dataTrackDC.onmessage = this.handleDataTrackMessage;
```
lukasIO, line 780: given that we need a newer SFU version anyways for data track support, is this still needed?

src/room/RTCEngine.ts
```diff
@@ -855,11 +927,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
   };
 
-  private handleBufferedAmountLow = (event: Event) => {
-    const channel = event.currentTarget as RTCDataChannel;
-    const channelKind =
-      channel.maxRetransmits === 0 ? DataPacket_Kind.LOSSY : DataPacket_Kind.RELIABLE;
-
+  private handleBufferedAmountLow = (_event: Event, channelKind: DataChannelKind) => {
```
lukasIO: we don't really need to pass the event here anymore, right? `channelKind` seems sufficient

src/room/RTCEngine.ts
```diff
@@ -1326,57 +1394,93 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-          if (this.lossyDataDropCount % 100 === 0) {
-            this.log.warn(
-              `dropping lossy data channel messages, total dropped: ${this.lossyDataDropCount}`,
-              this.logContext,
-            );
-          }
-          return;
+      if (!this.isBufferStatusLow(kind)) {
+        // Depending on the exact circumstance that data is being sent, either drop or wait for the
+        // buffer status to not be low before continuing.
```
lukasIO, line 1439: what's the intention behind this differentiation? Why not just drop them for lossy?

src/room/RTCEngine.ts
1egoman: Yes - talking with raja, both this and the `dataTrackDCSub` path ([here](https://github.com/livekit/client-sdk-js/blob/ff315e72aa04da950ed2adc9f4b781872dce986a/src/room/RTCEngine.ts#L829-L830)) are required to handle data tracks in single and dual peer connection modes.

src/room/RTCEngine.ts
1egoman: Oops, I forgot to remove this. Thanks!

src/room/RTCEngine.ts
1egoman: That's what I originally was doing but this didn't work for large data track payloads - here is some more context as to why from some notes I wrote up:

> - If the `_data_tracks` rtc data channel buffer was full, the existing logic would drop packets (what the lossy data channel currently does).
>   - This doesn't work though for data track packets, because if the first `n` packets of a frame get enqueued but the last packet(s) get dropped because there isn't enough room in the rtc data channel buffer, the receiver only receives a partial set of packets which will never fully decode into a data track frame.
>   - If packets aren't buffered like this, then it becomes impossible to send data track payloads > ~75k bytes (the e2e tests are checking this case)
>   - Solution ([commit](https://github.com/livekit/client-sdk-js/commit/c461a0cc)): buffer data track packets instead of dropping them if the rtc data channel buffer is full

If there's some other way to handle this, I'm open to alternatives but what I did here was the most straightforward way I could think of.

src/room/data-track/LocalDataTrack.ts
ladvoc: From sync: giving some more thought to this, I would say _publish_ should throw whereas _unpublish_ should not. For publish, this is something the application might want to handle and respond to.

1egoman: ^ Made these updates here: a3714aaf

src/room/data-track/LocalDataTrack.ts
1egoman: Per [the other thread](https://github.com/livekit/client-sdk-js/pull/1820#discussion_r2941645482), closing this.

PR #1827, opened 2026-02-27
Reuse inactive publisher transceivers in RTCEngine to reduce memory growth on republish

moufmouf:
## Why?

I'm developing WorkAdventure, a virtual space platform and we use Livekit for big meetings.

The issue that triggered this PR is that when we start publishing a screen sharing, the memory goes up by about 400MB. We stop publishing the screen sharing with `localParticipant.unpublishTrack`, the memory does not go down. We start publishing another screen sharing again (maybe another window), the memory goes up by another 400MB, etc...
In short, we have a memory leak, where each started screen sharing is never releasing memory.

As a side note, using `track.pauseUpstream` to stop the track then replace the track with another track later seems to work better, but this workflow causes other issues (like the fact that egress CompositeRoom will display a black box for a paused video instead of not showing it at all)

## How to reproduce the issue

You can see the issue in https://meet.livekit.io/

- Simply start a meeting with another user. 
- Open the Google Chrome task manager and note the memory consumption of your current tab
- Start a screen share => the memory goes up
- Stop  the screen share => the memory does not go down
- Start another screen share => the memory keeps going up
- etc...

After 3/4 screenshares, your tab takes more than 1GB RAM.

## Finding the memory leak

I tried to track down where the memory leak was, and after a few trial and errors (and with the help of Codex), I ended up finding that the transceivers in the `RTCEngine` seem to be piling up each time we start a new track. The `RTCRtpTransceiver` seem to hold a reference to the `MediaTrack` that is never freed and eating a lot of memory. I did not find any way to properly free the `MediaTrack` reference, but Codex (again) proposed me the patch below. Basically, it tries to reuse an inactive existing sender that shares the same track "kind".

I tested this with WorkAdventure and the patch works. When I stop / start many screen sharing many times, the memory does not pile up. I also re-read the patch and validated it myself (to avoid any AI slop).

Note: I tried an alternative: freeing the memory as soon as `localParticipant.unpublishTrack` is called (using `replaceTrack(null)`) instead of waiting for the next track to start to reuse the transceiver. Alas, I could not make this work (that would be ideal as the memory would be freed as soon as we finish publishing... here, we have to wait getting out of the Livekit room for the memory to be freed)

What do you think of this patch? I can happily work on it if it needs any improvement.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -910,6 +910,11 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       track.codec = opts.videoCodec;
     }
 
+    const reusedSender = await this.reuseInactivePublisherSender(track.mediaStreamTrack, streams);
+    if (reusedSender) {
+      return reusedSender;
+    }
+
     const transceiverInit: RTCRtpTransceiverInit = { direction: 'sendonly', streams };
     if (encodings) {
       transceiverInit.sendEncodings = encodings;
@@ -932,6 +937,15 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     if (!this.pcManager) {
       throw new UnexpectedConnectionState('publisher is closed');
     }
+    const reusedSender = await this.reuseInactivePublisherSender(simulcastTrack.mediaStreamTrack);
+    if (reusedSender) {
+      if (!opts.videoCodec) {
+        return;
+      }
+      track.setSimulcastTrackSender(opts.videoCodec, reusedSender);
+      return reusedSender;
+    }
+
     const transceiverInit: RTCRtpTransceiverInit = { direction: 'sendonly' };
     if (encodings) {
       transceiverInit.sendEncodings = encodings;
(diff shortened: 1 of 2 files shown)


moufmouf:
Ping. Kindly looking for feedback as this removes a big memory leak in the browser.

src/room/RTCEngine.ts
```diff
@@ -910,6 +910,11 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
       track.codec = opts.videoCodec;
     }
 
+    const reusedSender = await this.reuseInactivePublisherSender(track.mediaStreamTrack, streams);
```
lukasIO, line 913: this doesn't appear to apply the encodings of the new track

lukasIO (commented):
thanks for the PR and sorry for the delayed review! 

We're reusing transceivers also on the subscriber side and with single peer connection mode (the default) those would also show up on the publisher peer connection. 

In general fully on board with addressing any memory leaks. However reusing the transceiver as proposed here needs some careful consideration as - additionally to ensure only publisher side transceivers are reused as explained above - there might also still be stale frames from the previous publication in the pipeline if we're not flushing actively. 

I also wasn't able to actually get this PR to work with the `demo.ts` sample (`pnpm run dev` in this repo). The second screen share never shows up.

moufmouf:
Thanks for the feedback @lukasIO :pray: 
I'm away this week. I'll level up on the "reusing transceivers" subject, and I'll come back next week with improvements :+1:

PR #1832, opened 2026-03-11
Support for large RPC messages using data streams

1egoman:
Currently this prototype does the below:
- Use the existing `RPCRequest` path when the advertised `clientProtocol` is less than 1.
- Otherwise, compress and send rpc messages via data streams rather than send actual `RPCRequest` / `RPCResponse` messages.

Also undertakes some signifigant refactoring to move RPC logic out into a `RpcClientManager` / `RpcServerManager`, which given the increase in complexity makes the code a little less fragmented / easier to reason about.

### Todo
- [x] Add changeset
- [x] Tests have not been updated, so do that
- [x] Add tests for the new non "legacy" paths - compression and data streams
- [x] I currently have the benchmark I ran checked in under `examples/rpc-benchmark` - this is what is leading to the large diff size. ~Before potentially merging this, think about if it makes sense to keep this checked in or not. And if so, go through and update some of the docs to be a little more modern / correct (in particular, it still refers to the "legacy path < 1kb" stuff).~ I decided to check it out of version control for now given the goal of this has shifted to worry less about performance in the short term.


src/room/rpc/client/RpcClientManager.ts
```diff
@@ -0,0 +1,353 @@
...
+    this.log = log;
+    this.outgoingDataStreamManager = outgoingDataStreamManager;
+    this.getRemoteParticipantClientProtocol = getRemoteParticipantClientProtocol;
+  }
+
+  setupEngine(engine: RTCEngine) {
+    this.engine = engine;
+
+    this.engine.on(EngineEvent.DataPacketReceived, this.handleDataPacket);
+  }
```
1egoman, line 57: Note to self: in attempting to update the tests for this, I realized that tightly coupling both `RpcClientManager` / `RpcServerManager` to `RTCEngine` is a bad idea. It should work more like the data tracks managers work where there are incoming and outgoing events, and the glue to wire those events up to the engine happens at the room level.

src/room/rpc/utils.ts
```diff
@@ -139,10 +139,156 @@ export class RpcError extends Error {
...
+ * Topic used for RPC payload data streams.
+ * @internal
+ */
+export const RPC_DATA_STREAM_TOPIC = 'lk.rpc_payload';
+
+/**
+ * Compress a string payload using gzip.
+ * @internal
+ */
+export async function gzipCompress(data: string): Promise<Uint8Array> {
```
lukasIO: for both compress helpers we need to handle unsupported browsers: https://caniuse.com/mdn-api_compressionstream

for regular payloads this should probably fall back to `'regular'` in these cases

1egoman: Definitely worth thinking about, yea. Just generally, the support for compression stream looks fairly good - are you saying there needs to be a fully developed "fallback path" here, or more just that if it doesn't exist it should fail with a nicer message than `CompressionStream is not defined` but still not work?

If it is the former, even if this were to fall back to sending old rpc requests, there is still the rpc response end to contend with. If you are at or above client protocol level 1 right now then that means you should be able to handle recieiving compressed rpc responses. 

A few options come to mind:
- Get rid of the compression outright for now, and address this as a future data stream level improvement.
- Rethink this whole client protocol thing to make it so features aren't additive like they are modeled currently.

src/room/rpc/client/RpcClientManager.ts
1egoman: I generally fixed this, but it exposed more questions. I will surface them in another comment!

src/room/rpc/utils.ts
1egoman: I talked with lukas, and it sounds like the answers to my questions are a) `CompressionStream` is too new to be expected to exist everywhere (but that was more of a "vibe", there aren't concrete browser requirements to point to here), and b) it probably would be better to make these compression changes to the data stream directly instead.

So I have (sadly, because it had a _big_ performance impact!) pulled the compression stuff out of this and will investigate re-adding it as part of some future scheduled data streams improvements.

src/room/rpc/client/RpcClientManager.test.ts
```diff
@@ -0,0 +1,297 @@
...
+      mockOutgoingDataStreamManager = {
+        streamText: vi.fn().mockResolvedValue(mockStreamTextWriter),
+      } as unknown as OutgoingDataStreamManager;
+
+      rpcClientManager = new RpcClientManager(
+        log,
+        mockOutgoingDataStreamManager,
+        (_identity) => CLIENT_PROTOCOL_DATA_STREAM_RPC, // (other participant is "v2")
+        () => undefined,
+      );
```
1egoman, line 155: question: I still am not too pleased with how tightly coupled all these manager pieces are here. I think getting `RtcEngine` out of `OutgoingDataStreamManager` is probably fairly easy and worthwhile (though I think I'd prefer to do that in a follow up change) but even if so, `RpcClientManager` and `OutgoingDataStreamManager` still are quite coupled and it would be awkward to decouple them...

I thought that a pattern closer to how data tracks works would be worthwhile and while that was definitely helpful, it really only works for "fire and forget" type events like sending a data packet. It doesn't work all that well for request response type interactions (ie, calling an async function where you want to block until the function is complete, not just fire and forget, like calling `outgoingDataTrackManager.streamText`).

I think what I've got here works but I think it can get better. If others have ideas on how to do that, I'd be interested in hearing them!

RPC_SPEC.md
```diff
@@ -0,0 +1,349 @@
+# RPC v2 Specification
+
+## Overview
+
```
1egoman: Note to other reviewers: this spec was partially LLM generated and the thought is will act as a plan for LLMs to implement these same features in other sdks.

This will NOT be merged with this change, but I wanted a place for it to live that was not notion / etc since it's fairly tied to this code and should evolve along with it if review comments result in protocol changes.

xianshijing-lk: I think this is a good idea. 

I do wonder if we should setup a repo to save those spec.mds, then we will use submodule to share it across sdks

src/room/rpc/server/RpcServerManager.ts
```diff
@@ -0,0 +1,288 @@
...
+        requestId,
+        null,
+        RpcError.builtIn('APPLICATION_ERROR'),
+      );
+      return;
+    }
+
+    this.publishRpcAck(callerIdentity, requestId);
+
+    if (version !== 2) {
```
lukasIO: question: should we define the `2` as a named constant somewhere?

src/room/rpc/server/RpcServerManager.ts
```diff
@@ -0,0 +1,288 @@
...
+    payload: string,
+  ) {
+    const callerClientProtocol = this.getRemoteParticipantClientProtocol(destinationIdentity);
+
+    if (callerClientProtocol >= CLIENT_PROTOCOL_DATA_STREAM_RPC) {
+      // Send response as a data stream
+      const writer = await this.outgoingDataStreamManager.streamText({
+        topic: RPC_RESPONSE_DATA_STREAM_TOPIC,
+        destinationIdentities: [destinationIdentity],
+        attributes: { [RPC_REQUEST_ID_ATTR]: requestId },
```
lukasIO: issue: looks like this is missing the version field that you're checking against above?

1egoman: @lukasIO The way it currently works, the version field is only being included on the "request" end's data stream, NOT the "response" end's data stream (what you are leaving this comment on), which matches with how rpc works today.

I went back and forth on this a bit but ended up coming to the conclusion that if a rpc client in the future wanted to switch based on the version of a response, they could just store a mapping of `request id` => `version number` and look up the response's request id in the mapping, so I opted to keep this as it is currently. The `version` of the response will always be 1:1 with the request's `version` so if a client sends a request of a certain `version`, it can be assumed that it will know how to process a response of that same `version`.

Maybe there's a case you can think of though where including this version on both ends would be important? I'm fine with adding it in the response end here as well if so.

src/room/rpc/server/RpcServerManager.ts
1egoman: I think this is a good idea, yea!

src/room/rpc/server/RpcServerManager.ts
lukasIO: yeah, that makes sense as is. Agree with your conclusion!

1egoman:
(rebased on top of latest `main`)

src/room/rpc/server/RpcServerManager.ts
1egoman: Done in f69d8a05

RPC_SPEC.md
1egoman: @xianshijing-lk @lukasIO @reenboog @ladvoc @MaxHeimbrock @pblazej

I am tagging a larger group of people in the hopes of getting this RPC v2 specification document reviewed. The web implementation is largely done at this point. Feel free to leave comments that would be relevant to the respective SDKs you work on, and then this document can serve as a prompt for a LLM to port this over to each respective SDK.

RPC_SPEC.md
```diff
@@ -0,0 +1,349 @@
...
+`clientProtocol` is a new integer that each participant advertises in their `ParticipantInfo` via
+the LiveKit signaling channel. It tells other participants what client-to-client features this SDK
+supports. It is distinct from the existing `protocol` field (which tracks signaling protocol
+version) - `clientProtocol` specifically governs peer-to-peer feature negotiation.
+
+| Value | Constant name | Meaning |
+|-------|---------------|---------|
+| `0` | `CLIENT_PROTOCOL_DEFAULT` | Legacy client. Only supports RPC v1. |
+| `1` | `CLIENT_PROTOCOL_DATA_STREAM_RPC` | Supports RPC v2 (data stream-based payloads). |
```
ladvoc: question: Would it make sense for this to be a bitfield rather than a single version number? I can maybe see a scenario where we have another new feature but not all clients immediately implement it.

RPC_SPEC.md
```diff
@@ -0,0 +1,349 @@
...
+  |    topic: "lk.rpc_response"        |
+  |    attrs: request_id               |
+  |    body: <response payload>        |
+  |                                    |
+```
+
+1. The **caller** opens a text data stream with:
+   - **Topic**: `lk.rpc_request`
+   - **Destination identities**: `[destinationIdentity]`
+   - **Attributes**:
```
ladvoc: question: Maybe these should be made into Protobuf fields? Currently, every data stream header has a "content header" which adds specific fields based on whether it is a text or byte stream. We could define a similar content header for RPC as well.

Pros:
- Type safety
- Easier to support ESP32 SDK. Deserializing hash maps on this platform is a potential performance issue.

Cons:
- Potential protocol bloat

RPC_SPEC.md
```diff
@@ -0,0 +1,349 @@
...
+  |                                    |  (reads stream, looks up handler, invokes it)
+  |<-- RpcAck (DataPacket) -----------|
+  |<-- Text data stream (response) ---|
+  |    topic: "lk.rpc_response"        |
+  |    attrs: request_id               |
+  |    body: <response payload>        |
+  |                                    |
+```
+
+1. The **caller** opens a text data stream with:
```
ladvoc: question: I am assuming for RPC we will only support finite streams?

RPC_SPEC.md
lukasIO: I've brought this up as well with @1egoman. Ryan is moderately strongly opposed to that. 
I think I can get on board with it being an incrementing value only because that's also how we handle it for the SFU protocol version and the support matrix will become a pain to deal with if we do the bitfield variant.

RPC_SPEC.md
lukasIO: fair point, I think a way to have a schema/generated types for attributes would be nice in general (we were experimenting with quicktype for that last year). 

Adding RPC specific fields to the stream envelope doesn't sound ideal

RPC_SPEC.md
lukasIO: yeah, in order not to change the user facing API surface for RPCs we will not allow users to stream things in directly. This also means that (at least for now) on the receiving end users won't have the ability to stream the response as it comes in. 

We could think about an overload for the handler that would allow for that though. Probably a good time to consider this a bit more.

lukasIO: Having thought about it a bit more: 
Given that we're dealing with strings as payload always (😞 ) the only advantage I see for streaming responses would be for "raw" strings. With us generally being in the dark about how users would serialize their payload to strings and @1egoman's work on auto serialization to JSON per default it becomes a lot less useful to read responses as a stream, I believe

1egoman: I generally agree with @lukasIO's perspective here, but also would like to add that I don't think there's anything at a protocol level that prevents this being introduced later if the client apis were to change. That being said, I also don't really see a large use case for this currently and would prefer to hold off on making client api changes until there is demand.

RPC_SPEC.md
1egoman: > Ryan is moderately strongly opposed to that.

Indeed I am, mostly because having to test every possible permutation of each feature across all clients in a room is going to be a large pain and multiply bug surface area.

I think there really are two different things that a client needs to be able to advertise:
1. "I support a feature" - ie, maybe something like supporting a certain compression algorithm, supporting a certain type of hardware acceleration, etc. These are things that are truly optional and which some platforms would potentially never be able to fulfill (ie, esp32). This list should be kept as small as possible.
2. "I support certain protocol level messages" - these RPC updates are a great example of that. These don't depend on platform specific features / abilities and can be updated independently of what types of the first item a certain platform may support.

Some things are definitely ambiguous, and it could be argued that a "protocol level message" is a subtype of a "feature" (but that makes the test matrix explode as lukas said). For now I think we can get away with just item 2, which is represented by the client protocol version. But I think for other upcoming tasks like adding compression to data streams, I think it's inevitable that we will need to have both.

### Example
Lets say there is a`supportsCompression` feature.

- In client protocol number 2 you introduce the ability to send compressed data streams
- Client protocol number 3 introduced protocol changes for `foo`
- Client protocol number 4 introduced protocol changes for `bar`
- The current client protocol version is 5 which introduced protocol changes for `baz`.

A web browser on client protocol 5 can advertise the `supportsCompression` feature conditionally based on if `window.CompressionStream` is defined, which allows other participants to know it can receive compressed data streams or not. If you didn't do this, then the web browser would have to advertise an old client protocol version (something before `2`, so maybe `1`) to indicate "I don't support compression" but that means that all the new changes from 3-5 must also be disabled. 

If you instead modeled each feature as a bit in a bitfield, then you now have to test all the combinations of all features always - you can't assume that `foo` would always exist on client protocol `4` alongside `bar` like the incrementing client protocol approach would give you, which makes the test matrix much bigger (exponentially so as new features would be introduced).

RPC_SPEC.md
1egoman: I'm ambivalent on this, curious what others may think here and if this would make things significantly easier on other platforms!

src/room/rpc/client/RpcClientManager.ts
```diff
@@ -0,0 +1,259 @@
...
+    responseTimeout: responseTimeoutMs = 15000,
+  }: PerformRpcParams): Promise<[id: string, completionPromise: Promise<string>]> {
+    const maxRoundTripLatencyMs = 7000;
+    const minEffectiveTimeoutMs = maxRoundTripLatencyMs + 1000;
+
+    const remoteClientProtocol = this.getRemoteParticipantClientProtocol(destinationIdentity);
+    const payloadBytes = byteLength(payload);
+
+    // Only enforce the legacy size limit when on rpc v1
+    if (payloadBytes > MAX_V1_PAYLOAD_BYTES && remoteClientProtocol < 1) {
```
xianshijing-lk: Nit, 

should it be < CLIENT_PROTOCOL_DATA_STREAM_RPC ?

src/room/rpc/client/RpcClientManager.ts
```diff
@@ -0,0 +1,259 @@
...
+    const serverVersion = this.getServerVersion();
+    if (serverVersion && compareVersions(serverVersion, '1.8.0') < 0) {
+      throw RpcError.builtIn('UNSUPPORTED_SERVER');
+    }
+
+    const effectiveTimeoutMs = Math.max(responseTimeoutMs, minEffectiveTimeoutMs);
+    const id = crypto.randomUUID();
+
+    await this.publishRpcRequest(
+      destinationIdentity,
```
xianshijing-lk: I think there can be a race condition here:
we publish the request here, then call this.pendingAcks.set in line 105.
A race condition can happen if the response comes before the code gets executed to line 105. 

Even though it rarely happens. 

I will suggest changing the order:
setup the pendingAck and timeout, then call publishRpcRequest

src/room/rpc/client/RpcClientManager.ts
```diff
@@ -0,0 +1,259 @@
...
+  /**
+   * Handle an incoming data stream containing an RPC response payload.
+   * @internal
+   */
+  async handleIncomingDataStream(reader: TextStreamReader, attributes: Record<string, string>) {
+    const associatedRequestId = attributes[RPC_REQUEST_ID_ATTR];
+    if (!associatedRequestId) {
+      this.log.warn(`RPC data stream malformed: ${RPC_REQUEST_ID_ATTR} not set.`);
+      this.handleIncomingRpcResponseFailure(
+        associatedRequestId,
```
xianshijing-lk: should you still call handleIncomingRpcResponseFailure if associatedRequestId is undefined ?

src/room/rpc/server/RpcServerManager.ts
```diff
@@ -0,0 +1,289 @@
...
+  unregisterRpcMethod(method: string) {
+    this.rpcHandlers.delete(method);
+  }
+
+  /**
+   * Handle an incoming RPCRequest message containing a payload.
+   * This handles "version 1" of rpc requests.
+   * @internal
+   */
+  async handleIncomingRpcRequest(callerIdentity: string, rpcRequest: RpcRequest) {
```
xianshijing-lk, line 63: I think this has been the code like that, both handleIncomingRpcRequest and handleIncomingDataStream are async and trigger the handler. If a participant  or bad actors floods the client with RPC requests, will this class spin up a large number of concurrent handler() executions ? Is it a concern ? If so, anything we can do to improve it ?

src/room/rpc/server/RpcServerManager.ts
```diff
@@ -0,0 +1,289 @@
...
+      this.publishRpcResponsePacket(
+        callerIdentity,
+        requestId,
+        null,
+        RpcError.builtIn('APPLICATION_ERROR'),
+      );
+      return;
+    }
+
+    this.publishRpcAck(callerIdentity, requestId);
```
xianshijing-lk, line 150: the same race condition, we probably want to setup the reader before calling publishRpcAck ?

src/room/rpc/client/RpcClientManager.ts
```diff
@@ -0,0 +1,259 @@
...
+    destinationIdentity: string,
+    requestId: string,
+    method: string,
+    payload: string,
+    responseTimeout: number,
+    remoteClientProtocol: number,
+  ) {
+    if (remoteClientProtocol >= CLIENT_PROTOCOL_DATA_STREAM_RPC) {
+      // Send payload as a data stream - a "version 2" rpc request.
+      const writer = await this.outgoingDataStreamManager.streamText({
```
lukasIO, line 145: thought: I think this is fine, just mentioning in case there are some caveats I'm missing:
we know already that we'll want to support compression in the next iteration of this, so wondering if there are any considerations that might make it easier to use `streamBytes` here right now. 
I believe as long as we think about compression to be an option on `streamText` directly it should be fine to keep using `streamText` here.

src/room/rpc/client/RpcClientManager.ts
```diff
@@ -0,0 +1,259 @@
...
+        },
+      });
+
+      await writer.write(payload);
+      await writer.close();
+      return;
+    }
+
+    // Fallback to sending a literal RpcRequest - a "version 1" rpc request.
+    this.emit('sendDataPacket', {
```
lukasIO, line 162: for cases where sending the packet fails (e.g. due to the engine being closed), is it correct to assume that users would only notice the legacy path failing due to the RPC timeout?

src/room/rpc/client/RpcClientManager.ts
```diff
@@ -0,0 +1,259 @@
...
+    } else {
+      this.log.error(`Ack received for unexpected RPC request: ${requestId}`);
+    }
+  }
+
+  /** @internal */
+  handleParticipantDisconnected(participantIdentity: string) {
+    for (const [id, { participantIdentity: pendingIdentity }] of this.pendingAcks) {
+      if (pendingIdentity === participantIdentity) {
+        this.pendingAcks.delete(id);
```
lukasIO, line 257: question: why are we not rejecting the ack here before deleting it?

src/room/rpc/utils.ts
```diff
@@ -139,10 +136,47 @@ export class RpcError extends Error {
...
+export const RPC_REQUEST_ID_ATTR = 'lk.rpc_request_id';
+
+/** @internal */
+export const RPC_REQUEST_METHOD_ATTR = 'lk.rpc_request_method';
+
+/** @internal */
+export const RPC_REQUEST_RESPONSE_TIMEOUT_MS_ATTR = 'lk.rpc_request_response_timeout_ms';
+
+/** @internal */
+export const RPC_REQUEST_VERSION_ATTR = 'lk.rpc_request_version';
```
lukasIO: suggestion: might be a bit nicer to deal with if all the attributes were scoped as a string enum?

src/room/rpc/server/RpcServerManager.ts
```diff
@@ -0,0 +1,289 @@
...
+    } catch (error) {
+      let responseError;
+      if (error instanceof RpcError) {
+        responseError = error;
+      } else {
+        this.log.warn(
+          `Uncaught error returned by RPC handler for ${rpcRequest.method}. Returning APPLICATION_ERROR instead.`,
+          error,
+        );
+        responseError = RpcError.builtIn('APPLICATION_ERROR');
```
lukasIO: suggestion: might be useful to include the error message (or stringified error) in the error meta data?

src/room/rpc/server/RpcServerManager.ts
```diff
@@ -0,0 +1,289 @@
...
+      });
+      await writer.write(payload);
+      await writer.close();
+      return;
+    }
+
+    // Legacy client: enforce size limit and send uncompressed payload inline
+    const responseBytes = byteLength(payload);
+    if (responseBytes > MAX_V1_PAYLOAD_BYTES) {
+      this.log.warn(
```
lukasIO, line 279: thinking about it, I think it would be nice if `registerRpcHandler` would take an optional `onError` callback that would be invoked with these kind of errors. Otherwise the rpc server side doesn't have any means to signal failure to users

src/room/rpc/client/RpcClientManager.ts
1egoman: Yep, we're on the same page - compression would be a thing for both types of data streams so I don't think it should matter.

src/room/rpc/server/RpcServerManager.ts
1egoman: I don't think there's that much that can be done to easily combat against this without introducing new rpc features like explicit rate limits. The existing rpc implementation also suffers from the same problem (though the attacker would be sending `RpcRequest`s instead) and I don't think this is making it any worse.

Since I don't think introducing a rate limit would require a protocol update, I would rather decouple this and think about it as a follow up task.

src/room/rpc/client/RpcClientManager.ts
1egoman: Oops, thanks!

src/room/rpc/client/RpcClientManager.ts
1egoman: Ah yea, good point - I suppose in this case logging is about all that can be done since there's no id to correlate the response with.

src/room/rpc/client/RpcClientManager.ts
1egoman: Yes, this is correct for v1, and this behavior has not changed for v2. It is worth noting though that if the client were to receive an ACK before the packet sending starts failing then the error would be different (`CONNECTION_TIMEOUT` vs `RESPONSE_TIMEOUT`).

src/room/rpc/client/RpcClientManager.ts
1egoman: This is because a few lines further down, the request's `completionFuture` is being rejected instead, which effectively supersedes the `pendingAcks` state and returns a `RpcError` with state `RECIPIENT_DISCONNECTED` from the `performRpc` method. I could convert the `pendingAcks` item to a future and explicitly reject is as well, but I'm not sure that would really change any behavior in practice.

If the described behavior doesn't match what you think is correct, would you be able to explain the behavior you are looking for here?

src/room/rpc/client/RpcClientManager.ts
1egoman: Good point, done in 068325a.

src/room/rpc/server/RpcServerManager.ts
1egoman: I'm not sure I follow this one - there isn't anything that `publishRpcAck` depends on being set as a prerequisite to operate properly? In the first race condition example, `pendingAcks` is checked by `this.publishRpcRequest`, but `this.publishRpcAck` does not have a similar type of dependency on `publishRpcpAck`l

Can you give me a step by step outline of what you are thinking? It's possible I'm missing something.

src/room/rpc/server/RpcServerManager.ts
1egoman: I could do something like this but I'm curious the use case you have in mind given that the failures are being sent back to the client already as a `RpcResponse`. Can you give me an example of what you'd do with these failures on the server end besides logging them or sending them to the client somehow?

src/room/rpc/server/RpcServerManager.ts
1egoman: Just generally - I left the `RpcError` fairly similar to how it was for RPC v1 but I think there's an opportunity to make some more substantial updates which would make something like this much more elegant (ie, port it to extend `LivekitReasonedError`, make the `RpcError.ErrorCode` entries the `LivekitReasonedError` generic, add support for `cause` like how the data track errors work, etc). I think this would all be possible to do in a backwards compatible way.

Do you think a broader `RpcError` refactor here would better address the root concern or is this concern specific to this one place?

src/room/rpc/utils.ts
1egoman: Good idea, done in 0eb011ea

src/room/rpc/server/RpcServerManager.ts
xianshijing-lk: Right, I overlooked it. There is no race condition here.

src/room/rpc/server/RpcServerManager.ts
1egoman: I thought about this overnight - I think at a minimum, I should add the `cause` stuff in here. But `LivekitReasonedError` being subclassed (with potentially new "types" being introduced implicitly as part of the subclass) I think is a fairly weird pattern so I'm not sure that part makes sense to do.

src/room/rpc/server/RpcServerManager.ts
lukasIO: Primarily a way for users to gather a bit more insight as to why something might be failing. The `Error` object of the callback could include stack traces and the like.

src/room/rpc/server/RpcServerManager.ts
lukasIO: I mean there's a `message` already on the RpcError and just a bit more insight than a very opaque `APPLICATION_ERROR` might be useful. 

However if we have the `onError` callback on the server side, then this could probably also stay opaque to the client calling

src/room/rpc/client/RpcClientManager.ts
lukasIO: That makes sense, just wanted to double check we're not leaving something pending here.

lukasIO (approved):

src/room/rpc/server/RpcServerManager.ts
1egoman: 👍 I will add some custom messages and where relevant a `cause` to link the errors so they display all inline when thrown, which I think should make this a bit better.

1egoman: Done in 80a0a904

src/room/rpc/server/RpcServerManager.ts
1egoman: ~Ah, I see - so this would be primarily for getting insight into errors coming from the server manager, NOT errors coming from the client. I think that makes sense - I will add this!~

I started working on this but realized that it is going to be fairly annoying for a chunk of the errors because there isn't a "rpc handler" yet in scope, so there's no obvious error callback. Therefore I think this probably would need to be global and not per-handler, but I think given I don't exactly have a use case in mind I'd prefer to add this if there is demand from users and not pre-emptively.

xianshijing-lk:
Are we planning to land this PR ? anything blocking us ?

1egoman:
Waiting for the rust version (https://github.com/livekit/rust-sdks/pull/1013) to be reviewed, and once it is / any relevant updates are made I will merge these two together.

RPC_SPEC.md
1egoman: ^ The above features notion has started to be implemented as "capabilities" in the protocol repo: https://github.com/livekit/protocol/pull/1505

1egoman:
(rebased on top of latest main)

Also, added a new rpc testing interface to the main demo app, for some more through manual testing:

Screenshot 2026-05-13 at 9 42 30 AM
Screenshot 2026-05-13 at 9 42 21 AM

PR #1873, opened 2026-04-06
Make "connected to" log show up after every join, not just after fresh connections

1egoman:
Previously, the code roughly did this:
```tsx
// in Room.connectSignal:
const joinResponse = await engine.join(/* ... */);

this.serverInfo = /* computed from joinResponse */;

this.log.debug('connected to Livekit Server ... (with this.serverInfo data in here)'); 
```

This meant that the `connected to Livekit Server ...` log would log only on connect, not for reconnects.

So instead, I moved the log one level deeper, into `RTCEngine.join`:
```tsx
// in RTCEngine.join:
// (preexisting join logic unchanged, joinResponse still comes from lower level this.client.join call)
const serverInfo = /* computed from joinResponse */;
this.log.debug('connected to Livekit Server ... (with serverInfo data in here)'); 
return { joinResponse, serverInfo };

// in Room.connectSignal:
const { joinResponse, serverInfo } = await engine.join(/* ... */);
this.serverInfo = serverInfo;
// (Log not here anymore)
```

With this change, the log shows up anytime a join occurs, whether it be fresh or for a reconnect. And the log still shows up in the same place on a fresh connect so the existing behavior should remain the same.

@boks1971 Does this address your concern? Is there a need to somehow disambiguate the log that is shown to determine if it is from a fresh connect or some type of reconnect, or is the context before / after the log good enough to tell this?

### DIFF
--- a/.changeset/spotty-pumas-jog.md
+++ b/.changeset/spotty-pumas-jog.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Moves "connected to Livekit Server" log within RTCEngine.join so it shows up for reconnects as well as connects
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -26,6 +26,7 @@ import {
   RoomMovedResponse,
   RpcAck,
   RpcResponse,
+  ServerInfo,
   SessionDescription,
   SignalTarget,
   SpeakerInfo,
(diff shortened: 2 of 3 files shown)


boks1971 (approved):
that's good @1egoman . Can get the connect/reconnect from context.

lukasIO (approved):
nice!

PR #1876, opened 2026-04-08
Add RTCEngine sendLossyBytes explicit queue to better handle sending large data track packets

1egoman:
Previously, sendLossyBytes ran `await this.waitForBufferStatusLow(kind);` to pause before enqueuing new data track packets onto the `_data_tracks` rtc data channel. In local cases this worked fairly well and the promises seemed to resolve in the order they were triggered.

However, when running the e2e tests, @boks1971 noticed an excessive number of out of order packets - in some cases the e2e tests returning as low as a 20% delivery rate for the low frequency large payload test.

I had a suspicion this `await this.waitForBufferStatusLow(kind);` bit was the problem, so I swapped it over to an explicit queuing implementation, and when I did, the packet delivery rate went up to 89-90%. In hindsight, relying on the promise resolution order like this was sloppy so I'm glad I'm getting this fixed now.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -249,7 +249,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   /** used to buffer lossy data track packets which arrive quickly so they don't overwhelm the data
    * channel buffer */
-  private lossyBytesWaitBuffer = new Map<DataChannelKind, Array<Future<void, never>>>();
+  private lossyBytesWaitQueue = new Map<DataChannelKind, Array<Future<void, never>>>();
   private lossyBytesMutexByKind = new Map<DataChannelKind, Mutex>();
 
   constructor(private options: InternalRoomOptions) {
@@ -1500,9 +1500,9 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
           // drain and go into "low status" before continuing.
           if (!this.isBufferStatusLow(kind)) {
             const future = new Future<void, never>();
-            const entries = this.lossyBytesWaitBuffer.get(kind) ?? [];
+            const entries = this.lossyBytesWaitQueue.get(kind) ?? [];
             entries.push(future);
-            this.lossyBytesWaitBuffer.set(kind, entries);
+            this.lossyBytesWaitQueue.set(kind, entries);
             await future.promise;
           }
           break;
@@ -1560,7 +1560,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
       // Now that there is space on the data channel buffer, attempt to fill up the remaining space
       // with bytes ready to be sent.
-      const buffer = this.lossyBytesWaitBuffer.get(kind) ?? [];
+      const buffer = this.lossyBytesWaitQueue.get(kind) ?? [];
       while (this.isBufferStatusLow(kind)) {
         let continueSendingFuture = buffer.shift();
         if (!continueSendingFuture) {
@@ -1569,7 +1569,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
         continueSendingFuture.resolve?.();
       }
       if (buffer.length === 0) {
-        this.lossyBytesWaitBuffer.delete(kind);
+        this.lossyBytesWaitQueue.delete(kind);
       }
     }
   };
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -85,6 +85,7 @@ import type { TrackPublishOptions, VideoCodec } from './track/options';
 import { getTrackPublicationInfo } from './track/utils';
 import type { LoggerOptions } from './types';
 import {
+  Future,
   isCompressionStreamSupported,
   isVideoCodec,
   isVideoTrack,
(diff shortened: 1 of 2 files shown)


src/room/RTCEngine.ts
```diff
@@ -1494,9 +1480,19 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+        // packets doesn't have the last half of the packets dropped due to not fitting into the
+        // data channel buffer all at once.
         switch (bufferStatusLowBehavior) {
           case 'wait':
-            await this.waitForBufferStatusLow(kind);
+            if (this.isBufferStatusLow(kind)) {
+              const future = new Future<void, never>();
+              const entries = this.lossyBytesWaitBuffer.get(kind) ?? [];
+              entries.push(future);
+              this.lossyBytesWaitBuffer.set(kind, entries);
```
lukasIO: probably unlikely, but wondering if there is the possibility for this to race where the `.set` call overrides with outdated entries when multiple packets are being sent in parallel?

src/room/RTCEngine.ts
lukasIO, line 1480: thought(non-blocking-just-context): an alternative would be to wrapt this function (its implementation) in a kind-dependent `Mutex` and buffer all calls that way. This would create back pressure inherently.

src/room/RTCEngine.ts
```diff
@@ -1494,9 +1480,19 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
         // Depending on the exact circumstance that data is being sent, either drop or wait for the
         // buffer status to not be low before continuing.
+        //
+        // An example of where this is used: data tracks, so that a large DataTrackFrame's worth of
+        // packets doesn't have the last half of the packets dropped due to not fitting into the
+        // data channel buffer all at once.
         switch (bufferStatusLowBehavior) {
           case 'wait':
-            await this.waitForBufferStatusLow(kind);
+            if (this.isBufferStatusLow(kind)) {
```
lukasIO: you're checking for whether the status is `low` here and in that case buffering the message 👀 

I think either we want to put all msgs in `wait` in that buffer or negate that condition?

If we negate it, I think there's still the potential for this to race as we're checking for the bufferstatus here and also when emptying the lossyBytesBuffer. I can imagine that we would run into situations where a packet sneaks through out of order that way

1egoman: Huh, very interesting... I am realizing that because of this incorrect negation when I ran the e2e tests my test results were invalid. Let me fix that and retest this and make sure it still has the desired effect.

Putting that aide... @lukasIO I'm not sure I understand where the race could occur. There's only one js thread involved so `sendLossyBytes` and `updateAndEmitDCBufferStatus` should never be running actually in parallel. If `lossyBytesWaitBuffer` is halfway through being cleared when `continueSendingFuture.promise` resolves then new entries will get pushed on the end (it's a queue, though now I wonder if maybe it should be renamed to make that more clear). It's possible I'm missing something though, can you give me a step by step situation outlining how a packet would end up out of order?

src/room/RTCEngine.ts
lukasIO: that's not how the event loop works, it turns out :D

src/room/RTCEngine.ts
1egoman: Chatted with lukas out of band - this is the problem he is identifying, rephrased:
> The problem isn't that packets could get enqueued into lossyBytesWaitBuffer out of order, the problem is that packets could be halfway through draining from lossyBytesWaitBuffer and ANOTHER unrelated packet could come in and get interleaved in the middle while that "drain" is occurring

I think it's a good point, and I hadn't considered that case!

1egoman: I believe I have addressed this in 2d30ccd.

I should say also, I'm still not 100% sure that it actually is a good tradeoff for data tracks to be buffering out of order packets like this, given the focus on timely delivery over reliability. @ladvoc I'm curious if you have any thoughts here!

src/room/RTCEngine.ts
1egoman: Talked about this in slack, this should be fine as is.

src/room/RTCEngine.ts
1egoman: FWIW - this was what I largely ended up doing. That being said, there still needs to be an individual queue of futures though, so I couldn't just do a wholesale replacement, because the draining code needs to "call back" into `sendLossyBytes` via the future to actually call `dc.send(...)`. But the mutex addition should mean that packets won't be interleaved during a buffer drain at least.

src/room/RTCEngine.ts
lukasIO: What I meant with the mutex approach is minimising the required changes in this PR and simply ensuring that the resolve order of that promise chain is strict, I opened this https://github.com/livekit/client-sdk-js/pull/1877 as an alternative to the mutex+buffer approach taken here, let me know what you think!

1egoman: Posted some rough numbers here for your alternate solution: https://github.com/livekit/client-sdk-js/pull/1877#issuecomment-4217421104. TL;DR though it performs roughly the same time wise but results in a much more variable amount of dropped packets run to run.

1egoman:
Things have shifted a bit since this was first opened, and I think generally we have decided to move in a slightly different direction. Instead of solving this problem on the sending end, we decided to:
1. Allow a subscriber to tolerate out of order packets by making the number of in flight frames the depacketizer managed configurable: https://github.com/livekit/client-sdk-js/pull/1923
2. Allow the sender to explicitly "flush" a data track if they want to ensure that all data has been sent: https://github.com/livekit/client-sdk-js/pull/1925

Closing this in favor of those two other changes!

PR #1879, opened 2026-04-10
Close streamcontrollers when datatrack gets unpublished

lukasIO:
Ensures that `read()` promises of a data track subscription are resolved once the publisher unpublishes the data track


1egoman (approved):
Nice! I did forget about this, thanks for catching this!

PR #1880, opened 2026-04-13
fix: avoid hard DOMException dependency in RN Hermes

KinanLak:
## Summary
- avoid hard dependency on global `DOMException` in runtime paths that run on React Native Hermes
- replace direct `new DOMException(...)` and `instanceof DOMException` usage with helper utilities that gracefully fall back when `DOMException` is unavailable
- remove `extends DOMException` from `DeferrableMapAbortError` to prevent module-load crash and add tests + patch changeset

## Motivation
`livekit-client@2.18.1` can crash at module load in React Native Hermes with:
`ReferenceError: Property 'DOMException' doesn't exist`

This is triggered by top-level `class DeferrableMapAbortError extends DOMException` and other direct `DOMException` references.

Fixes #1871

## Validation
- `pnpm format:check`
- `pnpm throws:check`
- `pnpm test`
- `pnpm build:clean`
- `pnpm compat`

### DIFF
--- a/.changeset/neat-pugs-tickle.md
+++ b/.changeset/neat-pugs-tickle.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Avoid hard dependency on global `DOMException` so React Native Hermes environments can initialize `livekit-client` without a manual polyfill.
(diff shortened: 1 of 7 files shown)


1egoman:
@davidliu Can you take a look at this?

davidliu:
Hmm, this will probably be better served by a polyfill on the RN SDK that loads before it touches the `livekit-client` package. We've had a couple of these "remove web-only APIs" commits, but it basically causes a legacy restriction on the JS SDK to remember and not touch going forward.

PR #1882, opened 2026-04-13
Wrap setE2EEEnabled call in a mutex

lukasIO:
with multiple parallel calls to `room.setE2EEEnabled` the room could get into a wrong state with publications timing out.

This PR ensures that 
- only one setE2EEEnabled call can be processed at a time, queuing up the next ones
- not doing any unnecessary work if the e2ee state is already correct

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -187,6 +187,8 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
 
   private e2eeManager: BaseE2EEManager | undefined;
 
+  private e2eeStateMutex: Mutex = new Mutex();
+
   private connectionReconcileInterval?: ReturnType<typeof setInterval>;
 
   private regionUrlProvider?: RegionUrlProvider;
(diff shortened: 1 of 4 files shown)


src/room/Room.ts
```diff
@@ -450,6 +459,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
         this.emit(RoomEvent.EncryptionError, error, participant);
       });
       this.e2eeManager?.setup(this);
+      this.e2eeManager?.setupEngine(this.engine);
```
1egoman, line 462: question: I am now sort of wondering why this wasn't here before... 🤔. Was this a small drive by fix or is this somehow integral to this change generally?

src/room/Room.ts
```diff
@@ -411,13 +412,21 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
   async setE2EEEnabled(enabled: boolean) {
-    if (this.e2eeManager) {
-      await Promise.all([this.localParticipant.setE2EEEnabled(enabled)]);
-      if (this.localParticipant.identity !== '') {
-        this.e2eeManager.setParticipantCryptorEnabled(enabled, this.localParticipant.identity);
+    const unlock = await this.e2eeStateMutex.lock();
+    try {
+      if (this.e2eeManager) {
+        if (this.isE2EEEnabled !== enabled) {
+          await Promise.all([this.localParticipant.setE2EEEnabled(enabled)]);
```
1egoman: nitpick: Is there some reason this has to be wrapped in a `Promise.all`? I see that is how it was before but I just want to make sure there's not some important reason why this is being done I am missing here.

1egoman (approved):

src/room/Room.ts
lukasIO: we recently made a change to only re-create non-new engines and I believe this is what makes this change necessary

src/room/Room.ts
lukasIO: no reason, just didn't change it, will change it now

PR #1885, opened 2026-04-14
Move over serializer implementation from components-js

1egoman:
Introduces a new set of message serialization primitives which will be used now in `useRpc` (in components-js), and in the future will be used for data tracks and (maybe) data streams.

Extracted from here: https://github.com/livekit/components-js/pull/1314


lukasIO (approved):
lgtm!

PR #1886, opened 2026-04-14
Add support for packet trailer features

chenosaurus:
- Add support for packet trailer features (user timestamp, frame ID).
- Add new web worker for packet trailer parsing.
- update example/demo to have options to enable packet trailer & render the parsed metadata.
- Unit tests for packet trailer parsing.

Browsers tested:
- chrome + packet trailer ✅
- FF + packet trailer ✅
- Safari + packet trailer ✅
- chrome + packet trailer  + e2ee ✅
- FF + packet trailer + e2ee ✅
- Safari + packet trailer + e2ee ✅
- chrome + e2ee ✅
- FF + e2ee ✅
- Safari + e2ee ✅
- chrome + packet trailer track w/o packet trailer worker ✅
- FF + packet trailer track w/o packet trailer worker ✅
- Safari + packet trailer track w/o packet trailer worker ✅


src/e2ee/E2eeManager.ts
```diff
@@ -479,11 +510,15 @@ export class E2EEManager
...
+    // @ts-ignore -- readableStream is set by PacketTrailerManager when it has pre-processed the receiver
+    const ptPreProcessed = !!receiver.readableStream && !!receiver.writableStream;
+
     if (
       isScriptTransformSupported() &&
       // Chrome occasionally throws an `InvalidState` error when using script transforms directly after introducing this API in 141.
       // Disabling it for Chrome based browsers until the API has stabilized
-      !isChromiumBased()
+      !isChromiumBased() &&
+      !ptPreProcessed
```
lukasIO: I'm not sure I follow why this is being checked for here, can you elaborate on the intention behind it?

src/e2ee/packetTrailer.ts
```diff
@@ -0,0 +1,198 @@
+export const PACKET_TRAILER_MAGIC = Uint8Array.from([
```
lukasIO, line 4: have you considered signalling packet-trailer presence as part of the publication info instead of relying on checking every frame on every publication for the magic bytes?

src/e2ee/packetTrailer.ts
```diff
@@ -0,0 +1,198 @@
...
+
+export function getFrameRtpTimestamp(
+  frame: RTCEncodedVideoFrame | RTCEncodedAudioFrame,
+): number | undefined {
+  try {
+    const metadata = frame.getMetadata() as Record<string, unknown>;
+    if (typeof metadata.rtpTimestamp === 'number') {
+      return metadata.rtpTimestamp;
+    }
+    if (typeof metadata.timestamp === 'number') {
```
lukasIO, line 204: what is this referring to? I don't see a `timestamp` value on `getMetadata` [here](https://developer.mozilla.org/en-US/docs/Web/API/RTCEncodedVideoFrame/getMetadata#browser_compatibility)

src/e2ee/packetTrailer.ts
lukasIO, line 1: is it intentional that this file lives in the `/e2ee` folder? 👀

src/version.ts
```diff
@@ -1,4 +1,4 @@
 import { version as v } from '../package.json';
 
 export const version = v;
-export const protocolVersion = 16;
+export const protocolVersion = 17;
```
lukasIO, line 4: The related protocol PR talks about stripping the trailer on the SFU side, but I don't think bumping protocolVersion is enough: what happens if a client with protocolVersion 17 doesn't have the trailer worker set?

src/e2ee/worker/FrameCryptor.ts
```diff
@@ -454,6 +460,34 @@ export class FrameCryptor extends BaseFrameCryptor {
...
+            postMessage(msg);
+          }
+          encodedFrame.data = (ptResult.data.buffer as ArrayBuffer).slice(
+            ptResult.data.byteOffset,
+            ptResult.data.byteOffset + ptResult.data.byteLength,
+          );
+        }
+      } catch {
+        // best-effort: never break the media pipeline if trailer parsing fails
+      }
```
1egoman, line 527: question: Is it worth at least logging something here to make it clear that a failure happened?

src/e2ee/packetTrailer.ts
```diff
@@ -0,0 +1,198 @@
...
+      return metadata.timestamp;
+    }
+  } catch {
+    // getMetadata() might not be available
+  }
```
1egoman, line 209: suggestion: Another place where it might be worth logging

src/packetTrailer/PacketTrailerManager.ts
```diff
@@ -0,0 +1,148 @@
...
+ * Manages packet trailer extraction for received video tracks.
+ *
+ * When a track's TrackInfo indicates packet trailer features, the manager
+ * wires up an Insertable Streams pipeline (via a dedicated worker) to strip
+ * the trailer from encoded frames and cache the metadata for lookup.
+ *
+ * When E2EE is active, the E2EE FrameCryptor worker handles trailer
+ * extraction directly (before decryption), so this manager only creates
+ * the extractor/metadata cache — no separate insertable-streams pipeline
+ * is needed.
```
1egoman: question: I'm not sure I understand why this packet trailer logic exists both in the e2ee worker and in the new packet trailer worker. Can you shed some light on why you made this decision? Does it have to do with the goal of encrypting the packet payload and trailer separately so the SFU can strip out the trailer without that process requiring decrypting a single large block?

I don't have enough context to be incredibly confident in this suggestion, but if you can't attach multiple transforms which run in series there might be some value in adding some sort of "stream transform composer" layer above both this and the e2ee manager which would connect a series of transforms together and allows each to fully own their respective responsibilities.

src/options.ts
```diff
@@ -100,6 +101,13 @@ export interface InternalRoomOptions {
 
   loggerName?: string;
 
+  /**
+   * @experimental
+   * Options for enabling packet trailer extraction on received video tracks.
+   * Packet trailers carry frame-level metadata such as user timestamps and frame IDs.
+   */
+  packetTrailer?: PacketTrailerOptions;
+
```
1egoman, line 110: thought: After this gets eventually is merged, you may want to add a todo for yourself to wire this through in components-js in a similar way to what we recently did for `encryption`, see here for more info: https://github.com/livekit/components-js/pull/1317

src/version.ts
1egoman: ^ This is probably my biggest question as well.

It sort of sounds like you want a "feature" you can communicate support of unrelated to a protocol / client version. I recently wrote a note about this in the context of the in flight rpc protocol updates here: https://github.com/livekit/client-sdk-js/pull/1832#discussion_r3065961915

Keeping in mind we haven't discussed mechanically how this pattern would work yet (ie, new fields in `ParticipantInfo`, something in the `JoinResponse`, etc?), I think if you did something like that then the SFU could know if packet trailer processing is enabled on the client independent of the protocol version and opt to strip or not strip conditionally.

Thoughts on that proposal?

chenosaurus: The packet trailer transform is lightweight enough to run on the main thread, we can as a backup attach the transform automatically (when worker is not passed in) when we see any tracks that declared packet trailer features in its TrackInfo.  That would always work and avoid having to change any signaling to the SFU.

1egoman: > The packet trailer transform is lightweight enough to run on the main thread

If this is the case, is there a reason why it shouldn't always be run on the main thread then rather than having the optional worker?

src/packetTrailer/PacketTrailerManager.ts
chenosaurus: The main reason is each track transceiver can only have a single transform attached, so if e2ee is enabled, we can't also attach the packetTrailer transform worker.  Thus the need to have the same logic inside e2ee for parsing packet trailers if both features are enabled.

I think your suggestion makes sense but would mean a much bigger refactor where e2eeManager & e2ee worker needs to be replaced by a more generic frameTransformManager & a new worker that can handle both e2ee & packetTrailer.

src/e2ee/E2eeManager.ts
chenosaurus: removed, it was left over.

src/e2ee/packetTrailer.ts
chenosaurus: There is already support on track publication to declare which packet trailer features are available on the track.  It's exposed to the subscriber in TrackInfo.  I have updated all place to check TrackInfo before attaching transformer.

src/e2ee/packetTrailer.ts
chenosaurus: cleaned up

src/e2ee/packetTrailer.ts
chenosaurus: no, that's a mistake.  Moved.

src/version.ts
chenosaurus: see response on a later comment, it's possible to attach the packet trailer transformer directly to the transceiver in the main thread to avoid the issue of worker not being initialized.  However, we should try to use the worker whenever possible to not have to process frames in the main thread.

src/e2ee/worker/FrameCryptor.ts
chenosaurus: this fires on every frame so it will be noisy.  If it's useful to log errors here we have a throttled log here.

src/e2ee/E2eeManager.ts
lukasIO, line 555: on transceiver reuse we need to signal to the worker the new state of `hasPacketTrailer` otherwise it will break if it changed between publications

src/e2ee/types.ts
```diff
@@ -51,6 +52,12 @@ export interface EncodeMessage extends BaseMessage {
     trackId: string;
     codec?: VideoCodec;
     isReuse: boolean;
+    /**
+     * Whether the published track advertises packet trailer features.
+     * When false, the cryptor skips the per-frame trailer extraction path
+     * entirely on decode.
+     */
+    hasPacketTrailer?: boolean;
```
lukasIO: let's make this a requirement instead of optional

src/e2ee/types.ts
```diff
@@ -221,4 +234,10 @@ export type ScriptTransformOptions = {
   participantIdentity: string;
   trackId: string;
   codec?: VideoCodec;
+  /**
+   * Whether the published track advertises packet trailer features.
+   * When false, the cryptor skips the per-frame trailer extraction path
+   * entirely on decode.
+   */
+  hasPacketTrailer?: boolean;
```
lukasIO: same here, requiring a boolean set here makes the intention clearer and shows places where we might unintentionally have missed adding it in

src/packetTrailer/PacketTrailerManager.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (room === this.room) {
+      return;
+    }
+    this.room = room;
+
+    if (this.worker) {
+      this.worker.onmessage = this.onWorkerMessage;
+      this.worker.onerror = this.onWorkerError;
+      this.worker.postMessage({ kind: 'init' });
+    }
```
1egoman, line 67: question: Is the intention that `this.worker.postMessage({ kind: 'init' });` should be run once per track, or once globally for the whole application?

src/packetTrailer/worker/packetTrailer.worker.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const pipelines = new Map<string, PipelineState>();
+
+onmessage = (ev: MessageEvent<PTWorkerMessage>) => {
+  const msg = ev.data;
+
+  switch (msg.kind) {
+    case 'init':
+      postMessage({ kind: 'initAck' });
+      break;
```
1egoman, line 28: question: What is `init` for - it looks like it is a no-op here right now?

src/room/track/RemoteVideoTrack.ts
```diff
@@ -38,6 +43,23 @@ export default class RemoteVideoTrack extends RemoteTrack<Track.Kind.Video> {
...
+   *
+   */
+  lookupFrameMetadata({
+    rtpTimestamp,
+  }: {
+    rtpTimestamp: number;
+  }): PacketTrailerMetadata | undefined {
+    return this.packetTrailerExtractor?.lookupMetadata(rtpTimestamp);
+  }
+
```
1egoman, line 62: question: Is there a need for some sort of async version of this lookup function to let a user asynchronously wait for a new rtp timestamp entry to show up given it's generated by the worker and could potentially race with frame generation?

To be clear - not sure if this is important or not, so if there isn't a need for this then disregard.

If you do want to do this though, check out `DeferrableMap` in `utils` which makes this pretty simple (it's what is being used to power similar type of behavior for remote data track lookup)

src/room/RTCEngine.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  it('enables encoded insertable streams when a packet trailer worker is configured', () => {
+    stubInsertableStreamsSupport();
+
+    const engine = new RTCEngine({
+      ...roomOptionDefaults,
+      packetTrailer: { worker: {} as Worker },
+    });
+
+    expect(makeRTCConfiguration(engine).encodedInsertableStreams).toBe(true);
+  });
```
1egoman, line 84: thought: If it's not too burdensome, it might be nice for all these tests to be smaller unit tests which don't require constructing the whole RTCEngine.

Context: I worry these in their current state will be annoying to maintain given RTCEngine tends to have quite a few tendrils into other systems / browser globals which could get hit if the exact access patterns of these tests change.

.size-limit.cjs
```diff
@@ -2,7 +2,7 @@ module.exports = [
   {
     path: 'dist/livekit-client.esm.mjs',
     import: '{ Room }',
-    limit: '100 kB',
+    limit: '101 kB',
```
1egoman, line 5: question: Was this an intentional change?

1egoman (commented):
FYI that some new linting updates [were just merged](https://github.com/livekit/client-sdk-js/commit/cbfb29835ba45a4c19dbc9efed7d2d2b0f137921) so you may want to rebase this and fix any lint errors.

Generally looks good to me and none of my comments are blocking. Lukas has been closer to this / testing it more directly so I'll save the final approval to him.

.size-limit.cjs
chenosaurus: yea, w/ the changes the bundle got just over 100kB and was failing the size limit check.  I had to bump it up to 101kB.

src/packetTrailer/worker/packetTrailer.worker.ts
chenosaurus: this pattern was carried over from the e2ee worker.  However you're right that in the packetTariler worker it doesn't actually do anything while the e2eeWorker has some post init work.  I left it in for consistency but should be ok to completely remove as well.

src/room/track/RemoteVideoTrack.ts
chenosaurus: there shouldn't be a case where the metadata is not available when the frame is rendered.  We also want this to be synchronous and fail immediately if the metadata is not in the map.

lukasIO (approved):

PR #1888, opened 2026-04-15
Harden remote data tracks map

1egoman:
An edge case was observed by a customer:
1. Participant A joins and publishes a data track.
2. Participant B calls `room.connect()`, joins the room, and the `IncomingDataTrackManager` "trackPublished" event fires for all newly discovered data tracks
3. Due to the order in which events process, `room.remoteParticipants` is unset while the "trackPublished" event handler fires, so the data tracks don't get set in the `RemoteParticipant.dataTracks` deferrable map.
4. The user can't find the data tracks in the map and are confused.

To fix this, add logic to ensure that the remote data tracks are set on the map immediately when the remote participant connects, using the data in the `ParticipantInfo` signaling struct. This means that the existing "trackPublished" event path is only ever handling additions AFTER the room.connect call which works because the Room state is fully initialized at that point.

Duplicate events are fine here since the operation is idempotent - `RemoteDataTrack` is purely a pointer to state in `IncomingDataTrackManager` and multiple inserts under the same key result in the same result map.

### DIFF
--- a/src/room/participant/RemoteParticipant.ts
+++ b/src/room/participant/RemoteParticipant.ts
@@ -6,7 +6,9 @@ import type {
 } from '@livekit/protocol';
 import type { SignalClient } from '../../api/SignalClient';
 import { DeferrableMap } from '../../utils/deferrable-map';
-import type RemoteDataTrack from '../data-track/RemoteDataTrack';
+import RemoteDataTrack from '../data-track/RemoteDataTrack';
+import type IncomingDataTrackManager from '../data-track/incoming/IncomingDataTrackManager';
+import { DataTrackInfo } from '../data-track/types';
 import { ParticipantEvent, TrackEvent } from '../events';
 import RemoteAudioTrack from '../track/RemoteAudioTrack';
 import type RemoteTrack from '../track/RemoteTrack';
@@ -48,6 +50,7 @@ export default class RemoteParticipant extends Participant {
     signalClient: SignalClient,
     pi: ParticipantInfo,
     loggerOptions: LoggerOptions,
+    manager: IncomingDataTrackManager,
   ): RemoteParticipant {
     return new RemoteParticipant(
       signalClient,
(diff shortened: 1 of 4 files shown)


src/utils/deferrable-map.ts
```diff
@@ -29,7 +29,7 @@ export class DeferrableMap<K, V> extends Map<K, V> {
     super.set(key, value);
 
     // Resolve any futures waiting on this key.
-    const futures = this.pending.get(key);
+    const futures = this.pending?.get(key);
```
1egoman, line 32: Note to reviewers - this was unexpected / weird. It seems by extending `Map` like `DeferrableMap` is doing, calling `new DeferrableMap([["key", "value"]])` actually behind the scenes (at least in chrome) calls `set` on the map for each tuple passed. So this means that this method can actually be called before the class is actually set up / `this.pending` is defined which results in this line throwing an error unless the conditional access is added here.

xianshijing-lk (approved):

src/room/participant/RemoteParticipant.ts
```diff
@@ -79,13 +86,18 @@ export default class RemoteParticipant extends Participant {
...
   ) {
     super(sid, identity || '', name, metadata, attributes, loggerOptions, kind);
     this.signalClient = signalClient;
     this.trackPublications = new Map();
     this.audioTrackPublications = new Map();
     this.videoTrackPublications = new Map();
-    this.dataTracks = new DeferrableMap();
+    this.dataTracks = new DeferrableMap(
+      remoteDataTracks.map((remoteDataTrack) => {
+        return [remoteDataTrack.info.name, remoteDataTrack];
```
xianshijing-lk, line 98: question

any chance that info does not exist

lukasIO (approved):
this makes sense. 

Just for context: we also have a `this.emitWhenConnected` helper that buffers events in a queue until the room connection is fully established

src/room/participant/RemoteParticipant.ts
1egoman: Nope, it should always be there, it's a required field in the protobuf structure.

PR #1892, opened 2026-04-16
Fix some data track subscription edge cases

1egoman:
A customer surfaced some issues with the `signal` parameter passed to data track subscriptions - when this was passed and aborted during certain parts of the subscription / participant connection lifecycle outside the happy path, it was possible that things could get into an inconsistent state:

- Aborting after participant disconnect - the stream would throw a `Cannot close an errored readable stream` error in this case because the manager tried to clean up a subscription that the user had already manually aborted. So, make sure the stream controller gets removed from the list so it doesn't get double-closed.

- The data track is unpublished during the brief window of time between the sfu subscription being established and the readable stream finishing being setup. If this happened, the readable stream would hang forever isntead of explicitly aborting.

- Abort signal listener not being cleaned up properly in all cases - if the manager closes the stream (the stream doesn't get into an error state, it is explicitly closed), then this wouldn't clean up the abort signal listener, so if a user triggered the abort signal later, it would cause an error to be throw.

These cases now all have explicit tests that test a few variations to ensure that this doesn't regress.

### DIFF
--- a/src/room/data-track/incoming/IncomingDataTrackManager.ts
+++ b/src/room/data-track/incoming/IncomingDataTrackManager.ts
@@ -126,15 +126,51 @@ export default class IncomingDataTrackManager extends (EventEmitter as new () =>
     let streamController: ReadableStreamDefaultController<DataTrackFrame> | null = null;
     const sfuSubscriptionComplete = new Future<void, DataTrackSubscribeError>();
 
+    const cleanup = () => {
+      signal?.removeEventListener('abort', onAbort);
+
+      if (!streamController) {
+        log.warn(`ReadableStream subscribed to ${sid} was not started.`);
+        return;
+      }
+      const descriptor = this.descriptors.get(sid);
+      if (!descriptor) {
+        log.warn(`Unknown track ${sid}, skipping cancel...`);
+        return;
+      }
+      if (descriptor.subscription.type !== 'active') {
+        log.warn(`Subscription for track ${sid} is not active, skipping cancel...`);
+        return;
+      }
+
+      descriptor.subscription.streamControllers.delete(streamController);
+
+      // If no active stream controllers are left, also unsubscribe on the SFU end.
+      if (descriptor.subscription.streamControllers.size === 0) {
+        this.unSubscribeRequest(descriptor.info.sid);
+      }
+    };
+
+    const onAbort = () => {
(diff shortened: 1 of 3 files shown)


chenosaurus (approved):

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -126,62 +129,93 @@ export default class IncomingDataTrackManager extends (EventEmitter as new () =>
...
+      }
+    };
+
+    const onAbort = () => {
+      if (!streamController) {
+        return;
+      }
+      const currentDescriptor = this.descriptors.get(sid);
+      if (currentDescriptor?.subscription.type === 'active') {
+        currentDescriptor.subscription.streamControllers.delete(streamController);
```
xianshijing-lk, line 167: I think cleanup() is also trying to clean up the streamController, should we move the code to one place instead?
like having cleanup() does the cleaning, and onAbort() handles the signaling and cleanup

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -126,62 +129,93 @@ export default class IncomingDataTrackManager extends (EventEmitter as new () =>
...
+      }
+
+      streamController.error(DataTrackSubscribeError.cancelled());
```
xianshijing-lk, line 170: onAbort() is called after abort(), does streamController guarantee that it is not closed yet ? 

and will it throw if streamController is closed already ?

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: Ah, good point - this is I think duplicative. I will remove this and let the copy in `cleanup` handle deleting the stream controller from the map instead. 👍

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: I think it should be ok, [the mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultController/error) for the error method explicitly say that it can be called multiple times.

lukasIO (approved):
lgtm!

PR #1894, opened 2026-04-16
TokenSource - add some more comprehensive tests

1egoman:
After the recent situation where a hosted agent broke due to a backwards incompatible protobuf message change which was not being handled properly (https://github.com/livekit/client-sdk-js/pull/1889), it made me realize that some more comprehensive integration tests of `TokenSource` would have caught this prior to merging. Previously, there were just some unit tests for utility functions that the `TokenSource` used but nothing comprehensive.

So, this pull request adds:
1. One more case to the previously existing unit tests to test token parsing specifically to ensure that extra fields are skipped
2. A more comprehensive test suite running through the whole `TokenSourceConfigurable` surface area, testing request/response serialization+deserialization, token caching behavior (another thing [there were some recent changes](https://github.com/livekit/client-sdk-js/pull/1839) made to), and synchronization between concurrent token fetches.

### DIFF
--- a/src/room/token-source/TokenSource.test.ts
+++ b/src/room/token-source/TokenSource.test.ts
@@ -1,37 +1,60 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
+import { sleep } from '../utils';
 import { TokenSource } from './TokenSource';
-import { TokenSourceFetchOptions } from './types';
-import { RoomConfiguration } from '@livekit/protocol';
+import { TOKENS } from './test-tokens';
+import { TokenSourceFetchOptions, TokenSourceResponseObject } from './types';
 
 const EXAMPLE_FETCH_OPTIONS: TokenSourceFetchOptions = {
-  roomName: "room name",
-  participantName: "participant name",
-  participantIdentity: "participant identity",
+  roomName: 'room name',
+  participantName: 'participant name',
+  participantIdentity: 'participant identity',
   participantMetadata: '{"example": "metadata here"}',
   participantAttributes: {},
 
-  agentName: "agent name",
+  agentName: 'agent name',
   agentMetadata: '{"example": "agent metadata here"}',
 };
 
 const EXAMPLE_TOKEN_ENDPOINT_RESPONSE_JSON = {
   server_url: 'wss://localhost:7000',
-  participant_token: 'bogus token'
+  participant_token: 'bogus token',
 };
+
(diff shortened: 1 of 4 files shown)


src/room/token-source/utils.test.ts
```diff
@@ -60,6 +41,17 @@ describe('decodeTokenPayload', () => {
...
+    const payload = decodeTokenPayload(TOKENS.EXTRA_FIELDS);
+    expect(payload.roomConfig?.name).toBe('test room name');
+    expect(payload.roomConfig?.agents).toHaveLength(1);
+    expect(payload.roomConfig?.agents![0].agentName).toBe('test agent name');
+    expect(payload.roomConfig?.agents![0].metadata).toBe('test agent metadata');
+
+    // Make sure the extra fields aren't in the payload, just the ones in the protobuf
+    expect((payload.roomConfig as any)?.extraField).toBeUndefined();
+    expect((payload.roomConfig?.agents![0] as any)?.extraField).toBeUndefined();
+  });
```
1egoman, line 54: Note to reviewers - this test specifically ensures the below type case that was discovered in production works:

```tsx
const serverGeneratedRoomConfiguration = { /* other fields */, tags: {} };
// (gets sent to client with older @livekit/protocol)
RoomConfiguration.fromJson(serverGeneratedRoomConfiguration) // <= this was throwing an error
```

src/room/token-source/TokenSource.test.ts
```diff
@@ -0,0 +1,337 @@
...
+    await Promise.all([
+      tokenSource.fetch({ roomName: 'room-1' }),
+      tokenSource.fetch({ roomName: 'room-2' }),
+      tokenSource.fetch({ roomName: 'room-3' }),
+    ]);
+
+    // The mutex should ensure only one fetch runs at a time
+    expect(maxConcurrentCalls).toStrictEqual(1);
+    expect(customFn).toHaveBeenCalledTimes(3);
+  });
```
1egoman, line 336: @lukasIO It would be good to get confirmation from you that is a good test to validate the changes you made in https://github.com/livekit/client-sdk-js/pull/1839.

davidzhao (approved):
lgtm

thomasyuill-livekit (approved):
LGTM

src/room/token-source/TokenSource.test.ts
lukasIO: looks good! but I think it's more relevant to the `force` test above

lukasIO (approved):

src/room/token-source/test-tokens.ts
lukasIO, line 1: we should add this to the tsconfig exclude so that it doesn't end up unintentionally in the bundle

PR #1895, opened 2026-04-17
Harden extension computation and retry publishing on negotiation failure

lukasIO:


src/room/PCTransport.ts
```diff
@@ -609,9 +613,6 @@ export default class PCTransport extends EventEmitter {
       if (this.ddExtID === 0) {
         let maxID = 0;
         sdp.media.forEach((m) => {
-          if (m.type !== 'video') {
```
lukasIO, line 613: this is supposed to ensure that we're not reusing any extension ids (be it audio or video)

boks1971 (approved):
lgtm, but would be good to get another set of 👀 as I do not understand all the nuances of paths invoved.

lukasIO:
going to merge this for now.

@1egoman if you get a chance, I'd appreciate your 👀 on it still

1egoman (commented):
@lukasIO Like raja I am not exactly sure I fully understand all the pieces in play here but nothing looks incredibly concerning to me.

PR #1877, opened 2026-04-09
Improve ordering of resolving waitForBufferStatus calls

lukasIO:

PR #1896, opened 2026-04-17
fix: avoid attaching a new Closing event listener for each waitForBufferStatusLow call

1egoman:
In the `waitForBufferStatusLow` method of RTCEngine, a bunch of this.once calls are currently being made to listen for `RTCEngine` `Closing` events. This can result potentially in errors like the below if `waitForBufferStatusLow` is called at a high rate (the below was captured in firefox, fwiw):

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 closing  
listeners added. Use emitter.setMaxListeners() to increase limit                     
    _addListener events.js:211                                                       
    addListener events.js:227                                                        
    once events.js:257                                                               
    waitForBufferStatusLow RTCEngine.ts:1569                                         
    __awaiter livekit-client.esm.mjs:8244                                            
    __awaiter livekit-client.esm.mjs:8240                                            
    waitForBufferStatusLow RTCEngine.ts:1561                                         
    TypedPromise TypedPromise.ts:8                                                   
    waitForBufferStatusLow RTCEngine.ts:1561                                         
    __awaiter livekit-client.esm.mjs:8244                                            
    __awaiter livekit-client.esm.mjs:8240                                            
    waitForBufferStatusLow livekit-client.esm.mjs:21539                              
    sendLossyBytes RTCEngine.ts:1499                                                 
    fulfilled livekit-client.esm.mjs:8241                                            
    promise callback*step livekit-client.esm.mjs:8243                                
    __awaiter livekit-client.esm.mjs:8244                                            
    __awaiter livekit-client.esm.mjs:8240                                            
    sendLossyBytes RTCEngine.ts:1485                                                 
    Room Room.ts:294                                                                 
    emit events.js:153                                                               
    tryProcessAndSend OutgoingDataTrackManager.ts:157                                
```

While this seems to have been a problem in theory for a while, the recent shift to listening to `bufferedamountlow` events in https://github.com/livekit/client-sdk-js/pull/1877 rather than `await sleep(10)` in a loop I think has made this inadvertently worse, as the `sleep`s previously were giving the listeners time to drain and keeping the RtcEngine listener count below the threshold.

So instead, listen once and fan out that single event to all in flight `waitForBufferStatusLow` calls.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -1594,13 +1594,9 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
           return;
         }
         this.bufferStatusLowClosingFuture.promise.catch((e) => reject(e));
-        dc.addEventListener(
-          'bufferedamountlow',
-          () => resolve(),
-          {
-            once: true,
-          },
-        );
+        dc.addEventListener('bufferedamountlow', () => resolve(), {
+          once: true,
+        });
       }
     });
   }
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -250,7 +250,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
   /** used to indicate whether the browser is currently waiting to reconnect */
   private isWaitingForNetworkReconnect: boolean = false;
 
-  private bufferStatusLowClosingFuture = new Future<void, UnexpectedConnectionState>();
+  private bufferStatusLowClosingFuture = new Future<never, UnexpectedConnectionState>();
 
   constructor(private options: InternalRoomOptions) {
     super();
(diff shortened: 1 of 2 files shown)


1egoman:
I ran the end to end tests, and got a few failures which looked to be just flaky tests. So I don't think this broke anything.

src/room/RTCEngine.ts
```diff
@@ -249,6 +250,8 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
   /** used to indicate whether the browser is currently waiting to reconnect */
   private isWaitingForNetworkReconnect: boolean = false;
 
+  private bufferStatusLowClosingFuture = new Future<void, UnexpectedConnectionState>();
```
lukasIO: nit: should this be `new Future<never, UnexpectedConnectionState>()`

lukasIO (approved):

src/room/RTCEngine.ts
1egoman: Good point, updated. 1e97f7cb

PR #1900, opened 2026-04-20
fix: ensure updated tokens get set on the regionUrlProvider

lukasIO:
and a small refactor to avoid the duplicate ownership of regionUrlProvider in both room and engine

closes #1849

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -332,7 +332,6 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
         this.shouldFailOnV1Path = false;
         throw ConnectionError.serviceNotFound('Simulated v1 path failure', 'v0-rtc');
       }
-      log.warn('joining signal with ', url);
       const joinResponse = await this.client.join(
         url,
         token,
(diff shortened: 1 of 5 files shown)


src/room/RTCEngine.ts
```diff
@@ -1137,11 +1133,6 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     this.log.debug(`reconnecting in ${delay}ms`, this.logContext);
 
     this.clearReconnectTimeout();
-    if (this.token && this.regionUrlProvider) {
-      // token may have been refreshed, we do not want to recreate the regionUrlProvider
-      // since the current engine may have inherited a regional url
-      this.regionUrlProvider.updateToken(this.token);
-    }
```
1egoman, line 1144: question: Should there also be a `this.emit(EngineEvent.TokenRefreshed, token);` here?

1egoman (approved):

src/room/RTCEngine.ts
lukasIO: good question. this should not be needed anymore as we're passing the token along at each token refreshed event, adding it nonetheless to ensure this isn't breaking any previous expectations.

PR #1902, opened 2026-04-20
Re-compute encoding params after device change

### LINKED ISSUE #1884: Video encoding parameters not recalculated after switchActiveDevice / replaceTrack
### Describe the bug

When switching cameras during an active call via room.switchActiveDevice('videoinput', newDeviceId), the video quality degrades significantly. Toggling the camera off/on (unpublish + publish) restores quality.

Expected: Switching cameras produces the same video quality as toggling camera off/on.

Actual: After switching, the remote side receives degraded video (low resolution appearance), even though MediaStreamTrack.getSettings() reports correct resolution (e.g. 1280x720) on the sender side.

**Root cause in the code:**
switchActiveDevice → setDeviceId() → restartTrack() → restart() performs sender.replaceTrack(newTrack) but never recalculates encoding parameters (scaleResolutionDownBy, maxBitrate, maxFramerate).

**Specifically:**
1. LocalVideoTrack.restartTrack() (LocalVideoTrack.ts:249-268) — replaces the track on simulcast senders via sender.replaceTrack() but does not update encoding parameters.
2. LocalTrack.restart() (LocalTrack.ts:326-384) — calls setMediaStreamTrack() which does sender.replaceTrack() at line 197, again without updating encodings.
3. LocalParticipant.publishTrack() (LocalParticipant.ts:1139-1170) — the only place that calls computeVideoEncodings() to calculate scaleResolutionDownBy based on the track’s actual dimensions.
4. sender.setParameters() is used in setDegradationPreference, prioritizePerformance, and setPublishingLayersForSender — none of these recalculate encoding dimensions after track replacement.

So after replaceTrack, the sender keeps scaleResolutionDownBy values calculated for the original camera’s dimensions. If the new camera provides a different native resolution, the scaling produces incorrect output resolution on the encoded stream.

### Reproduction

1. Connect to a room and publish a camera track (camera A)
2. Call room.switchActiveDevice('videoinput', cameraBDeviceId) to switch to camera B
3. Observe degraded video quality on the remote participant’s side
4. Toggle camera off (localParticipant.setCameraEnabled(false)) then on (localParticipant.setCameraEnabled(true))
5. Observe quality is restored

The issue is more pronounced when camera A and camera B have different native resolutions or different default capture modes.

### Logs

```shell

```

### System Info

```shell
System:
  OS: macOS 26.3.1
  CPU: (10) arm64 Apple M4
Browsers:
  Chrome: 147.0.7727.56
npmPackages:
  livekit-client: 2.17.2
```

### Severity

annoyance

### Additional Information

_No response_

lukasIO:
closes https://github.com/livekit/client-sdk-js/issues/1884

### DIFF
--- a/src/room/track/LocalVideoTrack.ts
+++ b/src/room/track/LocalVideoTrack.ts
@@ -279,6 +279,10 @@ export default class LocalVideoTrack extends LocalTrack<Track.Kind.Video> {
     // leave the sender's encoding parameters (scaleResolutionDownBy, maxBitrate, etc.)
     // based on the old dimensions. Recompute them so the encoded output matches the
     // new source.
+    await this.onSenderTrackSwapped();
+  }
+
+  protected override async onSenderTrackSwapped(): Promise<void> {
     await this.refreshSenderEncodings();
   }
 
--- a/src/room/track/LocalVideoTrack.ts
+++ b/src/room/track/LocalVideoTrack.ts
@@ -69,6 +69,9 @@ export default class LocalVideoTrack extends LocalTrack<Track.Kind.Video> {
   /* @internal */
   publishOptions?: TrackPublishOptions;
 
+  /* @internal */
+  lastEncodedDimensions?: Track.Dimensions;
+
   get sender(): RTCRtpSender | undefined {
     return this._sender;
   }
@@ -308,6 +311,14 @@ export default class LocalVideoTrack extends LocalTrack<Track.Kind.Video> {
       return;
     }
 
+    if (
+      this.lastEncodedDimensions &&
+      this.lastEncodedDimensions.width === dims.width &&
+      this.lastEncodedDimensions.height === dims.height
+    ) {
+      return;
+    }
+
     const isScreenShare = this.source === Track.Source.ScreenShare;
     const newEncodings = computeVideoEncodings(isScreenShare, dims.width, dims.height, {
       ...this.publishOptions,
@@ -331,6 +342,8 @@ export default class LocalVideoTrack extends LocalTrack<Track.Kind.Video> {
       await this.applyEncodingsToSender(sc.sender, backupEncodings);
       sc.encodings = backupEncodings;
     }
+
+    this.lastEncodedDimensions = dims;
   }
 
   private async applyEncodingsToSender(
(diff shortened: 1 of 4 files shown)


1egoman (approved):

src/room/track/LocalVideoTrack.ts
```diff
@@ -265,6 +276,121 @@ export default class LocalVideoTrack extends LocalTrack<Track.Kind.Video> {
...
+      return;
+    }
+
+    const isScreenShare = this.source === Track.Source.ScreenShare;
+    const newEncodings = computeVideoEncodings(isScreenShare, dims.width, dims.height, {
+      ...this.publishOptions,
+    });
+
+    await this.applyEncodingsToSender(this.sender, newEncodings);
+    this.encodings = newEncodings;
```
xianshijing-lk: [question](https://pullpo.io/cc?l=question) 

what will happen if setParameters fails at the above this.applyEncodingsToSender( ? 

It looks like we still continue and set the encodings ? 

Is it intentional ?

src/room/track/LocalVideoTrack.ts
```diff
@@ -265,6 +276,121 @@ export default class LocalVideoTrack extends LocalTrack<Track.Kind.Video> {
...
+    ) {
+      return;
+    }
+
+    const isScreenShare = this.source === Track.Source.ScreenShare;
+    const newEncodings = computeVideoEncodings(isScreenShare, dims.width, dims.height, {
+      ...this.publishOptions,
+    });
+
+    await this.applyEncodingsToSender(this.sender, newEncodings);
```
xianshijing-lk: [question](https://pullpo.io/cc?l=question) 

what is the thread model for this LocalVideoTrack.ts? will it be called by multi threads? 

I wonder if this.sender would have been changed when it gets here, given things are async like
dims = await this.waitForDimensions();

xianshijing-lk (approved):
two questions, lgtm if you address them

src/room/track/LocalVideoTrack.ts
lukasIO: the `sender` property should only ever be set once for a given LocalTrack. We can think about moving the `senderLock` one level higher into `refreshSenderEncodings`. I'll do that now to be safe

src/room/track/LocalVideoTrack.ts
lukasIO: that's a good point, I moved things around and it fits quite nicely with pushing the `senderLock` one level higher

PR #1904, opened 2026-04-20
Ignore engine closed events for data track packets which are in flight / un delivered on engine close

1egoman:
I have noticed in running the data track test, that "engine closed" gets thrown fairly often at the end of the test, AFTER the actual test case passes. I had seemingly fixed this in https://github.com/livekit/client-sdk-js/pull/1892 but after merging this on top of `main`, it turns out https://github.com/livekit/client-sdk-js/pull/1877 changed the semantics of how this worked and caused it to break subtly in a different way. It looks like the same error but it's being triggered for a reason now.

What I think is happening now: I think the move away from the while/sleep loop in waitForBufferStatusLow https://github.com/livekit/client-sdk-js/pull/1877 (which wasn't in my data track subscription updates PR) has newly resulted in more `waitForBufferStatusLow` calls being in flight at once so there's now a much higher likelihood that the data channel won't be fully drained upon engine close. Before, the `sleep(10)` inadvertently throttled this.

To fix this, I've added a param to `waitForBufferStatusLow` to allow ignoring engine closed events. I am not sure if this is the best thing to do, but it fixes the tests. And I think matches the ethos of "send packets as fast as possible" / doesn't introduce any sort of explicit locking - if packets aren't delivered when the engine closes, there isn't really a good reason to throw an error if the data being sent is lossy anyway.

There's an argument also that maybe the better thing to do here is to make `tryPush` return a promise that resolves once the whole pipeline (including all packets being enqueued into the data channel, and then a subsequent buffer low event occurring) completes. But the downside to that is then you're blocking extra time on each `tryPush` call which lengthens the amount of time each frame takes to send, which is antithetical to the whole data track "send data as fast as possible" / lossy delivery notion.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -1519,7 +1519,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
         // buffer status to not be low before continuing.
         switch (bufferStatusLowBehavior) {
           case 'wait':
-            await this.waitForBufferStatusLow(kind);
+            await this.waitForBufferStatusLow(kind, 'ignore');
             break;
           case 'drop':
             // this.log.warn(`dropping lossy data channel message`, this.logContext);
@@ -1580,9 +1580,12 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
   };
 
-  async waitForBufferStatusLow(kind: DataChannelKind) {
+  async waitForBufferStatusLow(
+    kind: DataChannelKind,
+    engineCloseBehavior: 'throw' | 'ignore' = 'throw',
+  ) {
     return new TypedPromise<void, UnexpectedConnectionState>(async (resolve, reject) => {
-      if (this.isClosed) {
+      if (this.isClosed && engineCloseBehavior === 'throw') {
         reject(new UnexpectedConnectionState('engine closed'));
       }
       if (this.isBufferStatusLow(kind)) {
@@ -1593,7 +1596,9 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
           reject(new UnexpectedConnectionState(`DataChannel not found, kind: ${kind}`));
           return;
         }
-        this.bufferStatusLowClosingFuture.promise.catch((e) => reject(e));
+        if (engineCloseBehavior === 'throw') {
+          this.bufferStatusLowClosingFuture.promise.catch((e) => reject(e));
+        }
         dc.addEventListener('bufferedamountlow', () => resolve(), {
           once: true,
         });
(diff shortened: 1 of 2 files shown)


src/room/RTCEngine.ts
```diff
@@ -1519,7 +1519,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
         // buffer status to not be low before continuing.
         switch (bufferStatusLowBehavior) {
           case 'wait':
-            await this.waitForBufferStatusLow(kind);
+            await this.waitForBufferStatusLow(kind, 'ignore');
```
lukasIO: Instead of continuing here, wouldn't it make more sense to try/catch and early return on an error?

1egoman: Yes, I had that actually as an earlier implementation of this before switching over to adding the extra param to `waitForBufferStatusLow`, mostly just because it was a little bit cumbersome to catch one variant of one error. But I don't have a strong opinion, I could go back to doing it that way.

lukasIO: I would still expect dc.send further down to throw if the engine is closed, right? 
so either check for engine closed further down as well for an early return or we do it here with the try/catch

1egoman: It's a good point, but I can say that the code as written here resulted in the data tracks test passing locally when I ran it, and it looks like [on mdn at least, `RTCDataChannel.send` doesn't seem to mention throwing an error](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/send#exceptions) when the data channel is closed, only when `readyState` is `connecting`.

That being said, if we decide to go forward with this general approach, I'm fine with moving back over to the try/catch version.

lukasIO: hm, maybe that's an omission there, [here](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/readyState#closing) it states that sending isn't possible anymore after transitioning to `closing`, not a 100% sure whether it would throw an error though. 

Cool, then let's do the try/catch version

1egoman: `dc.send` might just silently drop them, I'm not sure. I can say the state of this branch as of writing this (0bb76f3) fixes the e2e tests for me so it definitely didn't throw an error in my tests.

Anyway, I moved back over to the try/catch version here: 6b35263

1egoman:
Things have shifted a bit since this was first opened, and I think generally we have decided to move in a slightly different direction. Instead of solving this problem on the sending end, we decided to allow the sender to explicitly "flush" a data track if they want to ensure that all data has been sent: https://github.com/livekit/client-sdk-js/pull/1925

Closing this in favor of this other change!

PR #1907, opened 2026-04-21
Improve logging foundation

lukasIO:


src/logger.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      room: 'foo',
+      participant: 'alice',
+    }));
+    log.debug('hello world', { extra: 1 });
+
+    const debugCalls = extension.mock.calls.filter((c) => c[0] === LogLevel.debug);
+    expect(debugCalls).toHaveLength(1);
+    const [level, msg, ctx] = debugCalls[0];
+    expect(level).toBe(LogLevel.debug);
+    expect(msg).toBe('[room=foo participant=alice] hello world');
```
1egoman: This is sweet! That being said, should there be a way to turn the prefix on or off? In particular, I'm thinking that if people were capturing these logs with a tool like datadog / ELK, they'd have the rich metadata and it would get annoying maybe to see it duplicated again in the log message.

src/logger.ts
```diff
@@ -37,20 +40,91 @@ export type StructuredLogger = log.Logger & {
...
+  'reconnectAttempt',
+  'region',
+] as const;
+
+/**
+ * Render the subset of `ctx` listed in `DISPLAY_KEYS` as a bracketed prefix
+ * suitable for prepending to a log message. Returns an empty string when
+ * no display keys are present. Pure function — safe to unit test directly.
+ */
+export function formatDisplayContext(ctx: object | undefined): string {
```
1egoman: suggestion: Could it be possible to make `ctx` something like `Partial<Record<keyof typeof DISPLAY_KEYS, unknown>>` (probably broken out into a type though and not inline, and also using this in `ContextProvider`?). If this were done everywhere that would make the typing information a bit more robust.

LOGGING.md
```diff
@@ -0,0 +1,61 @@
...
+  `connecting` / `connected` / `reconnecting (attempt N)` / `reconnected` /
+  `disconnected (reason)`, track `published` / `unpublished` /
+  `subscribed` / `unsubscribed`, permission changes, region switched,
+  e2ee enabled/disabled, signal (re)connected, major engine state
+  transitions. Nothing that can fire more than about once per second
+  under normal use.
+- **debug** — everything else: individual signal messages, per-ICE-
+  candidate, SDP, DTX/simulcast/codec negotiation, data channel
+  lifecycle, reconnection internal states, timing details.
+- **trace** — reserved for deliberate deep dives; unused by default.
```
1egoman, line 25: nitpick: It might be worth mentioning here what you told me in our 1:1 that `loglevel`'s trace logs include stack traces, and that's the main reason this level isn't being used more extensively.

1egoman (approved):

src/logger.ts
lukasIO: generally the suggestion makes sense to me, but expanding it to everywhere (which would be the only desired outcome) would also be breaking the type of the setLogExtension override that's exposed to users

src/logger.test.ts
lukasIO: Yeah, in a branch that builds on top of this one I also removed all display keys. 

I think while this is valuable in some cases, for most it's probably undesirable to do both the context and the display keys. So it should probably instead be a switch to decide either attach the object or convert all of the object keys into this format - but never do both. 

I removed it for now and we can revisit this

PR #1908, opened 2026-04-21
chore: replace explicit logContext passing with bound version, add info logs

lukasIO:


src/room/participant/LocalParticipant.ts
```diff
@@ -958,9 +941,6 @@ export default class LocalParticipant extends Participant {
     if (!isE2EESimulcastSupported() && this.roomOptions.e2ee) {
       this.log.info(
         `End-to-end encryption is set up, simulcast publishing will be disabled on Safari versions and iOS browsers running iOS < v17.2`,
-        {
-          ...this.logContext,
-        },
```
1egoman, line 963: question: Is there a good reason this shallow copy was being done here, and if so does a shallow copy need to be done in the `ContextProvider` like `() => ({ ...this.logContext })`?

src/room/RTCEngine.ts
```diff
@@ -227,6 +227,8 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private log = log;
 
+  private reconnectLog = log;
+
```
1egoman: question: Just in general, what do you have in mind with this pattern? Potentially creating `n` different -`Log` suffixed entries for each different type of log?

1egoman (approved):

src/room/participant/LocalParticipant.ts
lukasIO: the spreading here was originally only supposed to happen when other fields are added here as well, the shallow copy aspect isn't important here

src/room/RTCEngine.ts
lukasIO: the idea was to differentiate into more granular logger names. RtcEngine has a lot of responsiblities so things might get overwhelming if there was "only" a engine logger. 
But rereading this I'm hesitant to do it now and will remove this one for now. If we were to introduce it it might make sense to extract the reconnection logic out of rtcengine where possible and provide it its own logger there

PR #1911, opened 2026-04-21
Fix data tracks extension encoding

alan-george-lk:
Addresses an issue with the packet encoding where the extension block length is encoded differently than what the SFU expected, leading to packets not being forwarded in some cases (e.g., E2EE is enabled). The spec is also updated for clarification.

Fix for Rust: 
- https://github.com/livekit/rust-sdks/pull/1031


ladvoc (approved):
LGTM, tested manually between the Rust _wgpu_room_ example and JS _data_track_ example and verified frame delivery works as expected with E2EE enabled (previously enabling E2EE caused the issue to present itself).

PR #1916, opened 2026-04-24
Pass optional LocalTrack reference to processors

lukasIO:


1egoman (approved):
Makes sense to me.

One general question that came to mind looking at this: since the goal here is to allow downstream track processors to be able to tap into a local track's mute status, maybe it could be a better to instead expose something more abstract (like a mechanism to just subscribe to track muted / unmuted events)? Or maybe there's additional value in providing the added api surface area for future integrations - not sure.

PR #1917, opened 2026-04-24
Ignore data track promise rejections after a subscription readable stream is discarded

1egoman:
When `track.subscribe({ signal })` is called and the signal is aborted, `IncomingDataTrackManager.openSubscriptionStream` rejects two things with the same `DataTrackSubscribeError`:
1. The `ReadableStream` (catchable via `try`/`catch`)
2. An internal `sfuSubscriptionComplete` promise, used primarily in the tests, but also used during the brief window before the sfu subscription is established to communicate errors back to the user.

Because `RemoteDataTrack.subscribe()` returns only the stream and discards the promise, no handler was ever attached to the `sfuSubscriptionComplete` promise, which meant that any rejections become `unhandledRejection`s.

This pull request works around this by adding a no-op `.catch(() => {})` to this promise to silence any errors.

### DIFF
--- a/.changeset/salty-turkeys-stick.md
+++ b/.changeset/salty-turkeys-stick.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Ignore data track promise rejections after a subscription readable stream is discarded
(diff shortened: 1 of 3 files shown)


xianshijing-lk (approved):

lukasIO (approved):
makes sense to me

PR #1920, opened 2026-04-27
Ensure priority isn't set on all simulcast layers when using Firefox on iOS

lukasIO:
Priority can be set on all layers only when using Firefox. 
On iOS the underlying browser engine is still Webkit so we need to guard against wrongly treating it as Firefox.

### DIFF
--- a/src/room/participant/publishUtils.ts
+++ b/src/room/participant/publishUtils.ts
@@ -15,7 +15,6 @@ import type { LoggerOptions } from '../types';
 import {
   compareVersions,
   getReactNativeOs,
-  isFireFox,
   isReactNative,
   isSVCCodec,
   isSafariBased,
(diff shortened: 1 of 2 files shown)


src/room/participant/publishUtils.ts
```diff
@@ -15,7 +15,6 @@ import type { LoggerOptions } from '../types';
 import {
   compareVersions,
   getReactNativeOs,
-  isFireFox,
```
MaxHeimbrock, line 18: Just a followup, but do we need to treat Firefox on iOS differently in general? I mean in these other cases:
https://github.com/search?q=repo%3Alivekit%2Fclient-sdk-js%20isFireFox&type=code

MaxHeimbrock (approved):
LGTM

src/room/participant/publishUtils.ts
lukasIO: that's a good point. The other instances are more about "preventing an advanced feature for FF" which won't result in an actual issue, but certainly isn't the greatest experience. 

Worth investigating whether `isFirefox` should only ever return true on non-iOS versions.

PR #1923, opened 2026-04-27
Improved data tracks depacketizer to support mutliple in flight packets

1egoman:
- Updates the depacketizer to take a `maxPartialFrames` which controls how many frames are being stored at once.
- Exposes this as a `RemoteDataTrack` level setting - there is one pipeline per track, not one pipeline per subscription, so this is where it makes most sense (also given that you won't really need to configure it differently by subscription since this is fully dependant on the data being received from the sender).

Code example:
```tsx
const remoteDataTrack: RemoteDataTrack = /* ... */;

// The new bit:
remoteDataTrack.setPipelineOptions({ maxPartialFrames: 10 }); // Defaults to 1 to maintain backwards compatibility.

for await (const frame of remoteDataTrack.subscribe()) {
  // ...
}
```

## Todo
- [x] Add changeset

### DIFF
--- a/src/room/data-track/RemoteDataTrack.ts
+++ b/src/room/data-track/RemoteDataTrack.ts
@@ -80,4 +80,15 @@ export default class RemoteDataTrack implements IRemoteTrack, IDataTrack {
       throw err;
     }
   }
+
+  /** Set the maximum number of in-flight partial frames the depacketizer will track
+   * concurrently for this track. Higher values give more out-of-order tolerance for
+   * high-frequency senders.
+   *
+   * The value applies to all current and future subscribers of this track. May be called
+   * before or after {@link RemoteDataTrack.subscribe}; live subscriptions pick up the new value
+   * immediately. Defaults to 1. */
+  setMaxPartialFrames(n: number): void {
+    this.manager.setTrackMaxPartialFrames(this.info.sid, n);
+  }
 }
--- a/src/room/data-track/RemoteDataTrack.ts
+++ b/src/room/data-track/RemoteDataTrack.ts
@@ -21,6 +21,14 @@ export type DataTrackSubscribeOptions = {
   bufferSize?: number;
 };
 
+export type RemoteDataTrackPipelineOptions = {
+  /** Set the maximum number of in-flight partial frames the depacketizer will track
+   * concurrently for this track. Higher values give more out-of-order tolerance for
+   * high-frequency senders. Defaults to 1.
+   */
+  maxPartialFrames?: number;
+};
+
 export default class RemoteDataTrack implements IRemoteTrack, IDataTrack {
   readonly trackSymbol = TrackSymbol;
 
@@ -81,14 +89,10 @@ export default class RemoteDataTrack implements IRemoteTrack, IDataTrack {
     }
   }
 
-  /** Set the maximum number of in-flight partial frames the depacketizer will track
-   * concurrently for this track. Higher values give more out-of-order tolerance for
-   * high-frequency senders.
-   *
-   * The value applies to all current and future subscribers of this track. May be called
-   * before or after {@link RemoteDataTrack.subscribe}; live subscriptions pick up the new value
-   * immediately. Defaults to 1. */
-  setMaxPartialFrames(n: number): void {
-    this.manager.setTrackMaxPartialFrames(this.info.sid, n);
+  /** Configure how incoming frames for this track are processed before they are handed out to
+   * subscribers (the "pipeline"). These options apply to all current and future subscriptions
+   * of this track, and may be set at any time. */
+  setPipelineOptions(options: RemoteDataTrackPipelineOptions): void {
+    this.manager.setTrackMaxPartialFrames(this.info.sid, options.maxPartialFrames ?? null);
   }
 }
--- a/src/room/data-track/RemoteDataTrack.ts
+++ b/src/room/data-track/RemoteDataTrack.ts
@@ -7,7 +7,7 @@ import {
   type IRemoteTrack,
   TrackSymbol,
 } from './track-interfaces';
-import { type DataTrackInfo } from './types';
+import { type DataTrackInfo, type RemoteDataTrackPipelineOptions } from './types';
 
 type RemoteDataTrackOptions = {
   publisherIdentity: Participant['identity'];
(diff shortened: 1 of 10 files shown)


src/room/data-track/RemoteDataTrack.ts
```diff
@@ -80,4 +80,15 @@ export default class RemoteDataTrack implements IRemoteTrack, IDataTrack {
...
   }
+
+  /** Set the maximum number of in-flight partial frames the depacketizer will track
+   * concurrently for this track. Higher values give more out-of-order tolerance for
+   * high-frequency senders.
+   *
+   * The value applies to all current and future subscribers of this track. May be called
+   * before or after {@link RemoteDataTrack.subscribe}; live subscriptions pick up the new value
+   * immediately. Defaults to 1. */
+  setMaxPartialFrames(n: number): void {
```
ladvoc: question: What are your thoughts on adding this to `DataTrackSubscribeOptions` instead? This would still have the caveat that "the value applies to all current and future subscribers of this track," but this is the case for `bufferSize` anyway. Also keeps API surface cleaner.

src/room/data-track/depacketizer.test.ts
```diff
@@ -439,4 +439,436 @@ describe('DataTrackDepacketizer', () => {
       new Uint8Array([0x01, 0x02, 0x03]),
     );
   });
+
+  it('should assemble multiple partial frames concurrently when maxPartialFrames is set', () => {
```
ladvoc, line 443: praise: Nice test coverage!

ladvoc (approved):

ladvoc (approved):
Just the one question about exposing the new option, but looks good otherwise! I will port this to Rust so these can be merged at the same time.

src/room/data-track/RemoteDataTrack.ts
1egoman: I had brought up in our 1:1 that this needs to be configured at the track level, and not the subscription level, since there is one `IncomingDataTrackPipeline` per `RemoteDataTrack`. Given this, I'm curious what you have in mind here - are you suggesting this be changed so there's a`IncomingDataTrackPipeline` per subscription instead?

I will say that just generally, I am not really a big fan of this method approach, and would like to find a better place to put this, so if there's something I'm missing here I'm definitely open to changing this.

ladvoc: I'm proposing putting it on `DataTrackSubscribeOptions` but this updates the track-level value under-the-hood. One disadvantage I can think of is there is no way to change max partial frames once you have a subscription (without obtaining a second one)—but not sure if this is something you would want to do anyway.

1egoman: I'm confused - if it was done this way, wouldn't that mean that if a user did `.subscribe({ maxPartialFrames: 5 })` first, then `.subscribe({ maxPartialFrames: 10 })` second, the last value would "win" and apply to both subscriptions since there is one pipeline for all subscriptions of a given track? The only way to work around this would be one pipeline per subscription rather than one pipeline per track?

ladvoc: Yes, the latest value would win. This behavior would have to be documented at the option level similar to [this disclaimer](https://docs.rs/livekit-datatrack/latest/livekit_datatrack/api/struct.DataTrackSubscribeOptions.html#method.with_buffer_size) in Rust for buffer size.

1egoman: Summary of our slack conversation on this: Keep this at the track level, but instead of it being a dedicated field / method for `maxPartialFrames`, add a new concept `PipelineOptions` which can be set on a track via a method that contains `maxPartialFrames`:

```ts
remoteDataTrack.setPipelineOptions({ maxPartialFrames: /* ... */ });
```

src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -113,6 +118,19 @@ export default class IncomingDataTrackManager extends (EventEmitter as new () =>
...
+  /** @internal */
+  setPipelineOptions(sid: DataTrackSid, options: RemoteDataTrackPipelineOptions): void {
+    const descriptor = this.descriptors.get(sid);
+    if (!descriptor) {
+      log.warn(`Unknown track ${sid}, cannot set pipeline options.`);
+      return;
+    }
+    descriptor.pipelineOptions = options;
+    if (descriptor.subscription.type === 'active') {
+      descriptor.subscription.pipeline.setOptions(options);
```
xianshijing-lk, line 130: nit, wonder if that makes sense to have a set pipelineOptions() function in the descriptor class ? 
set pipelineOptions(options) {
    this._pipelineOptions = options;

    if (this.subscription?.type === 'active') {
      this.subscription.pipeline.setOptions(options);
    }
  }

src/room/data-track/depacketizer.ts
```diff
@@ -129,21 +140,30 @@ export default class DataTrackDepacketizer {
...
+    if (this.partials.size >= maxPartialFrames) {
+      const oldestPartialFrameNumber = this.peekOldestPartialFrameNumber();
+      if (typeof oldestPartialFrameNumber !== 'number') {
+        // @throws-transformer ignore - this should be treated as a "panic" and not be caught
+        throw new Error(
+          `Depacketizer.frameFromSingle: no oldest frame number found, but partials.size is ${this.partials.size}.`,
+        );
+      }
+      if (options?.throwOnInterruption) {
+        this.partials.delete(oldestPartialFrameNumber);
```
xianshijing-lk: curiously, why you only delete it if (options?.throwOnInterruption)

Or we should delete it regardless throwOnInterruption is true or false ?

xianshijing-lk (approved):
two comments, please address them.

src/room/data-track/incoming/IncomingDataTrackManager.ts
1egoman: `Descriptor` isn't a class, it's an regular js object: https://github.com/livekit/client-sdk-js/blob/f8d8323cb560efdf1d385a3a8b8318d0bef61b12/src/room/data-track/incoming/IncomingDataTrackManager.ts#L64-L68

So I'm not sure there's much that could be done to make this more abstract without larger structural changes.

src/room/data-track/depacketizer.ts
1egoman: I think you are right, in this case it should be being deleted in both the log and error throwing branches. Fixed in 58d126cd

PR #1925, opened 2026-04-28
Add local data track flush method

1egoman:
When called, waits for all in flight packets to be sent before resolving. Use this to:

1. Send frames exactly in order:
```ts
await localDataTrack.tryPush(/* ... */);
await localDataTrack.flush();

await localDataTrack.tryPush(/* ... */);
await localDataTrack.flush();

// ... etc ...
```

2. Wait for frames to all be delivered before unpublishing a local data track:

```ts
await localDataTrack.tryPush(/* ... */);
await localDataTrack.tryPush(/* ... */);
await localDataTrack.tryPush(/* ... */);
// ... etc ...
await localDataTrack.flush();
await localDataTrack.unpublish();
```

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -1751,6 +1751,8 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     this.bufferedEvents = [];
     this.transcriptionReceivedTimes.clear();
     this.incomingDataStreamManager.clearControllers();
+    this.incomingDataTrackManager.reset();
+    this.outgoingDataTrackManager.reset();
     if (this.state === ConnectionState.Disconnected) {
       return;
     }
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -293,7 +293,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       .on('packetAvailable', ({ handle, bytes }) => {
         this.engine
           .sendLossyBytes(bytes, DataChannelKind.DATA_TRACK_LOSSY, 'wait')
-          .then(() => this.outgoingDataTrackManager.handlePacketSendComplete(handle));
+          .finally(() => this.outgoingDataTrackManager.handlePacketSendComplete(handle));
       });
 
     this.disconnectLock = new Mutex();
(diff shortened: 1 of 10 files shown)


src/room/data-track/LocalDataTrack.ts
```diff
@@ -28,12 +29,20 @@ export default class LocalDataTrack implements ILocalTrack, IDataTrack {
 
   protected log: StructuredLogger = log;
 
+  /** Resolves once the data track has sent all pending packets the rtc data channel buffer. */
+  protected flushedFuture = new Future<void, never>();
```
lukasIO, line 33: what should happen to this future if the room disconnects while flushing?

src/room/Room.ts
```diff
@@ -290,8 +290,10 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       .on('trackUnpublished', (event) => {
         this.emit(RoomEvent.LocalDataTrackUnpublished, event.sid);
       })
-      .on('packetAvailable', ({ bytes }) => {
-        this.engine.sendLossyBytes(bytes, DataChannelKind.DATA_TRACK_LOSSY, 'wait');
+      .on('packetAvailable', ({ handle, bytes }) => {
+        this.engine
+          .sendLossyBytes(bytes, DataChannelKind.DATA_TRACK_LOSSY, 'wait')
```
lukasIO, line 295: if `sendLossyBytes` rejects should we still call the `handlePacketSendComplete` ? Otherwise the counter might get out of sync as there are no retries?

1egoman: Good point, I missed this. This should really be a `.finally`, not `.then`. Fixed in d7c20414

src/room/data-track/LocalDataTrack.ts
1egoman: Good point - I think the two options are rejecting with an error or just resolving it normally. Also I would argue that the better signal might be engine close, not room disconnect, though I think in practice the former should always happen when the latter occurs.

Any thoughts on either of these points before I add this?

lukasIO: engine close also sounds fine to me! 

no strong opinion on whether it should be rejecting or resolving, but _something_ should happen!

1egoman: Sweet, done in 8c3b58e and da85f11 (new test).

lukasIO (approved):

lukasIO:
nice! merging this so we can include it in the next release

PR #1926, opened 2026-04-28
Add super verbose data track logs for debugging

1egoman:
This, while verbose, would make debugging of data track related weirdness observed much easier.


1egoman:
I am interested in hearing from others if they think this is a good idea or not to add logging this verbose.

lukasIO:
how about releasing this verbose version as a snapshot release only for initial debugging?

1egoman:
I think I'm going to opt to close this for now and revisit in the future if we have additional debugging needs.

PR #1927, opened 2026-04-29
Implement negotiation tracking based on offerId

lukasIO:
## Fix publisher negotiation hang under stacked offer cycles
Under specific conditions the Promise returned by `setCameraEnabled(true)` could neither resolve nor reject, even though the track was successfully published from the server's perspective. Sessions hitting this also showed MaxListenersExceededWarning for negotiationComplete/negotiationStarted on the publisher within seconds of connecting.

### Root cause
`PCTransportManager.negotiate()` resolved on the next NegotiationComplete event from the publisher. That event only fires when an answer is processed and `renegotiate === false`. If a new `publisher.negotiate() `lands while an offer is in flight, the next answer recurses into another offer instead of emitting NegotiationComplete (PCTransport.ts:239-243). With enough cycle starts (data-channel sends, server-driven MediaSectionsRequirement, concurrent track publishes) the cycle never converges, the resettable timeout from #1813 keeps firing, and the Promise wedges. cleanup() also never offed the `once(NegotiationComplete)` listener, which is why the warning surfaced.

### Change
Switch from an event-cycle-based wait to an offerId checkpoint:

- Capture `checkpoint = publisher.latestOfferId `at the moment `negotiate()` is called.
- Resolve when any offer with offerId > checkpoint has had its answer applied — that offer necessarily contains the changes that motivated the call.
- New `PCEvents.OfferAnswered` fires for every successful answer, including ones that recurse via renegotiate=true (which the existing NegotiationComplete deliberately suppresses).
- Single non-resettable deadline as a backstop. No NegotiationStarted reset path.
- Single listener type, idempotent `cleanup()` — no leaks across abort/timeout/error paths.
- Concurrent callers each get their own checkpoint and resolve independently when their changes are negotiated.
- NegotiationComplete and RTPVideoPayloadTypes are kept intact for existing consumers.

### Tests
PCTransportManager.test.ts covers:

- Resolution on offer past checkpoint, including via the renegotiate-recursion path.
- Non-resolution on answers at or before the checkpoint.
- Concurrent callers resolving independently.
- Race where an answer past the checkpoint already arrived at entry.
- Abort, deadline, and publisher.negotiate error paths.
- Listener cleanup across all paths plus a 12-iteration leak guard.
- Regression test for the field hang: spurious cycle noise + non-matching answers no longer wedge the promise.

### Risk
Behavior change is contained to the publisher negotiation handshake. `latestOfferId` is exposed publicly (was private) and latestAcknowledgedOfferId is added — both internal SDK fields. No public API surface changes.

### DIFF
--- a/src/room/PCTransportManager.test.ts
+++ b/src/room/PCTransportManager.test.ts
@@ -38,7 +38,23 @@ class StubPC {
 }
 
 class FakePublisher extends EventEmitter {
+  latestOfferId = 0;
+
+  latestAcknowledgedOfferId = 0;
+
   negotiate = vi.fn(async (_onError?: (e: Error) => void) => {});
+
+  /** Simulate a publisher offer cycle: bump latestOfferId. */
+  startOffer() {
+    this.latestOfferId += 1;
+    return this.latestOfferId;
+  }
+
+  /** Simulate a successful answer for the given offerId. */
+  answer(offerId: number) {
+    this.latestAcknowledgedOfferId = offerId;
+    this.emit(PCEvents.OfferAnswered, offerId);
+  }
 }
 
 describe('PCTransportManager.negotiate', () => {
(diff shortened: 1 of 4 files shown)


src/room/PCTransportManager.test.ts
```diff
@@ -0,0 +1,282 @@
...
+    // answered. We resolve on the second answer.
+    const { manager, pub } = makeManager();
+    pub.latestOfferId = 1;
+    const p = manager.negotiate(new AbortController());
+
+    setTimeout(() => pub.answer(1), 10); // does not satisfy checkpoint=1
+    setTimeout(() => {
+      const id = pub.startOffer(); // 2
+      pub.answer(id);
+    }, 30);
```
1egoman: thought: Two things:
- I wonder if there's a way to do a lot of these same testing scenarios without explicit sleeps. Are the lengths of these delays meaningful here, or is the goal just to have a series of events all occur in a defined order all on their own event loop macrotasks?
- I think linearizing these would make the tests a bit easier to follow, so you could scan the test top to bottom and have to jump between different functions which get scheduled to run at different times. ie, something like:
```ts
// Some sort of explanation here
// does not satisfy checkpoint=1
pub.answer(1);

await sleep(20); // Or ideally if the delay isn't important, `sleep(0)` / `setImmediate`.

// Some sort of explanation here
const id = pub.startOffer(); // 2
pub.answer(id);
```

src/room/PCTransportManager.ts
```diff
@@ -227,47 +228,56 @@ export class PCTransportManager {
...
+      // applied — that offer is the one that includes any transceiver/SDP
+      // changes that motivated this negotiate call. Concurrent callers each
+      // get their own checkpoint and resolve independently.
+      const checkpoint = this.publisher.latestOfferId;
+
+      // Race: an answer past our checkpoint already arrived before we had a
+      // chance to subscribe.
+      if (this.publisher.latestAcknowledgedOfferId > checkpoint) {
+        resolve();
+        return;
```
1egoman, line 250: suggestion: Maybe it's worth logging here so that it would be evident that there's a lot of spurious renegotiation occurring in the context of other issues? Not sure.

src/room/PCTransportManager.test.ts
```diff
@@ -0,0 +1,282 @@
...
+  answer(offerId: number) {
+    this.latestAcknowledgedOfferId = offerId;
+    this.emit(PCEvents.OfferAnswered, offerId);
+  }
+}
+
+describe('PCTransportManager.negotiate', () => {
+  let originalRTCPeerConnection: unknown;
+
+  beforeEach(() => {
```
1egoman, line 70: Really glad to see some tests here for this! Looking forward to seeing this grow moving forward.

1egoman (approved):
Generally the whole approach makes sense to me!

src/room/PCTransportManager.test.ts
lukasIO: yeah, that probably makes things clearer, will change it to this pattern

PR #1928, opened 2026-04-29
Enable packet trailer on tracks published by JS

chenosaurus:
- add Packet Trailers to the videoTracks published by the JS sdk.
- update demo app to have configurable options for packet trailers to attach to video frames.


src/api/SignalClient.ts
```diff
@@ -1184,6 +1184,12 @@ function createConnectionParams(
   if (info.browserVersion) {
     params.set('browser_version', info.browserVersion);
   }
+  if (info.capabilities.length > 0) {
```
lukasIO: I don't think there's a scenario where we have a LiveKit server that understands the capabilities without the `v1` connection path, so this shouldn't be needed?

src/room/participant/LocalParticipant.ts
```diff
@@ -1369,6 +1383,8 @@ export default class LocalParticipant extends Participant {
       await negotiate();
     }
 
+    this.reconcilePublishedPacketTrailerOptions(opts, ti);
```
lukasIO: why is this being called both here and in the `else` clause above?

src/room/participant/LocalParticipant.ts
```diff
@@ -1567,8 +1621,9 @@ export default class LocalParticipant extends Participant {
       await this.engine.negotiate();
     };
 
-    const rets = await Promise.all([this.engine.addTrack(req), negotiate()]);
-    const ti = rets[0];
+    const ti = await this.engine.addTrack(req);
+    this.reconcilePublishedPacketTrailerOptions(opts, ti);
+    await negotiate();
```
lukasIO: sequentialising this likely has a negative impact on the performance/speed, and is now being done for all cases even if there's no packet trailer involved at all

src/room/Room.test.ts
```diff
@@ -28,3 +30,62 @@ describe('Active device switch', () => {
     expect(kind).toBe('audioinput');
   });
 });
+
+describe('Room signaling options', () => {
+  it('advertises packet trailer capability when E2EE can handle trailers', async () => {
```
1egoman, line 35: thought: I think my biggest concern with this change at this point is still the breakdown of the packet trailer worker / e2ee worker. I think ideally, there could be one worker which could do both so that there does't have to be a separate packet trailer code paths in two places. You've abstracted this fairly well so I think if we _had_ to do this, barring a larger refactor where "steps" are registered (which you said was too intense for this feature) this is probably is about the cleanest it could be.

But if you haven't considered it, I wonder if it would be worth renaming the e2ee worker to be something more generic like "frame processing worker" (while still exporting it under the same `/e2ee-worker` path to maintain backwards compatibility, maybe with a deprecation) and then moving all docs / future usage to instead something like `/frame-processing-worker`.

To be clear, this isn't a blocking ask. More just we have an opportunity to pick how these frame processing worker external interfaces work and I want to make sure a conscious choice is being made here that won't inadvertently paint us into a corner later.

1egoman (commented):
Amazing progress on this generally - other than Lukas's comments I didn't see anything major which popped out at me.

src/room/Room.test.ts
chenosaurus: @1egoman agree with the approach of moving the e2ee-worker to something more generic like a `frame-transformer-worker` that can chain the processes together.    I've added a ticket to do this work in the near future.

src/api/SignalClient.ts
chenosaurus: ur right, removed this

src/room/participant/LocalParticipant.ts
chenosaurus: reverting this as we don't need to reconcile published track's packet trailer feature w/ what the server returns in TrackInfo.

src/packetTrailer/packetTrailer.ts
```diff
@@ -61,6 +61,35 @@ export function appendPacketTrailer(
   return result;
 }
 
+export function hasPacketTrailerPublishOptions(options?: PacketTrailerPublishOptions): boolean {
```
lukasIO: nit: this is used in other files as well, should arguably live in utils instead

src/e2ee/E2eeManager.ts
```diff
@@ -588,7 +593,12 @@ export class E2EEManager
    * a frame encoder.
    *
    */
-  private handleSender(sender: RTCRtpSender, trackId: string, codec?: VideoCodec) {
+  private handleSender(
+    sender: RTCRtpSender,
+    trackId: string,
+    codec?: VideoCodec,
+    packetTrailer?: TrackPublishOptions['packetTrailer'],
```
lukasIO, line 601: issue: the `packetTrailer` here needs to propagate down to the - currently hardcoded - `hasPacketTrailer: false`
 field on the worker messages. There are two instances where this needs to change to something like `hasPacketTrailer: !!packetTrailer`

lukasIO (approved):
barring the last two things in the comments below, this looks good to me!

PR #1929, opened 2026-04-30
Ensure that flush called when there are no in flight data track packets works

1egoman:
Previously, if flush was called on a freshly created `LocalDataTrack` or on a `LocalDataTrack` which had properly sent all of its packets, it would return a promise which if no more data were sent on the local data track, would never resolve.

Fixes this issue and also adds a few more tests around this and some other flush related edge cases.

Follow up to https://github.com/livekit/client-sdk-js/pull/1925

### DIFF
--- a/.changeset/cool-lands-smile.md
+++ b/.changeset/cool-lands-smile.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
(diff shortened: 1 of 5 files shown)


lukasIO (approved):
lgtm!

PR #1946, opened 2026-05-20
add smoke tests for package consumption

lukasIO:


1egoman (approved):
Fantastic, this will be really useful!

PR #1954, opened 2026-05-29
Disable publisher offer with join in firefox

cnderrauber:
To address the connectivity issue that we can't reproduced in firefox.


lukasIO (approved):
thanks!

PR #1961, opened 2026-06-03
fix: rtpMap event listener leak on multiple negotiations without video tracks

lukasIO:
addresses https://community.livekit.io/t/rtpvideopayloadtypes-listener-leak-on-publisher-pctransport-during-audio-only-renegotiations/1320

Additionally hardens codec detection in e2ee path by inspecting frame metadata's `mimeType` field if available without relying on codec mapping.

### DIFF
--- a/src/e2ee/worker/FrameCryptor.ts
+++ b/src/e2ee/worker/FrameCryptor.ts
@@ -8,7 +8,8 @@ import {
 } from '../../packetTrailer/packetTrailer';
 import type { PacketTrailerPublishOptions } from '../../packetTrailer/types';
 import { hasPacketTrailerPublishOptions } from '../../packetTrailer/utils';
-import type { VideoCodec } from '../../room/track/options';
+import { type VideoCodec, videoCodecs } from '../../room/track/options';
+import { mimeTypeToVideoCodecString } from '../../room/track/utils';
 import { ENCRYPTION_ALGORITHM, IV_LENGTH, UNENCRYPTED_BYTES } from '../constants';
 import { CryptorError, CryptorErrorReason } from '../errors';
 import { type CryptorCallbacks, CryptorEvent } from '../events';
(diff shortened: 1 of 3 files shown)


src/room/RTCEngine.ts
```diff
@@ -1746,20 +1746,8 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-          rtpTypes.forEach((rtp) => {
-            const codec = rtp.codec.toLowerCase();
-            if (isVideoCodec(codec)) {
-              rtpMap.set(rtp.payload, codec);
-            }
-          });
-          this.emit(EngineEvent.RTPVideoMapUpdate, rtpMap);
-        },
-      );
+      this.pcManager.publisher.off(PCEvents.RTPVideoPayloadTypes, this.onRtpMapAvailable);
```
1egoman, line 1749: question: What's the thinking behind adding this here and not adding this in the `finally`? Do you want to keep at least one `RTPVideoPayloadTypes` listener in flight always?

src/e2ee/worker/FrameCryptor.ts
```diff
@@ -795,36 +802,51 @@ export class FrameCryptor extends BaseFrameCryptor {
...
-        ...this.logContext,
-      });
+      if (!this.loggedNALUFallbacks.has(fallbackKey)) {
+        this.loggedNALUFallbacks.add(fallbackKey);
+        workerLogger.warn('NALU processing failed, falling back to VP8 handling', {
+          error: e,
+          payloadType,
+          ...this.logContext,
+        });
+      }
```
1egoman: nitpick: It might be a little cleaner to move all this into a method which handles the set appending and logging if relevant.

1egoman (approved):

src/room/RTCEngine.ts
lukasIO: yeah, just a defensive measure to ensure that _if_ there are cases when this fired unrelated to this specific `negotiate` call it would keep working as it used to. 

I think you're right that it would be cleaner to unregister in the `finally` block though. 

If it's nothing you feel strongly about I'd keep it as is for now to ensure we're not missing an edge case where this would fire at a different time.

src/e2ee/worker/FrameCryptor.ts
lukasIO: done in https://github.com/livekit/client-sdk-js/pull/1961/commits/5d4950eeb0dc0f254d49a155eb792065b7caec78

src/room/RTCEngine.ts
1egoman: Nope I don't have a strong opinion, sounds good to keep it as is.

PR #1962, opened 2026-06-05
Add 64k max data packet size

1egoman:
Ports this corresponding android sdk pull request to the web sdk: https://github.com/livekit/client-sdk-android/pull/948

PR #1966, opened 2026-06-08
Properly handle max payload length of zero

1egoman:
As a follow up to https://github.com/livekit/client-sdk-js/pull/1962, I realized there were a few small discrepancies with the [swift](https://github.com/livekit/client-sdk-swift/pull/1031) and [rust](https://github.com/livekit/rust-sdks/pull/1137) versions.

- Both have a `min(maxMessageSize, 64_000)` check rather than using the `maxMessageSize` value verbatim.
- Both (per RFC 8841) assume that `0` is a special value that means "no limit", the web implementation didn't do this
- Both log if the data channel gets closed before the connection is gracefully cleaned up.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -127,6 +127,11 @@ export enum DataChannelKind {
   DATA_TRACK_LOSSY = 2,
 }
 
+// Default data-channel max message size (bytes), used when the remote SDP
+// answer does not advertise an `a=max-message-size` attribute (RFC 8841).
+// `0` means "no limit".
+const DEFAULT_MAX_MESSAGE_SIZE = 64_000;
+
 /** @internal */
 export default class RTCEngine extends (EventEmitter as new () => TypedEventEmitter<EngineEventCallbacks>) {
   client: SignalClient;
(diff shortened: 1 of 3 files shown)


lukasIO (approved):

src/room/RTCEngine.ts
```diff
@@ -1537,9 +1562,16 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
     const msg = packet.toBinary() as Uint8Array<ArrayBuffer>;
 
-    const maxPublisherMessageSizeBytes = this.pcManager?.getMaxPublisherMessageSize();
+    // Clamp to the SDK default - libwebrtc advertises larger (~256 KiB)
+    // than LiveKit/pion can deliver end-to-end (~64 KiB), so we trust
+    // the answer up untilthe built in ceiling.
+    const maxPublisherMessageSizeBytes = Math.min(
+      this.pcManager?.getMaxPublisherMessageSize() ?? DEFAULT_MAX_MESSAGE_SIZE,
```
davidliu, line 1569: Realized a bit late that the react-native-webrt. didn't fill out the sctp field so the previous PR didn't actually work for it. Adding the default now covers react-native nicely 👍

davidliu (approved):

src/room/RTCEngine.ts
```diff
@@ -1537,9 +1562,16 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+      DEFAULT_MAX_MESSAGE_SIZE,
+    );
     if (
       typeof maxPublisherMessageSizeBytes !== 'undefined' &&
+      maxPublisherMessageSizeBytes !== 0 /* 0 means "no limit" */ &&
```
xianshijing-lk, line 1574: can maxPublisherMessageSizeBytes be 0 at all ? I think it is protected to be min(DEFAULT_MAX_MESSAGE_SIZE)

1egoman: Yep, it can be zero - the expression `0 ?? n` will always be zero ([more info](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing), this is actually an important key difference between `??` and `||`), so `min(0 ?? n, n)` where `n` > 0 will also always be `0`.

PR #1971, opened 2026-06-10
Add deployment to TokenSourceFetch options

lukasIO:


1egoman (approved):

1egoman:
I am a bit confused by the typescript failure in ci: `Error: src/room/token-source/utils.ts(67,15): error TS2322: Type '"deployment"' is not assignable to type 'never'.` - is this using the right `@livekit/protocol` version? That line number makes no sense for this error though...

1egoman:
You may or may not be aware but for this to propagate to `useSession` in components-js properly, there's a bit of [technical debt that needs to be cleaned up](https://github.com/livekit/components-js/blob/c51167da5ea8b255cdb801a8415a3f3a83a89683/packages/react/src/hooks/useSession.ts#L232-L264) which I keep forgetting to address.

src/room/token-source/types.ts
```diff
@@ -63,6 +63,7 @@ export type TokenSourceFetchOptions = {
 
   agentName?: string;
   agentMetadata?: string;
+  deployment?: string;
 };
```
1egoman, line 67: thought: Does this `CreateDispatchOptions` data need to be included in here as well?

https://github.com/livekit/node-sdks/blob/80b5d172f4ba0813bebebee42857a41163fec8f2/packages/livekit-server-sdk/src/AgentDispatchClient.ts#L16-L27

lukasIO:
> I am a bit confused by the typescript failure in ci: Error: src/room/token-source/utils.ts(67,15): error TS2322: Type '"deployment"' is not assignable to type 'never'. - is this using the right @livekit/protocol version? That line number makes no sense for this error though...

there were two equality checks (one in TokenSource and one in utils) and I forgot about the one in utils. 
I've now simply replaced the other one with the one from utils.

src/room/token-source/types.ts
lukasIO: the equivalent to `metadata` there is the `agentMetadata` we set explicitly here as well, I believe?

I don't think we have another option to set this as part of RoomConfig ?

1egoman: 👍 on `metadata`, yes good point.

Does `restartPolicy` need to be able to be set in here too? As of now, in the node version, I think it is settable so assuming it isn't server specific, it maybe needs to be able to be configured here as well?

lukasIO: let's ship this as a focused update around the `deployment` field and we can follow up with `restartPolicy` any time

PR #1974, opened 2026-06-11
Add `restrictOwnAudio` experimental param to `AudioCaptureOptions`

1egoman:
Works around an issue reported by a customer.

Note that this `restrictOwnAudio` param is only supported by chrome / chrome-derived browsers. I think the better long term move rather than hardcoding individual settings in here would be to somebow derive `AudioCaptureOptions` from `MediaTrackSettings` ([link](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings)) so that any globally configured values per the user's `tsconfig.json` `lib` setting would automatically cascade through this api.

### DIFF
--- a/.changeset/eighty-pears-cheat.md
+++ b/.changeset/eighty-pears-cheat.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Add restrictOwnAudio experimental param to AudioCaptureOptions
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
lgtm. 

One caveat for your long term suggestion is that chrome usually moves quicker than the spec/TS update with new features.

PR #1978, opened 2026-06-16
Fix data channel close race condition

### LINKED ISSUE #1953: Unknown DataChannel error logged as console.error for graceful close events (audio-only rooms)
#### What I'm expecting
Joining and leaving an audio-only room (`<LiveKitRoom video={false} audio={true} ...>`) should not log `console.error` lines when no actual error has occurred.

#### What happens instead
On every room leave / route change, `Unknown DataChannel error on lossy` and/or `Unknown DataChannel error on reliable` is logged at `error` severity. The underlying browser event is a generic `Event` (not `ErrorEvent`) emitted during graceful close / ICE restart cleanup — `event.message` and `event.error` are `undefined`. There is no actual failure; audio works correctly. But the red `console.error` lines:

- Get flagged by error-tracking tools (Sentry / Datadog RUM) as production errors
- Confuse developers reviewing console output
- Have no diagnostic value (payload `{ event }` contains no usable information)

#### Code location

`src/room/RTCEngine.ts:996-1006`:

```ts
private handleDataError = (event: Event) => {
  const channel = event.currentTarget as RTCDataChannel;
  const channelKind = channel.maxRetransmits === 0 ? 'lossy' : 'reliable';

  if (event instanceof ErrorEvent && event.error) {
    const { error } = event.error;
    this.log.error(`DataChannel error on ${channelKind}: ${event.message}`, { error });
  } else {
    this.log.error(`Unknown DataChannel error on ${channelKind}`, { event });  // ← issue
  }
};
```

The `else` branch fires when `event` is NOT an `ErrorEvent` with `event.error` populated — i.e. there is no actual error data, yet it is logged at `error` severity.

#### Proposed fix

Downgrade `else` branch from `log.error` to `log.debug` (one-line diff):

```diff
   } else {
-    this.log.error(`Unknown DataChannel error on ${channelKind}`, { event });
+    this.log.debug(`Unknown DataChannel error on ${channelKind}`, { event });
   }
```

Rationale:
- Real errors (with `event.error` populated) still take the `if` branch — unaffected
- `log.debug` is invisible by default but available via `setLogLevel('debug')` for users actively debugging data channels
- Zero behavioral change — only severity adjustment of an informational log

### Reproduction

1. Create a minimal audio-only room with `livekit-client` 2.19.0:

```ts
import { Room } from 'livekit-client';

const room = new Room();
await room.connect(serverUrl, token);
await room.localParticipant.setMicrophoneEnabled(true);
// ... use room for a bit ...
await room.disconnect();
```

Or, equivalently, using `@livekit/components-react`:

```tsx
<LiveKitRoom token={token} serverUrl={url} video={false} audio={true} connect>
  <RoomAudioRenderer />
</LiveKitRoom>
```

2. Trigger a leave (`room.disconnect()`, route change, or component unmount).
3. Observe DevTools Console — red `Unknown DataChannel error on lossy` / `Unknown DataChannel error on reliable` lines appear.

Reproduces consistently across Chrome 130 / Firefox 132 / Safari 18 (tested) on every leave.

### Logs

```shell
Unknown DataChannel error on lossy
  at HTMLDocument.<anonymous> (livekit-client.esm.mjs:21183)
  ...
Unknown DataChannel error on reliable
  at HTMLDocument.<anonymous> (livekit-client.esm.mjs:21183)
  ...
```

Payload object (when inspected): `{ event: Event { isTrusted: true, type: 'error', target: RTCDataChannel } }` — no `message`, no `error`, no `stack` beyond the SDK's own log call site.

### System Info

```shell
  System:
    OS: macOS 12.6.3
    CPU: (4) x64 Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
    Memory: 40.92 MB / 4.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 24.15.0 - /usr/local/bin/node
    npm: 11.12.1 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 110.1.48.171
    Safari: 17.6
  npmPackages:
    @livekit/components-react: ^2.9.21 => 2.9.21
    livekit-client: ^2.19.0 => 2.19.0
    livekit-server-sdk: ^2.15.3 => 2.15.3
```

Also reproduced on Chrome 130 and Firefox 132 (separate test machine).

### Severity

annoyance

#### Workarounds attempted (none satisfactory)

1. **`setLogExtension`** (Sentry-style log forwarding) — only forwards logs downstream, does not filter or suppress upstream `console.error`.
2. **`setLogLevel('warn')`** on the engine logger — silences *all* errors, too broad.
3. **`RoomOptions.logger` / custom logger injection** — not in stable public API.
4. **Avoid creating data channels** when `video: false` — `RTCEngine.createDataChannels()` (line 827) is called unconditionally regardless of published media tracks.
5. **Override `console.error` globally** — works but global side effect, fragile, breaks other error tooling.
6. **`patch-package` with the proposed one-line diff** — current workaround in our project (Next.js 16 + React 19, audio-only tutoring app); applied at build time via `postinstall`.

#### Willing to submit PR

Yes — happy to open a PR with the one-line diff above if maintainers agree on the approach. Open to alternative approaches (e.g. add an option to silence specific log categories) if preferred.

1egoman:
Fixes #1953 (I think, still needs confirmation from the reporter - I was unable to reproduce locally).

In `main`, there is a possible race condition that could happen when disconnecting a room. The below three tasks run in this order:
1. Close peer connection
2. Close data channel
3. Remove handlers from data channel

What could happen is that when 1 would run, it would close the data channels pre-emptively and lead to some of the handlers (3) firing before they were cleaned up, leading to a `Unknown DataChannel error on ...` error being logged.

So to fix this, reorder the above to now run in 3 -> 2 -> 1 order. This means that by the time any data channel errors that occur during disconnect are fired (at 1), the handlers are already detached (3).

--------

As an unrelated fix, also, `Unknown DataChannel error on ...` shouldn't have been logged in this case. It looks like this was because of the `event instanceof ErrorEvent` check - actually, `event` would be a `RTCErrorEvent` and confusingly given their names, `RTCErrorEvent` does not extend `ErrorEvent`. So, alter this code to check the right error which should make these error paths show more detail.

### DIFF
--- a/.changeset/gold-candies-switch.md
+++ b/.changeset/gold-candies-switch.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Fix data channel close race condition
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -456,33 +456,24 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
   }
 
   async cleanupPeerConnections() {
-    // Detach the data channel handlers before closing the peer connections. Closing a peer
-    // connection tears down the SCTP transport, which can dispatch `error`/`close` events on
-    // the still-open data channels; if our handlers are still attached at that point,
-    // handleDataError logs a spurious "Unknown DataChannel error" during an otherwise graceful
-    // disconnect. Detaching first makes this deterministic regardless of how/when the browser
-    // dispatches those teardown events. See livekit/client-sdk-js#1953.
-    const detachHandlers = (dc: RTCDataChannel | undefined) => {
-      if (!dc) return;
+    const dcCleanup = (dc: RTCDataChannel | undefined) => {
+      if (!dc) {
+        return;
+      }
+
+      // Detach the data channel handlers before closing anything. Closing a peer connection tears
+      // down the SCTP transport, which can dispatch `error`/`close` events on the still-open data
+      // channels; if our handlers are still attached at that point, handleDataError logs a spurious
+      // "Unknown DataChannel error" during an otherwise graceful disconnect. Removing the handlers
+      // before dc.close()/pcManager.close() makes this deterministic regardless of how/when the
+      // browser dispatches those teardown events. See livekit/client-sdk-js#1953.
       dc.onbufferedamountlow = null;
       dc.onclose = null;
       dc.onclosing = null;
       dc.onerror = null;
       dc.onmessage = null;
       dc.onopen = null;
-    };
-    detachHandlers(this.lossyDC);
-    detachHandlers(this.lossyDCSub);
-    detachHandlers(this.reliableDC);
-    detachHandlers(this.reliableDCSub);
-    detachHandlers(this.dataTrackDC);
-    detachHandlers(this.dataTrackDCSub);
 
-    await this.pcManager?.close();
-    this.pcManager = undefined;
-
-    const dcCleanup = (dc: RTCDataChannel | undefined) => {
-      if (!dc) return;
       dc.close();
     };
     dcCleanup(this.lossyDC);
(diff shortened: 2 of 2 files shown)


lukasIO (approved):
makes sense!

PR #1979, opened 2026-06-18
Ensure that sortPresets does NOT mutate values passed in via options into Room

### LINKED ISSUE #1345: Unexpected room reconnect when publishing video with custom simulcast layers
### Select which package(s) are affected

\@livekit/components-react

### Describe the bug

When `videoSimulcastLayers` or `screenShareSimulcastLayers` are passed in options to `<LiveKitRoom />` (or `useLiveKitRoom`), publishing a video track can trigger a full room teardown and reconnect.

The cause is a combination of two things:

1. `useLiveKitRoom` recreates the `Room` instance whenever `JSON.stringify(options)` changes between renders.
2. The livekit-client SDK's internal `sortPresets` function sorts simulcast layer arrays in place when a track is published, mutating the original `options` object.
 
After the first publish, the simulcast layers array inside `options` has been reordered. On the next render, `JSON.stringify(options)` produces a different string, the effect fires, a new `Room` is created, and the old one is disconnected.

### Reproduction

Pass videoSimulcastLayers in an order that isn't already sorted by maxBitrate ascending, then publish a camera track. You'll see the room disconnect and reconnect.

### Logs

```shell

```

### System Info

```shell
System:
    OS: macOS 26.5.1
    CPU: (10) arm64 Apple M4
    Memory: 116.52 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.16.0 - /Users/jk/.nvm/versions/node/v24.16.0/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 11.13.0 - /Users/jk/.nvm/versions/node/v24.16.0/bin/npm
    pnpm: 10.28.1 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 149.0.7827.115
    Firefox: 152.0
    Safari: 26.5
  npmPackages:
    @livekit/components-core: ^0.12.13 => 0.12.13 
    @livekit/components-react: ^2.9.21 => 2.9.21 
    @livekit/components-styles: ^1.2.0 => 1.2.0 
    livekit-client: ^2.19.1 => 2.19.1
```

### Severity

annoyance

### Additional Information

_No response_

1egoman:
What could happen previously is if the user passed in the layers into `LiveKitRoom` in `components-js` in the wrong order, then the `JSON.stringify(...)` comparison [here](https://github.com/livekit/components-js/blob/97bf9dcc603b50530059904cdb4e34d5825e6baf/packages/react/src/hooks/useLiveKitRoom.ts#L64) could be attempting to compare the sorted (via side effect) version with the new value the user passed in (which has a different ref and would not be sorted), always recreating the room on every render.

Fixes https://github.com/livekit/components-js/issues/1345

### DIFF
--- a/.changeset/tired-views-turn.md
+++ b/.changeset/tired-views-turn.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Ensure that sortPresets does NOT mutate values passed in via options into Room
--- a/src/room/participant/publishUtils.test.ts
+++ b/src/room/participant/publishUtils.test.ts
@@ -174,6 +174,12 @@ describe('customSimulcastLayers', () => {
     expect(sortedPresets[1].encoding.maxFramerate).toBe(15);
     expect(sortedPresets[2].encoding.maxFramerate).toBe(20);
   });
+  it('does not mutate the passed-in array', () => {
+    const presets = [VideoPresets.h1440, VideoPresets.h360, VideoPresets.h1080, VideoPresets.h90];
+    const original = [...presets];
+    sortPresets(presets);
+    expect(presets).toEqual(original);
+  });
 });
 
 describe('screenShareSimulcastDefaults', () => {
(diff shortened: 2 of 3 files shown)


lukasIO (approved):
nice!

PR #1982, opened 2026-06-22
Rename PacketTrailer to FrameMetadata

lukasIO:


src/room/PCTransport.ts
```diff
@@ -544,6 +544,7 @@ export default class PCTransport extends (EventEmitter as new () => TypedEmitter
       return;
     }
     this.pendingInitialOffer = undefined;
+    this.log.warn(`closing peer connection`);
     this._pc.close();
```
1egoman, line 547: nitpick: Was this intended to be added in? I think it makes sense but just wanted to make sure as it is unrelated to the broader renaming change.

package.json
```diff
@@ -17,9 +17,14 @@
       "require": "./dist/livekit-client.e2ee.worker.js"
     },
     "./packet-trailer-worker": {
-      "types": "./dist/src/packetTrailer/worker/packetTrailer.worker.d.ts",
-      "import": "./dist/livekit-client.pt.worker.mjs",
-      "require": "./dist/livekit-client.pt.worker.js"
+      "types": "./dist/src/frameMetadata/worker/frameMetadata.worker.d.ts",
+      "import": "./dist/livekit-client.fm.worker.mjs",
+      "require": "./dist/livekit-client.fm.worker.js"
+    },
```
1egoman, line 23: thought: I wonder if there's anything you can do to deprecate this old import path... 🤔

1egoman (approved):

src/room/PCTransport.ts
lukasIO: no, leftover from debugging a different issue, thanks!

```suggestion
    }
    this.pendingInitialOffer = undefined;
    this._pc.close();
```

chenosaurus (approved):
LGTM, TY!

package.json
lukasIO: it's an interesting thought, let me know if you can think of a way to deprecate it!

PR #1983, opened 2026-06-22
Add support for user data frame metadata trailer type

chenosaurus:
- Add frame metadata parser for new user_data trailer type
- Update examples/demo to render parsed user_data


examples/demo/demo.ts
```diff
@@ -347,6 +347,13 @@ const appActions = {
                   `\nReceive:  ${fmt(receiveTime)}` +
                   `\nLatency:  ${latencyDisplay}`;
               }
+              if (meta.userData && meta.userData.byteLength > 0) {
```
chenosaurus, line 350: this just renders some info on the demo app overlay when user_data is detected.

lukasIO (approved):
lgtm

PR #1985, opened 2026-06-23
Data streams v2

1egoman:
> [!WARNING]
> This change relies on some new SFU updates to be able to function. Checkout `main` (really, https://github.com/livekit/livekit/commit/86a79f83fc03706054c0faae65e93fc8c7e9d7ea or later), build + run the SFU locally, and point any test app at this local SFU for testing.

## Overview

Data streams v2 consists of some updates to data streams to add three things:
1. Single packet data streams for short (< 15kb), finite length payloads
2. DEFLATE compression, when both the sender and all receivers support compression/decompression
3. A maximum 15kb size for data stream attributes to close this as a DOS vector

This pull request contains the initial web implementation and initial data streams spec which will be used to propagate this across all other client sdks.

## Performance

Both 1 and 2 together at least **doubles** ⚠️ data stream throughput across the spectrum of network conenctions and payload sizes. 

Before:
image

After:
image

## TODO
- [x] Remove benchmark and spec before merging
- [x] Maybe make the changeset a minor release, not a patch? This is a fairly substantial change to a core sdk feature.
- [x] Cleanup cdc741a0, I think I might need to rebase this


examples/data-stream-benchmark/api.ts
1egoman, line 1: NOTE: The diff is quite large, and a big reason why it is is because I have both this example app AND the spec checked in. I'll remove these before an eventual merge which will make the diff a lot more reasonable.

Feel free to ignore this `examples/data-stream-benchmark` directory when reviewing.

Image

src/room/data-stream/outgoing/header-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/** The data-streams-v2 wire signals carried directly on the header: the compression flag and the
+ * inline single-packet payload. Both used to live in reserved header attributes; they are now
+ * first-class protobuf fields on `DataStream.Header`. */
+export interface StreamHeaderV2Fields {
+  /** Compression applied to the inline/chunked payload. Defaults to `NONE` when omitted. */
+  compression?: DataStream_CompressionType;
+  /** The full payload smuggled into the header for single-packet (inline) sends. */
+  inlineContent?: Uint8Array;
+}
```
1egoman, line 20: Note to self: get rid of this and inline these two fields directly into the relevant function calls. Or name this something more generic if I decide to keep it.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -47,8 +81,52 @@ export default class OutgoingDataStreamManager {
...
+    const textInBytes = textEncoder.encode(text);
     const totalTextLength = textInBytes.byteLength;
+    const compress = options?.compress ?? true;
+
+    let info: TextStreamInfo = {
+      id: streamId,
+      mimeType: 'text/plain',
+      timestamp: Date.now(),
+      topic: options?.topic ?? '',
+      size: totalTextLength, // NOTE: size is always the pre-compression byte length
```
1egoman, line 95: question: I'm not 100% sure this is the right call, making the data stream header `size` always the uncompressed byte length. This means effectively that the value could be larger than the literal compressed bytes sent, which is a little weird.

The reason why I decided to do it is it is impossible to determine the compressed byte length ahead of time before compressing all the data. So if this were the compressed length, you'd have to receive all file chunks, compress each one, then buffer all in memory before you could finish the stream and determine the full length, which obviously isn't going to work.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -60,20 +138,46 @@ export default class OutgoingDataStreamManager {
...
+        streamId,
+        options?.destinationIdentities,
+        deflateRawCompressReadable(readableFromBytes(textEncoder.encode(text))),
+      );
 
-    await writer.write(text);
-    // set text part of progress to 1
-    handleProgress(1, 0);
+      // set text part of progress to 1
+      handleProgress(1, 0);
```
1egoman: Note to self: add some tests for the `onProgress` stuff. I'm not sure I've fully exercised it to make sure I didn't break anything here. Though worth noting I did find some bugs where in some paths `onProgress` wasn't being called before so I'm actually not convinced it was working properly before either.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -109,43 +285,21 @@ export default class OutgoingDataStreamManager {
...
+    await sendHeaderPacket(this.engine, packet);
 
     let chunkId = 0;
     const engine = this.engine;
 
+    // Incremental text streams are never compressed (CompressionStream does not support flushing
+    // mid-stream); one-shot compression lives in sendText.
+    //
+    // Note that a future streamText could send a context-takeover style deflate-raw stream with
+    // intermedia explicit `Z_SYNC_FLUSH`s - receivers already will handle this properly today.
```
1egoman, line 400: Note to reviewers - this is an important callout. The `Outgoing`- half of this change doesn't support compression for `stream`ed data streams (only `sendText` / `sendFile`) but the infrastructure exists for this on the `Incoming`- half so this could be added in the future in a fully backwards compatible way without needing a new `clientProtocol` bump.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -186,6 +330,8 @@ export default class OutgoingDataStreamManager {
       await writer.close();
     };
 
+    // FIXME: make this a global event to ensure "max listener" warning won't get logged for lots of
+    // in flight data streams.
     engine.once(EngineEvent.Closing, onEngineClose);
```
1egoman, line 436: Note to self: fix this pre-existing bug as part of this larger change, I think right now there will me a "max listener" warning if you try to send too many data streams concurrently.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -201,7 +347,60 @@ export default class OutgoingDataStreamManager {
...
+    options?: SendFileOptions,
+  ): Promise<ByteStreamInfo> {
+    const destinationIdentities = options?.destinationIdentities;
+    const compress = options?.compress ?? true;
+
+    // Phase 1: Try to send as a single packet data stream
+    //
+    // This is not being done explictly for files, because it's challenging to determine ahead of
+    // time how well the file contents will compress (and whether the total output will be under the
+    // MTU). Revisit this in the future though.
```
1egoman, line 486: Important callout for reviewers - `sendFile` has some special behavior (it doesn't do single packet data streams) because it's difficult / impossible to determine if the whole file will fit into a single packet without buffering in memory aggressively.

src/room/data-stream/compression.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/** Decompresses a raw-deflate byte array in full (inverse of {@link deflateRawCompress}). */
+export async function deflateRawDecompress(data: Uint8Array): Promise<Uint8Array> {
+  const ds = new DecompressionStream('deflate-raw');
+  const writer = ds.writable.getWriter();
+  writer.write(data as NonSharedUint8Array);
+  writer.close();
+  return collect(ds.readable);
+}
```
1egoman: Note to reviewers: I opted to use `deflate-raw` instead of `deflate` or `gzip` because `deflate-raw` is the most compact representation and the extra metadata that `deflate` includes (checksum and ordering) is superfluous in a SCTP delivered reliable data channel context.

DATA_STREAMS_SPEC.md
1egoman, line 1: Note to reviewers: here is the (mostly complete) spec for this feature.

DATA_STREAMS_SPEC.md
```diff
@@ -0,0 +1,684 @@
...
+  |              ...                     |
+  |--- streamTrailer -------------------->|  (merges trailer attrs, closes the reader)
+  |                                      |
+```
+
+- The receiver routes on `topic`: if no handler is registered for that topic, the stream is ignored
+  (chunks/trailer for an unhandled header are dropped).
+- Content is delivered incrementally as chunks arrive — a receiver must not wait for the trailer to
+  begin yielding content.
+- Chunk content larger than the MTU budget is split across multiple chunks with contiguous indices.
```
lukasIO: maybe worth calling out in the spec that text content splitting needs to account for multi-byte utf8 chars (which aren't allowed to be split), see https://github.com/livekit/client-sdk-js/blob/main/src/room/utils.ts#L758

DATA_STREAMS_SPEC.md
```diff
@@ -0,0 +1,684 @@
...
+  yield byte arrays), or
+- **read to completion** — await the full concatenated content (string for text, list/array of byte
+  arrays for bytes).
+
+The reader counts received content bytes against `totalLength` (when present) and surfaces an error
+if the stream ends short, or if more bytes than declared arrive.
+
+### E2EE
+
+When data-channel encryption is enabled, the header's `encryptionType` is `GCM` and each chunk
```
lukasIO: technically this isn't really accurate. The IV lives on the envelope of every data stream packet type (header/chunk/trailer) and not on the `chunk`.

DATA_STREAMS_SPEC.md
```diff
@@ -0,0 +1,684 @@
...
+carries an `iv`. The receiver MUST enforce a consistent `encryptionType` across a stream's
+header and chunks; a mismatch errors the stream. `totalLength` and `attributes` semantics are
+unchanged by encryption.
```
lukasIO: can you what this is supposed to convey? 
what around `attributes` semantics could potentially be changed by encryption?

DATA_STREAMS_SPEC.md
```diff
@@ -0,0 +1,684 @@
...
+---
+
+## Part 3: Send APIs
+
+Four send operations, two finite (full payload known up front) and two incremental:
+
+| API | Content | Payload known up front? | Eligible for inline? | Eligible for compression? |
+|-----|---------|-------------------------|----------------------|---------------------------|
+| `sendText(text, opts)` | text | yes | yes | yes |
+| `sendFile(file, opts)` | bytes | yes (streamed from disk) | **no** | yes |
```
lukasIO: for other SDKs this is called `sendBytes` which is arguably the better (and more generic) version of `sendFile`. 

I think we should strive for `sendBytes` as the canonical version across SDKs and maybe just keep `sendFile` as a browser special case additionally to `sendBytes`

DATA_STREAMS_SPEC.md
```diff
@@ -0,0 +1,684 @@
...
+
+Common options: `topic`, `destinationIdentities` (omit ⇒ broadcast), `attributes`
+(map<string,string>), and for the finite APIs a `compress` boolean (default `true`, opt-out).
+`sendText` additionally supports `attachments` (each becomes an attached byte stream referenced by
+`attachedStreamIds` in the text header). `streamText` additionally supports `type`
+(`create`/`update`) and `version` for streaming edits/updates of a prior stream.
+
+The v2 signals are carried in dedicated header fields, **not** attributes: `inlineContent` (the
+single-packet payload, as raw bytes) and `compression` (`NONE` / `DEFLATE_RAW`). `attributes` carries
+only caller-supplied metadata. (Earlier drafts smuggled these into reserved `lk.inline_payload` /
```
lukasIO: I think we can remove the mention of the earlier drafts ?

DATA_STREAMS_SPEC.md
```diff
@@ -0,0 +1,684 @@
...
+   = uncompressed length, then the compressed content as chunks, then the trailer.
+4. **Chunked uncompressed** (fallback): send the payload as a normal multi-packet text stream
+   (UTF-8-boundary-split chunks), header has `compression = NONE`.
+5. Send each attachment as its own byte stream (`sendFile` semantics), referenced by
+   `attachedStreamIds` in the text header.
+
+### `sendFile` send algorithm
+
+`sendFile` is fully streamed from the file's byte stream and is **never** sent inline (file uploads
+are an edge case; the inline single-packet optimization is intentionally dropped for them):
```
lukasIO: this might be true for file uploads, but is less so for the aforementioned `sendBytes`

DATA_STREAMS_SPEC.md
1egoman: Oops, fixed in b5f1c6b5, thanks

DATA_STREAMS_SPEC.md
1egoman: It's briefly mentioned further down, but I will add a more detailed callout here.

DATA_STREAMS_SPEC.md
1egoman: Throughout much of my work on this, the LLM has kept getting confused by [this deprecated iv field](https://github.com/livekit/protocol/blob/39fc751df610243c1bfdf85d3be6b3928ecef014/protobufs/livekit_models.proto#L868) on the data stream chunk. I thought I got all the references of this out of here but I must have missed this one, thanks!

DATA_STREAMS_SPEC.md
1egoman: Disregard this, see [here](https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3475815213). I will remove this section.

DATA_STREAMS_SPEC.md
1egoman: Ah, interesting. Yes, I agree, `sendBytes` makes more sense as the standard - sounds good.

1egoman: I've added in `sendBytes` here: faf1c79. This now goes down the same path / has the same tradeoffs as `sendText`.

I've kept `sendFile` in the spec for now but labelled it as optional - see a1466412. I'll revisit this after attempting to propagate it into the first other language but if it's valuable in the web sdk it's probably also valuable in other sdks which have platforms that have a first class "file" type object.

DATA_STREAMS_SPEC.md
1egoman: See https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3476140191

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -134,99 +136,179 @@ export default class IncomingDataStreamManager {
...
+          size: streamHeader.totalLength ? Number(streamHeader.totalLength) : undefined,
+          topic: streamHeader.topic,
+          timestamp: bigIntToNumber(streamHeader.timestamp),
+          attributes: streamHeader.attributes,
+          encryptionType,
+        };
+
+        // Both inline and chunked byte payloads are deflate-raw compressed; inline as a one-shot
+        // buffer, chunked as a single stream spanning all chunks (mirrors text).
+        const compressed = streamHeader.compression === DataStream_CompressionType.DEFLATE_RAW;
```
lukasIO: this currently checks only for this specific compression algorithm. 

I realise this shouldn't happen, but it would feel cleaner if it would check for any compression type other than `NONE` and then log an error if the algorithm isn't one of the expected supported ones

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -134,99 +136,179 @@ export default class IncomingDataStreamManager {
...
+          topic: streamHeader.topic,
+          timestamp: Number(streamHeader.timestamp),
+          attributes: streamHeader.attributes,
+          encryptionType,
+          attachedStreamIds: streamHeader.contentHeader.value.attachedStreamIds,
+        };
+
+        // Both inline and chunked text payloads are deflate-raw compressed; inline as a one-shot
+        // buffer, chunked as a single stream spanning all chunks.
+        const compressed = streamHeader.compression === DataStream_CompressionType.DEFLATE_RAW;
```
lukasIO: same as above

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -295,3 +377,174 @@ export default class IncomingDataStreamManager {
...
+}
+
+/**
+ * Unwraps a stream of compressed `DataStream_Chunk`s to their compressed bytes (in `chunkIndex`
+ * order), guarding ordering for the stateful decompressor that consumes them. A stateful
+ * decompressor silently corrupts on duplicated or out-of-order input, so duplicates are dropped
+ * (with a warning - in-order delivery is expected on the reliable channel, but reconnect handling
+ * may replay) and a gap is a hard error. Shared by the text and byte deflate-raw decoders.
+ */
+function orderedCompressedBytes(
```
lukasIO: when thinking about how the streams API is designed, shouldn't this be a `TransformStream` instead?

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -295,3 +377,174 @@ export default class IncomingDataStreamManager {
...
+/**
+ * Transforms a stream of deflate-raw-compressed byte `DataStream_Chunk`s into a stream of
+ * decompressed chunks, so `ByteStreamReader` consumes it unchanged. All chunk contents are fed (in
+ * `chunkIndex` order) through ONE raw-deflate decompressor for the stream's lifetime; decompressed
+ * output is re-wrapped as chunks as soon as it is produced. The sender (sendFile) compresses the
+ * whole payload in one shot, but the format also supports a single context-takeover stream
+ * sync-flushed at write boundaries, so a future incremental streamBytes could compress with no
+ * protocol change. Errors on the source stream propagate to the reader.
+ */
+function inflateRawByteChunkStream(
```
lukasIO: same question about transform stream

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -295,3 +377,174 @@ export default class IncomingDataStreamManager {
...
+
+/**
+ * Transforms a stream of deflate-raw-compressed text `DataStream_Chunk`s into a stream of
+ * decompressed chunks, so `TextStreamReader` consumes it unchanged. Builds on
+ * {@link inflateRawByteChunkStream} (single decompressor + ordering guard) and adds a streaming
+ * `TextDecoder` that reframes the decompressed bytes on UTF-8 character boundaries (a write larger
+ * than the MTU spans several packets, which may split a codepoint) so each synthesized chunk
+ * decodes independently. Errors on the source stream propagate to the reader.
+ */
+function inflateRawChunkStream(
```
lukasIO: one more transform question

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
lukasIO: yeah, it's a bit odd, but your explanation makes sense to me

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -47,8 +82,52 @@ export default class OutgoingDataStreamManager {
...
+      // makes tiny strings larger). The compression flag is carried in the header's `compression`
+      // field; user attributes are left untouched.
+      let inlineContent: Uint8Array = textInBytes;
+      let compression = DataStream_CompressionType.NONE;
+      if (
+        compress &&
+        isCompressionStreamSupported() &&
+        this.allRecipientsSupportCompression(options?.destinationIdentities)
+      ) {
+        const compressed = await deflateRawCompress(textInBytes);
```
lukasIO: is there a way we can avoid running the compression twice? currently worst case we compress here to check for the inline content size and if it doesn't fit we compress again for the chunked stream

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -88,14 +193,163 @@ export default class OutgoingDataStreamManager {
...
+   */
+  async sendBytes(bytes: Uint8Array, options?: SendBytesOptions): Promise<ByteStreamInfo> {
+    const streamId = crypto.randomUUID();
+    const destinationIdentities = options?.destinationIdentities;
+    const compress = options?.compress ?? true;
+
+    const info: ByteStreamInfo = {
+      id: streamId,
+      name: 'unknown',
+      mimeType: 'application/octet-stream',
```
lukasIO: I think I get why you hardcoded this, but wouldn't we want to let users still decide the mime type? 🤔

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
lukasIO: alternatively if the pattern is expected we can also bump the maxListener count.

lukasIO (commented):
this is looking great already. 

Really like that you consolidated a bunch of spread out things into helpers. Makes it a lot nicer to reason about.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -134,99 +136,179 @@ export default class IncomingDataStreamManager {
...
+
+        // Both inline and chunked byte payloads are deflate-raw compressed; inline as a one-shot
+        // buffer, chunked as a single stream spanning all chunks (mirrors text).
+        const compressed = streamHeader.compression === DataStream_CompressionType.DEFLATE_RAW;
 
-          this.byteStreamControllers.set(streamHeader.streamId, {
+        if (compressed && !isCompressionStreamSupported()) {
+          // NOTE: this shouldn't really ever happen, if this warning is logged then the sender
+          // isn't properly abiding by the data streams v2 protocol.
+          log.warn(
```
xianshijing-lk: nit, should it log as error or even panic as this shouldn't happen at all?

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
1egoman: I'm not a fan of just bumping the maxListener count, since you're just increasing the high water mark threshold. So IMO it either should be disabled (I forget if that's possible in the node `EventEmitter` implementation) or will address using a similar pattern as I did in [waitForBufferStatusLow here](https://github.com/livekit/client-sdk-js/pull/1896).

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -134,99 +136,179 @@ export default class IncomingDataStreamManager {
...
+            `Data stream ${streamHeader.streamId} received with deflate-raw compression, but this browser does not have support for DecompressionStream. Dropping...`,
+          );
+          return;
+        }
+
+        // Single-packet stream: the entire payload was packaged into the header's `inlineContent`.
+        // Synthesize an already-complete stream and skip waiting for chunk/trailer packets.
+        const inlineContent = streamHeader.inlineContent;
+        if (typeof inlineContent !== 'undefined') {
```
xianshijing-lk, line 202: nit, any chance inlineContent is empty ? should we handle it ?

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -134,99 +136,179 @@ export default class IncomingDataStreamManager {
...
+            // `totalLength` is the pre-compression size, and the reader counts decompressed bytes,
+            // so it applies to both paths (mirrors text).
+            bigIntToNumber(streamHeader.totalLength),
+          ),
+          {
+            identity: participantIdentity,
+          },
         );
-        return;
+        break;
```
xianshijing-lk: add a comment why it breaks rather than returns here ?

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -134,99 +136,179 @@ export default class IncomingDataStreamManager {
...
-        { identity: participantIdentity },
-      );
+            compressed ? inflateRawChunkStream(stream, streamHeader.streamId) : stream,
+            // `totalLength` is the pre-compression size, and the reader sees decompressed bytes, so
+            // it applies to both paths.
+            bigIntToNumber(streamHeader.totalLength),
+          ),
+          { identity: participantIdentity },
+        );
+        break;
```
xianshijing-lk: why not just return ?

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: An error I think could make sense, yea.

I worry about some sort of panic / uncaught error because it means any other participant could effectively DOS a client by sending it fairly trivially malformed data.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
lukasIO: yeah, the latter option sounds nice!

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -295,3 +377,174 @@ export default class IncomingDataStreamManager {
...
+        if (value.byteLength > 0) {
+          controller.enqueue(makeChunk(streamId, outIndex++, value));
+          return;
+        }
+        // Inflate can emit empty reads; keep pulling until there is content or the stream ends.
+      }
+    },
+    cancel: (reason) => {
+      decompressedReader.cancel(reason).catch(() => {});
+      srcReader.cancel(reason).catch(() => {});
```
xianshijing-lk: does decompressedReader.cancel() trigger orderedCompressedBytes.cancel() ? 
If yes, then we are double cancel() it, is it intentional ?

xianshijing-lk: curiously, do we need to call releaseLock() when it is done or cancelled ?

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -295,3 +377,174 @@ export default class IncomingDataStreamManager {
...
+
+  const encoder = new TextEncoder();
+  return new ReadableStream<DataStream_Chunk>({
+    pull: async (controller) => {
+      while (true) {
+        const { done, value } = await byteReader.read();
+        if (done) {
+          const tail = decodeOrThrow();
+          if (tail.length > 0) {
+            controller.enqueue(makeChunk(streamId, outIndex, encoder.encode(tail)));
```
xianshijing-lk: nit, just do 
controller.enqueue(makeChunk(streamId, outIndex++, encoder.encode(tail)));

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -295,3 +377,174 @@ export default class IncomingDataStreamManager {
...
+            outIndex += 1;
+          }
+          controller.close();
+          return;
+        }
+        const text = decodeOrThrow(value.content);
+        if (text.length > 0) {
+          controller.enqueue(makeChunk(streamId, outIndex, encoder.encode(text)));
+          outIndex += 1;
```
xianshijing-lk: nit, controller.enqueue(makeChunk(streamId, outIndex++, encoder.encode(tail)));

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: Discussed in 1:1 - make this logic have three states: 1) no compression supported, 2) deflate-raw supported, 3) Unknown compression algorithm supported (in this case log a warning / error and bail out early)

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: Discussed in 1:1 - I agree

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
1egoman: Discussed in 1:1 - This was an oversight, I think this can be consolidated to just do it once!

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
1egoman: Discussed in 1:1 - make `mimeType` configurable, plus make `name` default to `""` (maybe also allow this to be configurable too?)

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: Fixed in 725e90e7

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3494284737

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: Fixed in 599547f8 and 3467df85

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3494728674

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3494728674

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: While I think is a weird case (I'm not sure why you'd send an empty data stream? Maybe because the metadata is in the headers?), an empty buffer should be a valid case here. This might be a good test case to add though!

1egoman: Added some tests for this in 3d54baaa

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: It could just as easily be a return, updated in 947d01a5

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3506830434

src/room/data-stream/incoming/IncomingDataStreamManager.ts
1egoman: Lukas suspected there was some subtle issues with this stream processing code in edge cases - he recommended I [migrate to transform streams instead](https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3486037825) which I think makes a lot more sense, and causes this whole chunk of code to go away.

Now this looks like:
```tsx
function inflateRawByteChunkStream(
  raw: ReadableStream<DataStream_Chunk>,
  streamId: string,
): ReadableStream<DataStream_Chunk> {
  return raw
    .pipeThrough(ensureOrderedChunks(streamId))
    .pipeThrough(chunksToBytes())
    .pipeThrough(inflateRawTransform())
    .pipeThrough(bytesToChunks(streamId));
}
```
Which IMO is way cleaner, and cancellation is propagated automatically through properly via `.pipeThrough(...)`.

1egoman: Same thing as https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3506857005 - that also addressed this edge case as well.

lukasIO (approved):
thanks for addressing all my comments!

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
1egoman: Fixed in a540eaeb

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
1egoman: After further discussion, we decided to keep it as `"unknown"`. Fixed the `mimeType` thing and exposed `name` in 919b5922

yeongkyuyoon:
Nice work!

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -134,99 +136,205 @@ export default class IncomingDataStreamManager {
...
+          );
+          return;
+        }
 
-          this.byteStreamControllers.set(streamHeader.streamId, {
+        const stream = new ReadableStream<DataStream_Chunk>({
+          start: (controller) => {
+            streamController = controller;
+
+            if (this.textStreamControllers.has(streamHeader.streamId)) {
```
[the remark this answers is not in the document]
1egoman: Fixed in 282e7b51fb363e1f109ba20998dff2e05ba010e8

src/room/participant/RemoteParticipant.ts
```diff
@@ -47,6 +48,16 @@ export default class RemoteParticipant extends Participant {
...
+  /** The client capabilities the remote participant advertises (e.g. deflate-raw compression
+   * support). Used to decide which peer-to-peer features can be used when sending to them.
+   *
+   * Differs from clientProtocol in that these are truely optional "additions" which can be used
+   * or not depending on client specific attributes rather than protocol level invariants.
+   *
+   * @internal
+   **/
+  capabilities: Array<ClientInfo_Capability>;
+
```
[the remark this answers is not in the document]
1egoman: `capabilities` will never change after publishing an initial `RemoteParticipant`.

src/room/rpc/client/RpcClientManager.test.ts
```diff
@@ -16,6 +16,8 @@ describe('RpcClientManager', () => {
       const outgoingDataStreamManager = new OutgoingDataStreamManager(
         {} as unknown as RTCEngine,
         log,
+        (_identity) => CLIENT_PROTOCOL_DEFAULT,
+        () => [],
       );
```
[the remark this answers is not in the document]
1egoman: A breaking change for internal callers is fine.

examples/data-stream-benchmark/api.ts
```diff
@@ -0,0 +1,121 @@
...
+    roomJoin: true,
+    canPublish: true,
+    canSubscribe: true,
+  });
+
+  res.json({
+    token: await token.toJwt(),
+    url: LIVEKIT_URL,
+  });
+});
```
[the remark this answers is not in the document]
1egoman: This will be deleted before merging.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
1egoman: Fixed this (sort of) in 4a6e2bb. It definitely works better than how it did before but I think there's still some work that could be done to improve it. Right now it measures bytes before the compression stage, which means that if the compressor were to take a long time it could potentially skew the progress.

src/room/data-stream/outgoing/OutgoingDataStreamManager.ts
```diff
@@ -47,55 +82,296 @@ export default class OutgoingDataStreamManager {
...
+      const packet = createStreamHeaderPacket(header, options?.destinationIdentities);
+      await this.sendChunkedByteStream(
+        packet,
+        streamId,
+        options?.destinationIdentities,
+        compressedStream
+          .stream()
+          .pipeThrough(
+            progressReportingStream(textInBytes.length, (progress) => handleProgress(progress, 0)),
+          ),
```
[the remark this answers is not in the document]
1egoman: Discussed this here: https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3617021769

Definitely not ideal but IMO better than what was happening before, so I'm inclined to go with it for now, and maybe fix it as a follow up task.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -273,9 +398,20 @@ export default class IncomingDataStreamManager {
         );
       } else {
         textBuffer.info.attributes = { ...textBuffer.info.attributes, ...trailer.attributes };
-        textBuffer.controller.close();
-        this.textStreamControllers.delete(trailer.streamId);
+        if (trailer.reason) {
+          // A non-empty reason marks an abnormal close by the sender (e.g. an aborted send);
+          // surface it as an error rather than pretending the stream completed.
+          textBuffer.controller.error(
```
lukasIO, line 405: nice catch!

src/room/utils.ts
```diff
@@ -781,6 +781,73 @@ export function splitUtf8(s: string, n: number): NonSharedUint8Array[] {
   return result;
 }
 
+/** Wraps a byte array in a `ReadableStream` that yields it as a single chunk and then closes. */
+export function readableFromBytes(bytes: Uint8Array): ReadableStream<Uint8Array> {
```
lukasIO: any reason the `NonSharedUint8Array` isn't already declared as the type in the passed argument?

src/room/utils.ts
```diff
@@ -781,6 +781,73 @@ export function splitUtf8(s: string, n: number): NonSharedUint8Array[] {
...
+}
+
+/**
+ * Re-chunks a byte stream into pieces of exactly `chunkSize` bytes (the final piece may be
+ * smaller), coalescing or splitting the source's pieces as needed. Memory use is bounded to roughly
+ * `chunkSize` plus one source read, so it never buffers the whole stream — used to pack
+ * `CompressionStream`/`file.stream()` output into MTU-sized data-stream chunks.
+ */
+export async function* readBytesInChunks(
+  source: ReadableStream<Uint8Array>,
```
lukasIO: same question as above

package.json
```diff
@@ -72,7 +72,7 @@
   },
   "dependencies": {
     "@livekit/mutex": "1.1.1",
-    "@livekit/protocol": "1.46.6",
+    "@livekit/protocol": "1.48.0",
```
lukasIO: should we directly bump to the latest available version here?

lukasIO (approved):
great work! 

just some questions around the `NonSharedUint8Array` usage, there are some internal `Uint8Array`s declared directly in some places

src/room/utils.ts
1egoman: Good point, in hindsight I'm not sure why I didn't propagate this type throughout.

In a0e49219, I've move the casts further up and all internal types of byte arrays throughout both data streams managers and utilities are now `NonSharedUint8Array`.

src/room/utils.ts
1egoman: See https://github.com/livekit/client-sdk-js/pull/1985#discussion_r3630737561

package.json
1egoman: Upgraded to latest as of this writing (`1.50.4`) in 0d410373.

PR #1987, opened 2026-06-29
Fix initial video quality blurriness for all codecs

xianshijing-lk:
Summary

  Improve initial video quality by:
  1. Capping x-google-start-bitrate at 1 Mbps to prevent BWE from starting too aggressively on high bitrate tracks
  2. Setting source-based degradationPreference defaults for optimal quality per use case

  Changes

  PCTransport.ts
  - Added 1 Mbps cap on x-google-start-bitrate: min(target * 0.9, 1000 kbps)
  - Prevents bandwidth estimator from overwhelming the network on high bitrate tracks (e.g., 4K video)

  publishUtils.ts
  - Updated getDefaultDegradationPreference() to return defaults based on track source:
    - Camera: maintain-framerate (smoother video for real-time communication)
    - Screen share: maintain-resolution (clarity is critical for text/UI)
    - Other/unknown: balanced

  options.ts
  - Updated documentation for degradationPreference option

  Behavior
  │              Setting               │          Before           │           After            │
  │ x-google-start-bitrate             │ 90% of target (unbounded) │ min(90% of target, 1 Mbps) │
  │ Camera degradationPreference       │ maintain-resolution       │ maintain-framerate         │
  │ Screen share degradationPreference │ maintain-resolution       │ maintain-resolution        │
  │ Other degradationPreference        │ maintain-resolution       │ balanced                   │
 
  Related

  Aligns with rust-sdks PRs:
  - #1197 - Fix initial video quality blurriness for all codecs
  - #1226 - fix x-google-start-bitrate and make it adaptive to network

See Rust PR: https://github.com/livekit/rust-sdks/pull/1197

### DIFF
--- a/src/room/PCTransport.ts
+++ b/src/room/PCTransport.ts
@@ -17,13 +17,13 @@ interface TrackBitrateInfo {
   maxbr: number;
 }
 
-/* The svc codec (av1/vp9) would use a very low bitrate at the begining and
-increase slowly by the bandwidth estimator until it reach the target bitrate. The
-process commonly cost more than 10 seconds cause subscriber will get blur video at
-the first few seconds. So we use a 70% of target bitrate here as the start bitrate to
+/* Video codecs use a very low bitrate at the beginning and increase slowly by
+the bandwidth estimator until they reach the target bitrate. The process commonly
+costs more than 10 seconds causing subscribers to get blurry video at the first
+few seconds. So we use 70% of target bitrate here as the start bitrate to
 eliminate this issue.
 */
-const startBitrateForSVC = 0.7;
+const startBitrateMultiplier = 0.7;
 
 const debounceInterval = 20;
 
@@ -379,11 +379,7 @@ export default class PCTransport extends (EventEmitter as new () => TypedEmitter
             }
 
             // mung sdp for bitrate setting that can't apply by sendEncoding
-            if (!isSVCCodec(trackbr.codec)) {
-              return true;
-            }
-
-            const startBitrate = Math.round(trackbr.maxbr * startBitrateForSVC);
+            const startBitrate = Math.round(trackbr.maxbr * startBitrateMultiplier);
 
             for (const fmtp of media.fmtp) {
               if (fmtp.payload === codecPayload) {
--- a/src/room/PCTransport.ts
+++ b/src/room/PCTransport.ts
@@ -17,13 +17,17 @@ interface TrackBitrateInfo {
   maxbr: number;
 }
 
-/* Video codecs use a very low bitrate at the beginning and increase slowly by
-the bandwidth estimator until they reach the target bitrate. The process commonly
-costs more than 10 seconds causing subscribers to get blurry video at the first
-few seconds. So we use 70% of target bitrate here as the start bitrate to
-eliminate this issue.
-*/
-const startBitrateMultiplier = 0.7;
+/*
+ * Video codecs use a very low bitrate at the beginning and increase slowly by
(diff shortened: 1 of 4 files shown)


src/room/PCTransport.ts
```diff
@@ -17,13 +17,13 @@ interface TrackBitrateInfo {
...
-process commonly cost more than 10 seconds cause subscriber will get blur video at
-the first few seconds. So we use a 70% of target bitrate here as the start bitrate to
+/* Video codecs use a very low bitrate at the beginning and increase slowly by
+the bandwidth estimator until they reach the target bitrate. The process commonly
+costs more than 10 seconds causing subscribers to get blurry video at the first
+few seconds. So we use 70% of target bitrate here as the start bitrate to
 eliminate this issue.
 */
-const startBitrateForSVC = 0.7;
+const startBitrateMultiplier = 0.7;
```
lukasIO: issue: the startBitrate is at `0.7` here because the assumption is that SVC codecs are roughly 30% more effective than non-SVC codecs. 

The reason we have this multiplier is that users don't have an obvious way to differentiate startBitrates for SVC vs non SVC codecs and our presets assume `vp8` efficiency as a default.

xianshijing-lk: Based on Raja's suggestion at https://live-kit.slack.com/archives/C02EWQYP4KE/p1782726523738039?thread_ts=1780912314.410699&cid=C02EWQYP4KE, we are changing the multiplier to 0.9 for all codecs

lukasIO (approved):

src/room/participant/LocalParticipant.ts
```diff
@@ -1248,12 +1249,17 @@ export default class LocalParticipant extends Participant {
...
           }
-        } else if (track.codec && isSVCCodec(track.codec) && encodings[0]?.maxBitrate) {
-          this.engine.pcManager.publisher.setTrackCodecBitrate({
-            cid: req.cid,
-            codec: track.codec,
-            maxbr: encodings[0].maxBitrate / 1000,
-          });
+        } else if (track.codec && isVideoCodec(track.codec)) {
+          // Apply start bitrate for all video codecs to prevent initial blurriness.
+          // Sum all encoding bitrates for simulcast (BWE needs to handle all layers combined).
```
lukasIO: don't we need to differentiate between SVC and non SVC here? (i.e. keeping it with first encoding only for SVC and summing for the others?)

src/room/PCTransport.ts
lukasIO, line 374: because this uses `some` it currently only applies to the first one it finds, calling that out to ensure that's desired behaviour

src/room/participant/LocalParticipant.ts
xianshijing-lk: Theoretically yes, technically no. 

For modern SVC ((Chrome M113+), they are  single encoding with scalability mode, so it is fine. 
for legacy SVC (Safari/RN/older Chrome): they have multiple encodings describing SVC layers. Summing the value might go beyond the target max bitrate. But  given
1. 1 Mbps cap - We cap at 1 Mbps anyway, so overcounting is limited
2. Higher is safer - Starting slightly high is better than too low (which causes blurriness)
3. Legacy SVC is rare - Modern Chrome uses single encoding; mainly Safari/RN use legacy

I fixed the code anyway.

src/room/PCTransport.ts
xianshijing-lk: This should be fine, assuming each codec name should have exactly one rtp entry with one payload number in the SDP.

PR #1994, opened 2026-07-01
Data track schema metadata

ladvoc:
Adds support for attaching schema metadata to data tracks mirroring the implementation in https://github.com/livekit/rust-sdks/pull/1159.

Usage:
```ts
// Publish
const track = await room.localParticipant.publishDataTrack({
  name: 'my_sensor_data',
  schema: { name: 'rgb', encoding: 'protobuf' },
  frameEncoding: 'protobuf',
});

// Subscribe
room.on(RoomEvent.DataTrackPublished, (track) => {
  console.log(track.info.schema);         // { name: 'rgb', encoding: 'protobuf' }
  console.log(track.info.frameEncoding);  // 'protobuf'
});
```

Note: the scope of this PR has been limited to exposing schema metadata fields; schema storage will be added in a follow-up.

Areas for reviewers to scrutinize:
- Convention for Protobuf conversion

Closes BOT-441

### DIFF
--- a/src/room/data-track/outgoing/OutgoingDataTrackManager.test.ts
+++ b/src/room/data-track/outgoing/OutgoingDataTrackManager.test.ts
@@ -252,6 +252,56 @@ describe('DataTrackOutgoingManager', () => {
   });
 
   it.each([
+    {
+      title: 'well-known encodings',
+      schema: { name: 'my_schema', encoding: 'jsonSchema' },
+      frameEncoding: 'json',
+    },
+    {
+      title: 'custom encodings',
+      schema: { name: 'my_schema', encoding: { custom: 'a' } },
+      frameEncoding: { custom: 'b' },
+    },
+  ] as const)(
+    'should forward schema and frame encoding on publish ($title)',
+    async ({ schema, frameEncoding }) => {
+      const manager = new OutgoingDataTrackManager();
+      const managerEvents = subscribeToEvents<DataTrackOutgoingManagerCallbacks>(manager, [
+        'sfuPublishRequest',
+      ]);
+
+      const localDataTrack = new LocalDataTrack({ name: 'test', schema, frameEncoding }, manager);
+
+      // 1. Publish a data track with schema metadata
+      const publishRequestPromise = localDataTrack.publish();
+
+      // 2. The publish request sent to the SFU carries the schema and frame encoding
+      const sfuPublishEvent = await managerEvents.waitFor('sfuPublishRequest');
+      expect(sfuPublishEvent.schema).toStrictEqual(schema);
+      expect(sfuPublishEvent.frameEncoding).toStrictEqual(frameEncoding);
+      const handle = sfuPublishEvent.handle;
+
+      // 3. Respond as the SFU would, echoing the metadata back on the DataTrackInfo
+      manager.receivedSfuPublishResponse(handle, {
+        type: 'ok',
+        data: {
+          sid: 'bogus-sid',
+          pubHandle: handle,
+          name: 'test',
+          usesE2ee: false,
+          schema,
+          frameEncoding,
+        },
+      });
+      await publishRequestPromise;
+
+      // 4. The metadata is reflected on the local track's info
+      expect(localDataTrack.info?.schema).toStrictEqual(schema);
+      expect(localDataTrack.info?.frameEncoding).toStrictEqual(frameEncoding);
+    },
+  );
+
+  it.each([
     // Single packet payload case
     [
       new Uint8Array([0x01, 0x02, 0x03, 0x04, 0x05]),
--- a/src/room/data-track/outgoing/OutgoingDataTrackManager.test.ts
+++ b/src/room/data-track/outgoing/OutgoingDataTrackManager.test.ts
@@ -11,6 +11,7 @@ import RTCEngine from '../../RTCEngine';
 import Room from '../../Room';
 import { DataTrackHandle } from '../handle';
 import { DataTrackPacket, FrameMarker } from '../packet';
+import { DataTrackSchemaError } from '../schema';
 import OutgoingDataTrackManager, {
   type DataTrackOutgoingManagerCallbacks,
   Descriptor,
@@ -130,6 +131,26 @@ describe('DataTrackOutgoingManager', () => {
     await expect(publishRequestPromise).rejects.toThrowError(
       'Data track publication limit reached',
     );
+  });
+
+  it('should reject publishing when schema metadata is invalid', async () => {
+    const manager = new OutgoingDataTrackManager();
+    const sfuPublishRequest = vi.fn();
+    manager.on('sfuPublishRequest', sfuPublishRequest);
+
+    // Providing a schema ID without a frame encoding is invalid.
+    const localDataTrack = new LocalDataTrack(
+      { name: 'test', schema: { name: 'my_schema', encoding: 'jsonSchema' } },
+      manager,
+    );
+
+    await expect(localDataTrack.publish()).rejects.toStrictEqual(
+      DataTrackPublishError.invalidSchema(DataTrackSchemaError.missingFrameEncoding()),
+    );
+
+    // The invalid request must not be sent to the SFU.
+    expect(sfuPublishRequest).not.toHaveBeenCalled();
+    expect(localDataTrack.isPublished()).toStrictEqual(false);
   });
 
   it('should test track publishing (cancellation half way through)', async () => {
--- a/src/room/data-track/outgoing/OutgoingDataTrackManager.test.ts
+++ b/src/room/data-track/outgoing/OutgoingDataTrackManager.test.ts
@@ -135,8 +135,9 @@ describe('DataTrackOutgoingManager', () => {
 
   it('should reject publishing when schema metadata is invalid', async () => {
     const manager = new OutgoingDataTrackManager();
-    const sfuPublishRequest = vi.fn();
-    manager.on('sfuPublishRequest', sfuPublishRequest);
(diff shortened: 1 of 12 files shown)


src/room/data-track/outgoing/OutgoingDataTrackManager.test.ts
```diff
@@ -251,6 +251,56 @@ describe('DataTrackOutgoingManager', () => {
...
+      schema: { name: 'my_schema', encoding: 'jsonSchema' },
+      frameEncoding: 'json',
+    },
+    {
+      title: 'custom encodings',
+      schema: { name: 'my_schema', encoding: { custom: 'a' } },
+      frameEncoding: { custom: 'b' },
+    },
+  ] as const)(
+    'should forward schema and frame encoding on publish ($title)',
```
1egoman, line 266: Whoa, very cool, I had no idea this type of `$title` thing was possible!

src/room/data-track/schema.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    'ros2Msg',
+    'ros2Idl',
+    'omgIdl',
+    'jsonSchema',
+  ];
+
+  it.each(wellKnown)('round-trips well-known encoding %s', (encoding) => {
+    const protobuf = DataTrackSchemaEncoding.toProtobuf(encoding);
+    expect(protobuf.value.case).toEqual('wellKnown');
+    expect(DataTrackSchemaEncoding.from(protobuf)).toEqual(encoding);
```
1egoman: nitpick: I would suggest preferring `toStrictEqual` instead of `toEqual` in tests for equality comparison wherever possible. This does a `===` vs `==` among some other things too.

src/room/data-track/schema.ts
```diff
@@ -0,0 +1,197 @@
...
+  [ProtocolWellKnownFrameEncoding.FLATBUFFER]: 'flatbuffer',
+  [ProtocolWellKnownFrameEncoding.CBOR]: 'cbor',
+  [ProtocolWellKnownFrameEncoding.MSGPACK]: 'msgpack',
+  [ProtocolWellKnownFrameEncoding.JSON]: 'json',
+};
+
+export const DataTrackSchemaEncoding = {
+  from(protocol: ProtocolDataTrackSchemaEncoding): DataTrackSchemaEncoding {
+    switch (protocol.value.case) {
+      case 'wellKnown':
```
1egoman, line 123: thought: I'm a big fan of this pattern where you have a type and value which share the same name, and typescript merges them into one identifier which can be used in either context. However, @lukasIO has raised in the past this can be somewhat confusing if you don't understand what is going on.

I'll let him weigh in but I'm fine with this as is especially since it's not directly part of the external interface.

src/room/data-track/schema.ts
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * Two IDs are equal only if both components match; the same name with a different
+ * encoding refers to a distinct schema.
+ */
+export type DataTrackSchemaId = {
+  /** Name component of the identifier. Must be non-empty and no longer than 256 characters. */
+  name: string;
+  /** Encoding component of the identifier. */
+  encoding: DataTrackSchemaEncoding;
+};
```
1egoman, line 90: thought: It is weird to me to have a type which ends in `-Id` that is a non primative type. Should this be called something maybe like `DataTrackSchemaReference` or even just `DataTrackSchema`?

src/room/data-track/schema.ts
```diff
@@ -0,0 +1,197 @@
...
+  Record<ProtocolWellKnownSchemaEncoding, DataTrackSchemaEncoding>
+> = {
+  [ProtocolWellKnownSchemaEncoding.PROTOBUF]: 'protobuf',
+  [ProtocolWellKnownSchemaEncoding.FLATBUFFER]: 'flatbuffer',
+  [ProtocolWellKnownSchemaEncoding.ROS1_MSG]: 'ros1Msg',
+  [ProtocolWellKnownSchemaEncoding.ROS2_MSG]: 'ros2Msg',
+  [ProtocolWellKnownSchemaEncoding.ROS2_IDL]: 'ros2Idl',
+  [ProtocolWellKnownSchemaEncoding.OMG_IDL]: 'omgIdl',
+  [ProtocolWellKnownSchemaEncoding.JSON_SCHEMA]: 'jsonSchema',
+};
```
1egoman: nitpick: I'm assuming this always is a 1:1 mapping? If so, you (I'll leave it up to you to decide) could avoid the duplication with something like:


```suggestion
const WELL_KNOWN_TO_SCHEMA_ENCODING = Object.fromEntries(
  Object.entries(SCHEMA_ENCODING_TO_WELL_KNOWN)
    .map(([key, value]) => [value, key])
) as Record<ProtocolWellKnownSchemaEncoding, string>;
```

src/room/data-track/types.ts
```diff
@@ -1,5 +1,8 @@
 import { Encryption_Type, DataTrackInfo as ProtocolDataTrackInfo } from '@livekit/protocol';
 import { type DataTrackHandle } from './handle';
+import { DataTrackFrameEncoding, DataTrackSchemaId } from './schema';
+
+export * from './schema';
```
1egoman: nitpick: IMO don't use `export *`, it means that somebody may accidentally add something to `./schema` and get it exported from the package without realizing.

```suggestion
export { DataTrackFrameEncoding, DataTrackSchemaEncoding, DataTrackSchemaId } from './schema';
```

src/room/data-track/outgoing/types.ts
```diff
@@ -1,11 +1,22 @@
...
 /** Options for publishing a data track. */
 export type DataTrackOptions = {
   name: string;
+
+  /** Schema describing frames sent on the track. */
+  schema?: DataTrackSchemaId;
+
+  /** Encoding of frames sent on the track. */
+  frameEncoding?: DataTrackFrameEncoding;
 };
```
1egoman, line 20: thought: To confirm - is there a relationship between schema encoding and frame encoding? Or can a user specify any type of frame encoding for any time of schema encoding?

If there is a relationship and there are some combinations that aren't allowed, I think it would be good at least in the public type (`DataTrackOptions`) to make sure these non allowed combinations will cause a type error. A way you could accomplish this could be add some generics to `DataTrackOptions` (with defaults so it is backwards compatible) and use those to key into `WELL_KNOWN_TO_SCHEMA_ENCODING` / `WELL_KNOWN_TO_FRAME_ENCODING` (or a similar type of structure which represents the allowed combinations) within the type.

1egoman (approved):
Generally looks good to me!

1egoman:
Longer term, I want to discuss how to get this using the [Serializer type](https://github.com/livekit/client-sdk-js/blob/2146b2074aeb9e6d29566ee5725b9f0ab7a9cd61/src/utils/serializer.ts#L15) but I think that can wait until schema related serialization / deserialization is being done in other sdks. And I don't think the migration path to that would be too crazy.

src/room/data-track/outgoing/types.ts
ladvoc: Great callout! Yes, the valid combinations are documented in the doc comments for the enum, however, I would like to have some validation of this. I'm not sure if it is feasible to make this validated at the type level (at least not in all client SDKs), but we could at least validate at publish time. Another option is to implement this on the SFU side and have it return an error during track publication if an invalid combination is used (possible advantage: no need to update this logic in every client if new schema/frame encoding cases are added in the future). What are your thoughts here?

src/room/data-track/schema.test.ts
ladvoc: Makes sense, addressed in [6bfa1d9](https://github.com/livekit/client-sdk-js/pull/1994/commits/6bfa1d9fddcd75206c42056a68fbd4a51d5200f7).

src/room/data-track/schema.ts
ladvoc: Open to changing this, but my thinking here was this acts like a compound ID or key in a database:
- When publishing a track, you provide the ID of the schema it uses, not the schema itself.
- When storing/retrieving the definition for the schema, it is referenced by this ID.

Maybe `DataTrackSchemaKey` would make more sense here since this isn't a primitive? Also worth noting, not that we necessarily need to align with it, but it is called `-Id` at the protocol level.

src/room/data-track/schema.ts
ladvoc: I had to add a `Partial` to get the types to work. However, I am still getting this error in CI:
```
error TS2550: Property 'fromEntries' does not exist on type 'ObjectConstructor'.
```
Probably missing something obvious.

src/room/data-track/outgoing/types.ts
1egoman: > I'm not sure if it is feasible to make this validated at the type level (at least not in all client SDKs)

It's probably not something that would need to be done in _all_ client sdks, but if this isn't something that is likely to change and in the future new relationships will be purely additive (ie, `a` -> `b` will always be invalid, but `a` -> (a newly introduced) `c` may be no longer added in the future) then I think at least in a web context, it would be typical for typescript to assert this. So IMO if it's not too hard it's worth doing.

Definitely though implement this as a "denylist" and not an "allowlist" so that future newly introduced combos will work in old sdks.

> Another option is to implement this on the SFU side and have it return an error during track publication if an invalid combination is used

IMO it's worth doing this in both places, just like you'd have both client side and server side form validation.

ladvoc: Client-side validation at publish time has been added in the Rust implementation and ported to here.

lukasIO (approved):
lgtm

PR #1996, opened 2026-07-07
Commit initial AGENTS.md file

1egoman:
I've been using this locally for the past few months with success, and I think I've covered at a high level a fairly wide swath of the data channel associated parts of the sdk.

Maybe folks with more firsthand media experience can flesh out that part a bit more!


AGENTS.md
```diff
@@ -0,0 +1,115 @@
+> Note for agents: If you discover useful context about this codebase during a conversation —
```
pblazej, line 1: Nice

pblazej (approved):

PR #1997, opened 2026-07-08
Include non shared uint8array global in published package

1egoman:
## Problem

`NonSharedUint8Array` is declared as an ambient global in `src/type-polyfills/non-shared-typed-arrays.d.ts`. Because that file is a standalone ambient .d.ts (no imports/exports) that nothing imports, `rollup-plugin-typescript2` never emits it into dist. Our own build tolerates this (the global is visible project-wide during compilation), but the ~20 published .d.ts files that reference the type (`Track`, `LocalParticipant`, `Room`, `E2eeManager`, `data-track/*`, ...) include a dangling reference to a type that consumers cannot resolve.

Symptoms downstream:

- tsc with `skipLibCheck: false` -> `Cannot find name 'NonSharedUint8Array'` - 37 errors from a bare import 'livekit-client'.
- API Extractor -> `Internal Error: Unable to follow symbol for "NonSharedUint8Array"`, which broke `@livekit/components-core` / `@livekit/components-react` at release time (if this doesn't get merged, they will need to add a local shim to work around it).

## Proposed fix

Fix: make the polyfill an ordinary exported type and import it explicitly (`import type { NonSharedUint8Array }`) in every module that uses it. As a normal module export it's emitted to `dist/src/type-polyfills/` (and downleveled into dist/ts4.2/), and each consuming .d.ts now carries a resolvable import - so the published types type-check standalone. (events.ts is left untouched; its only references are in JSDoc.)

Because it's now a real module:
- tsc emits `dist/src/type-polyfills/non-shared-typed-arrays.d.ts` (and downlevel-dts produces `dist/ts4.2/type-polyfills/...`) automatically - no build-script changes required.
- The side-effect import is preserved in the emitted `index.d.ts`, so downstream consumers load the global automatically. declare global keeps the type available everywhere without touching the ~20 call sites.

The behavior of the type is unchanged (`ReturnType<typeof Uint8Array.from>`).

## Verification

- `pnpm build` and `pnpm type:check` pass.
- Polyfill is now emitted and declared in both `dist/src/type-polyfills/` and `dist/ts4.2/type-polyfills/`, and both `index.d.ts` files carry the `import './type-polyfills/non-shared-typed-arrays';`.
- A consumer that does import `livekit-client` type-checks with 0 errors under skipLibCheck: false:
  - modern types (dist/src, TS 5.9) - was 37 errors, now 0
  - downlevel types (dist/ts4.2, TS 4.8) - 0 errors

### DIFF
--- a/.changeset/ship-non-shared-uint8array.md
+++ b/.changeset/ship-non-shared-uint8array.md
@@ -2,7 +2,7 @@
 "livekit-client": patch
 ---
 
-Ship the `NonSharedUint8Array` type polyfill in the published declarations.
+Include the `NonSharedUint8Array` type polyfill in the published declarations.
 
 The type was declared in an ambient `.d.ts` that was resolved during our own
 build but never emitted to `dist`, so the published `.d.ts` files referenced a
--- a/.changeset/ship-non-shared-uint8array.md
+++ b/.changeset/ship-non-shared-uint8array.md
@@ -1,0 +1,14 @@
+---
+"livekit-client": patch
+---
+
+Ship the `NonSharedUint8Array` type polyfill in the published declarations.
+
+The type was declared in an ambient `.d.ts` that was resolved during our own
+build but never emitted to `dist`, so the published `.d.ts` files referenced a
+type consumers could not resolve — surfacing as `Cannot find name
+'NonSharedUint8Array'` under `skipLibCheck: false` and as `Unable to follow
+symbol for "NonSharedUint8Array"` in API Extractor (which broke downstream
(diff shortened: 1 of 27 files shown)


lukasIO (approved):

.changeset/ship-non-shared-uint8array.md
```diff
@@ -0,0 +1,14 @@
+---
+"livekit-client": patch
+---
+
+Ship the `NonSharedUint8Array` type polyfill in the published declarations.
```
1egoman:
```suggestion
Include the `NonSharedUint8Array` type polyfill in the published declarations.
```

PR #1998, opened 2026-07-09
chore: improve e2e signalling test tooling

lukasIO:
client side for https://github.com/livekit/livekit/pull/4653

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,0 +1,64 @@
+name: E2E
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  e2e:
+    name: Signal e2e (browser / ubuntu)
+    runs-on: ubuntu-latest
+
+    services:
+      testserver:
+        image: livekit/test-server:latest
+        ports:
+          - 9999:9999
+
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v4.4.0
+
+      - name: Use Node.js 24
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 24
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Cache Playwright browsers
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: ~/.cache/ms-playwright
+          key: ${{ runner.os }}-playwright-${{ hashFiles('pnpm-lock.yaml') }}
+
+      - name: Install Playwright Chromium
+        run: pnpm exec playwright install --with-deps chromium
+
+      - name: Run signal e2e tests
+        run: pnpm test:e2e 2>&1 | tee e2e.log
+        env:
+          # Connect to the test-server service above (validate + WS on 9999).
+          LK_TEST_SERVER_URL: ws://127.0.0.1:9999
+
+      - name: Collect test-server logs
+        if: failure()
+        run: docker logs ${{ job.services.testserver.id }} > test-server.log 2>&1 || true
+
+      - name: Upload logs
+        if: failure()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: e2e-logs
+          path: |
+            e2e.log
+            test-server.log
+          retention-days: 7
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -2,7 +2,7 @@ name: E2E
 
 on:
   push:
-    branches: [main]
+    branches: [main, "lukas/signal-e2e"]
   pull_request:
     branches: [main]
 
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -2,7 +2,7 @@ name: E2E
 
 on:
   push:
-    branches: [main, "lukas/signal-e2e"]
+    branches: [main]
   pull_request:
     branches: [main]
 
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,3 +1,11 @@
+catalog:
+  typescript: 5.9.3
+
+# downlevel-dts depends on `typescript@next`, which now resolves to the TS 7 native
+# preview (no `.` export -> ERR_PACKAGE_PATH_NOT_EXPORTED). Pin it to our canonical TS.
+overrides:
+  downlevel-dts>typescript: 'catalog:'
+
 minimumReleaseAge: 2880
 minimumReleaseAgeExclude:
   - '@livekit/*'
--- a/src/api/SignalClient.e2e.test.ts
+++ b/src/api/SignalClient.e2e.test.ts
@@ -1,8 +1,8 @@
+import { type LeaveRequest, LeaveRequest_Action } from '@livekit/protocol';
 import { afterEach, beforeEach, describe, expect, inject, it, vi } from 'vitest';
-import { LeaveRequest_Action, type LeaveRequest } from '@livekit/protocol';
 import { ConnectionErrorReason } from '../room/errors';
-import { SignalClient, SignalConnectionState, type SignalOptions } from './SignalClient';
 import { createInvalidToken, createToken } from '../test/signalToken';
+import { SignalClient, SignalConnectionState, type SignalOptions } from './SignalClient';
 
 // Provided by src/test/signalServerSetup.ts (spawns the livekit-server test-server).
 const serverUrl = inject('serverUrl');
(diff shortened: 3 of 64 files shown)


src/test/signalToken.ts
```diff
@@ -0,0 +1,59 @@
...
+  secret?: string;
+  ttlSeconds?: number;
+}
+
+export async function createToken(opts: TokenOptions = {}): Promise<string> {
+  const {
+    signal,
+    leaveAction,
+    room = 'e2e-room',
+    identity = `test-${Math.random().toString(36).slice(2, 8)}`,
```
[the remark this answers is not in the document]
lukasIO: this is only used for internal testing

src/api/SignalClient.e2e.test.ts
```diff
@@ -0,0 +1,284 @@
...
+const unavailable = inject('e2eUnavailable');
+
+const defaultOpts = (): SignalOptions => ({
+  autoSubscribe: true,
+  maxRetries: 0,
+  e2eeEnabled: false,
+  websocketTimeout: 5_000,
+});
+
+const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
```
1egoman: nitpick: could this be imported from [here](https://github.com/livekit/client-sdk-js/blob/3da1eb3a1367f29aa1a4a4734f3287aa2ef29416/src/room/utils.ts#L46-L48) instead?

src/api/SignalClient.e2e.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    it('stays connected past the ping timeout while the server pongs', async () => {
+      const onClose = vi.fn();
+      await join('happy');
+      client.onClose = onClose;
+      // Server ping timeout is 3s; a healthy pong loop must keep us alive.
+      await sleep(4_000);
+      expect(onClose).not.toHaveBeenCalled();
+      expect(client.currentState).toBe(SignalConnectionState.CONNECTED);
+    });
```
1egoman, line 79: thought: I added this `subscribeToEvents` test helper [here](https://github.com/livekit/client-sdk-js/blob/main/src/utils/subscribeToEvents.ts) when building data tracks (and I've also used it in [the data streams v2 pr](https://github.com/livekit/client-sdk-js/pull/1985)). I found it really nice for testing classes which emit events rather than manually registering / unregistering `vi.fn()` type mocks.

I realize `SignalClient` isn't an `EventEmitter` today but that helper looks like it could maybe be useful here as well. If you agree, maybe it should be converted to an `EventEmitter` and used here too.

If you want to invest in this `subscribeToEvents` helper more, it might be worth also adding something in the AGENTS.md file about it.

src/api/SignalClient.e2e.test.ts
```diff
@@ -0,0 +1,284 @@
...
+      const leave = await withTimeout(left, 5_000, 'onLeave');
+      expect(leave.action).toBe(LeaveRequest_Action.RECONNECT);
+    });
+
+    it('rejects the join when a leave arrives as the first message', async () => {
+      const err = await join('leave_first_message').then(
+        () => undefined,
+        (e) => e as Error,
+      );
+      expect(err).toBeInstanceOf(Error);
```
1egoman, line 152: nitpick: Maybe it's worth asserting more about this error here like the message?

src/api/SignalClient.e2e.test.ts
```diff
@@ -0,0 +1,284 @@
...
+      // first-message validation while in RECONNECTING (path-independent: it
+      // doesn't rely on the mock detecting reconnect, which v1 hides inside the
+      // gzipped join_request the mock ignores).
+      await join('happy');
+      const token = await createToken({ signal: 'leave_first_message' });
+      const err = await client.reconnect(serverUrl, token, 'RM_session').then(
+        () => undefined,
+        (e) => e as Error,
+      );
+      expect(err).toBeInstanceOf(Error);
```
1egoman, line 186: nitpick: Maybe it's worth asserting more about this error here like the message?

src/api/SignalClient.e2e.test.ts
```diff
@@ -0,0 +1,284 @@
...
+      const outcome = await Promise.race([
+        joinPromise.then(
+          () => 'resolved',
+          () => 'rejected',
+        ),
+        // Well past websocketTimeout (1s): a real read-timeout would have fired.
+        new Promise<typeof STILL_PENDING>((resolve) =>
+          setTimeout(() => resolve(STILL_PENDING), 3_000),
+        ),
+      ]);
```
1egoman: thought: I've found myself also needing to test situations like this, something like "is this promise still pending", sometimes within a timeout. This is the pattern I've been using: https://github.com/livekit/client-sdk-js/blob/3da1eb3a1367f29aa1a4a4734f3287aa2ef29416/src/room/data-track/incoming/IncomingDataTrackManager.test.ts#L386-L388

It you broke this out into a test only helper I think it could be valuable across a bunch of different parts of the codebase.

src/test/signalServerSetup.ts
```diff
@@ -0,0 +1,151 @@
...
+  const deadline = Date.now() + 20_000;
+  while (Date.now() < deadline) {
+    try {
+      if ((await fetch(validateUrl)).status === 200) {
+        return true;
+      }
+    } catch {
+      // not up yet
+    }
+    await new Promise((r) => setTimeout(r, 200));
```
1egoman: nitpick: IMO use `sleep` here, you've got that defined earlier already for this exact purpose?

src/test/signalServerSetup.ts
```diff
@@ -0,0 +1,151 @@
...
+  if (!serverDir || !existsSync(serverDir)) {
+    return skip(
+      'set LK_TEST_SERVER_URL (a running server) or LK_SERVER_DIR (a livekit-server checkout)',
+    );
+  }
+
+  const binary = join(serverDir, 'bin', 'test-server');
+  if (!process.env.LK_E2E_SKIP_BUILD) {
+    try {
+      execFileSync('go', ['build', '-o', 'bin/test-server', './cmd/test-server'], {
```
1egoman, line 90: question(non-blocking): is this guaranteed to always be the right command, or would it be better to call out to some sort build tool (maybe `mage`? Or maybe a bespoke script?) in the `livekit/livekit` repo?

src/test/signalToken.ts
```diff
@@ -0,0 +1,59 @@
...
+  /** LeaveRequest action the leave modes should send (0=DISCONNECT,1=RESUME,2=RECONNECT). */
+  leaveAction?: number;
+  room?: string;
+  identity?: string;
+  apiKey?: string;
+  secret?: string;
+  ttlSeconds?: number;
+}
+
+export async function createToken(opts: TokenOptions = {}): Promise<string> {
```
1egoman, line 26: question(non-blocking): would it be a good idea to pull in the node server sdk as a dev dependency to generate this token? Or would that be nonviable since you need to add the custom `lk.mock` attribute?

I mostly bring it up because this could potentially silently break if new fields were added to tokens that should be included by default in newly generated ones. Relying on the existing library should make that a non issue.

pnpm-workspace.yaml
```diff
@@ -1,3 +1,11 @@
+catalog:
+  typescript: 5.9.3
+
+# downlevel-dts depends on `typescript@next`, which now resolves to the TS 7 native
+# preview (no `.` export -> ERR_PACKAGE_PATH_NOT_EXPORTED). Pin it to our canonical TS.
+overrides:
+  downlevel-dts>typescript: 'catalog:'
```
1egoman, line 7: Makes sense, but 😬

vitest.e2e.config.mts
```diff
@@ -0,0 +1,24 @@
...
+    testTimeout: 20_000,
+    hookTimeout: 120_000,
+    // One browser context; scenarios isolate via unique-per-mode room names.
+    fileParallelism: false,
+    browser: {
+      enabled: true,
+      provider: playwright(),
+      headless: true,
+      instances: [{ browser: 'chromium' }],
+    },
```
1egoman, line 22: thought: Is there benefit to running this across multiple browsers? Guessing not for signaling, but maybe there could be for the follow up peer connection management state machine work?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,64 @@
...
+          key: ${{ runner.os }}-playwright-${{ hashFiles('pnpm-lock.yaml') }}
+
+      - name: Install Playwright Chromium
+        run: pnpm exec playwright install --with-deps chromium
+
+      - name: Run signal e2e tests
+        run: pnpm test:e2e 2>&1 | tee e2e.log
+        env:
+          # Connect to the test-server service above (validate + WS on 9999).
+          LK_TEST_SERVER_URL: ws://127.0.0.1:9999
```
1egoman, line 50: question: Just curious, how stable have these been so far in ci?

1egoman (commented):
Generally makes sense to me!

Just wanted to confirm one thing to make sure I am reading this right - running `npm test` wouldn't cause the e2e tests to run by default correct? You'd need to explictly run `test:e2e` and set the relevant `LK_` prefixed env vars to the server environment?

src/api/SignalClient.e2e.test.ts
lukasIO: I'll leave converting SignalClient to an EventEmitter for a potential follow up

src/test/signalServerSetup.ts
1egoman: Still am somewhat curious here, is there a more robust way to handle this?

1egoman (approved):

src/test/signalServerSetup.ts
lukasIO: it's the command that is also used in the server repo for the test server. 
note that this is only ever used for local tests against custom server changes if we're not pulling in the existing docker image (which is what the CI does).

.github/workflows/e2e.yaml
lukasIO: very stable so far!

PR #2000, opened 2026-07-09
Handle data track SID reassignment

ladvoc:
Equivalent to: https://github.com/livekit/rust-sdks/pull/1228

Closes BOT-457

### DIFF
--- a/src/room/data-track/incoming/IncomingDataTrackManager.ts
+++ b/src/room/data-track/incoming/IncomingDataTrackManager.ts
@@ -147,6 +147,10 @@ export default class IncomingDataTrackManager extends (EventEmitter as new () =>
     let streamController: ReadableStreamDefaultController<DataTrackFrame> | null = null;
     const sfuSubscriptionComplete = new Future<void, DataTrackSubscribeError>();
 
+    // Hold the descriptor by reference: SID reassignment re-keys the map, so lookups
+    // with the SID captured at subscribe time would fail.
+    const descriptor = this.descriptors.get(sid);
+
     const detachSignal = () => {
       signal?.removeEventListener('abort', onAbort);
     };
@@ -158,8 +162,7 @@ export default class IncomingDataTrackManager extends (EventEmitter as new () =>
         log.warn(`ReadableStream subscribed to ${sid} was not started.`);
         return;
       }
-      const descriptor = this.descriptors.get(sid);
-      if (!descriptor) {
+      if (!descriptor || this.descriptors.get(descriptor.info.sid) !== descriptor) {
         log.warn(`Unknown track ${sid}, skipping cancel...`);
         return;
       }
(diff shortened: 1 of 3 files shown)


src/room/data-track/incoming/IncomingDataTrackManager.ts
```diff
@@ -499,6 +502,62 @@ export default class IncomingDataTrackManager extends (EventEmitter as new () =>
...
+    descriptor.info.sid = newSid;
+
+    switch (descriptor.subscription.type) {
+      case 'none':
+        break;
+      case 'pending':
+      case 'active':
+        // The SFU does not carry subscriptions across a publisher's full
+        // reconnect; re-request the subscription under the new SID.
+        this.emit('sfuUpdateSubscription', { sid: newSid, subscribe: true });
```
lukasIO, line 550: is there a scenario in which the user might have unsubscribed to the data track and we're re-forcing it to subscribe here?

1egoman: One possibility that comes to mind:
1. A user completes their final subscription for a data track, which calls `IncomingDataTrackManager.unSubscribeRequest` ([see here](https://github.com/livekit/client-sdk-js/blob/3da1eb3a1367f29aa1a4a4734f3287aa2ef29416/src/room/data-track/incoming/IncomingDataTrackManager.ts#L154-L192)). but otherwise leaves the descriptor in `active` type
2. A SFU full reconnect occurs, and the client must re subscribe via the new SFU to all remote data tracks
3. Because of the `active` type on the descriptor, the unsubscribed data track from 1 is resubscribed

@ladvoc Given this, maybe it's worth introducing a `unsubscribing` or similar descriptor type which can mark that the descriptor is on its way out and should be dropped during a reconnect. Or maybe another way to do this could be to only republish if `descriptor.streamControllers.size > 0`. Thoughts? I think this would also need to be done on the rust one too, as that also [only has the three none / pending / active states](https://github.com/livekit/rust-sdks/blob/16742b4248992b8bd0cd3b0676e4633d12b375ba/livekit-datatrack/src/remote/manager.rs#L451-L468).

ladvoc: `unSubscribeRequest` doesn't leave the descriptor in the active state:
https://github.com/livekit/client-sdk-js/blob/3da1eb3a1367f29aa1a4a4734f3287aa2ef29416/src/room/data-track/incoming/IncomingDataTrackManager.ts#L407-L409

1egoman (approved):
Talked about it in a 1:1, generally makes sense

src/room/data-track/incoming/IncomingDataTrackManager.ts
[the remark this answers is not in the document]
ladvoc: This is accurate, but it was not introduced by this PR; will address in a follow-up.

PR #2002, opened 2026-07-10
fix: remove Firefox VP9 SVC workaround

jibon57:
Removes the workaround that disabled VP9 publishing for Firefox clients. According to the linked Bugzilla thread, this issue has been resolved in modern Firefox versions.

Fixes: https://bugzilla.mozilla.org/show_bug.cgi?id=1633876
Discussion: https://livekit-users.slack.com/archives/C025KM0S1CK/p1783681641812789

### DIFF
--- a/.changeset/silly-crews-show.md
+++ b/.changeset/silly-crews-show.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Remove Firefox VP9 SVC workaround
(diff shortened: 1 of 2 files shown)


1egoman:
Thanks for the contribution! I think @lukasIO / @cnderrauber are probably best to review this.

lukasIO:
thanks for the PR. Just as a heads up, we’ll need to verify that this is actually fixed and addresses the specific SVC case and not “just” simulcast.
Also the fix would have to be gated on a specific version rather to ensure it’s still gated for older FF versions

PR #2004, opened 2026-07-11
fix: keep attached audio elements muted in startAudio() when webAudioMix is enabled

### LINKED ISSUE #2003: startAudio() unmutes attached audio elements even when webAudioMix is enabled, causing double playback on iOS Safari
### Describe the bug

**What I'm expecting**

With `webAudioMix` enabled, remote audio should only be output through the WebAudio path (gain node). Attached `HTMLAudioElement`s should stay muted, as `RemoteAudioTrack.attach()` itself deliberately sets `element.muted = true` / `element.volume = 0` when a web audio connection is made.

**What happens instead**

`Room.startAudio()` unconditionally unmutes **all** attached remote audio elements:

```ts
// src/room/Room.ts (v2.20.1, ~L1385)
...elements.map((e) => {
  e.muted = false;
  return e.play();
}),
```

This breaks the invariant established by `attach()`. Note that `startAudio()` runs automatically — not only when called by app code:

1. On `ParticipantEvent.AudioStreamAcquired`, i.e. whenever the local microphone track is created (e.g. `setMicrophoneEnabled(true)` right after connecting)
2. On `visibilitychange` when the page becomes visible again, via the iOS dummy audio element handler inside `startAudio()`

On desktop browsers this goes unnoticed because `element.volume = 0` still applies, so the unmuted element remains silent. **On iOS Safari, setting `volume` on a media element has no effect**, so once the element is unmuted the same track plays twice: once through the element's direct output and once through the WebAudio gain node — audible as doubled/echoing voices.

The unmute-on-mic-acquisition case is intermittent, which makes it hard to diagnose: it only affects tracks that were already attached when `startAudio()` ran (a race between `TrackSubscribed` and microphone acquisition). The background/re-foreground case reproduces reliably since all elements are attached by then.

### Reproduction

1. Create a `Room` with `webAudioMix: true` (or `webAudioMix: { audioContext }`) and connect on iOS Safari, joining a room that already has a publishing remote participant
2. Control per-participant volume via `participant.setVolume()` (gain node path), keeping the attached elements muted
3. Enable the local microphone after the remote tracks are subscribed (`setMicrophoneEnabled(true)`), or background and re-foreground Safari
4. Inspect the attached audio elements: `[...document.querySelectorAll('audio')].map(a => a.muted)` → they have been flipped to `false`, and remote voices play doubled (element output + web audio output)

### Logs

```shell
n/a — no errors are logged; the behavior is a state change (element.muted flipped to false by startAudio()).
```

### System Info

```shell
livekit-client: 2.18.8 (the relevant code is unchanged in v2.20.1)
OS/Browser where audible: iOS 18 Safari (any recent version; element.volume is ineffective on iOS)
Desktop browsers: element is also unmuted, but stays silent because volume = 0 still applies
```

### Severity

### Additional Information

Suggested fix — skip the unmute when webAudioMix is handling output, while still calling `play()` to satisfy autoplay recovery:

```ts
...elements.map((e) => {
  if (!this.options.webAudioMix) {
    e.muted = false;
  }
  return e.play();
}),
```

We have been running exactly this as a local patch (patch-package) in production (a WebXR social VR app using webAudioMix for spatial audio). It resolves the double playback on iOS Safari without regressing desktop browsers or the non-webAudioMix path.

Happy to submit a PR.

sawa-zen:
Fixes #2003

## Problem

`Room.startAudio()` unconditionally sets `e.muted = false` on all attached remote audio elements. When `webAudioMix` is enabled, those elements are deliberately kept muted by `RemoteAudioTrack.attach()` (audio is routed through the web audio gain node instead), so unmuting them re-enables the element's direct output.

On desktop browsers this stays inaudible because `element.volume = 0` still applies, but on iOS Safari `element.volume` has no effect — the same track then plays twice (element output + web audio output), audible as doubled/echoing voices.

Since `startAudio()` also runs automatically (on `ParticipantEvent.AudioStreamAcquired` and on the iOS `visibilitychange` handler), this happens without any app-side call to `startAudio()`.

## Fix

Skip the unmute when `webAudioMix` is enabled, restoring the invariant established by `attach()`. `play()` is still called so autoplay recovery keeps working. Behavior without `webAudioMix` is unchanged.

We have been running this exact change as a local patch in production (a WebXR social VR app using webAudioMix for spatial audio) and it resolves the double playback on iOS Safari without regressions on desktop browsers.

### DIFF
--- a/.changeset/spotty-donuts-remain.md
+++ b/.changeset/spotty-donuts-remain.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Keep attached audio elements muted in startAudio() when webAudioMix is enabled, preventing double playback on platforms where element.volume has no effect (e.g. iOS Safari)
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -1383,7 +1383,13 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       await Promise.all([
         this.acquireAudioContext(),
         ...elements.map((e) => {
-          e.muted = false;
+          // when webAudioMix is enabled, attached elements are deliberately kept muted by
+          // RemoteAudioTrack.attach() and audio is routed through the web audio graph instead.
+          // unmuting them here would cause double playback on platforms where element.volume
+          // has no effect (e.g. iOS Safari)
+          if (!this.options.webAudioMix) {
+            e.muted = false;
+          }
           return e.play();
         }),
       ]);


sawa-zen:
The failing `test` job is unrelated to this PR — the `throws:check` step is currently broken for **all** new PRs in this repo.

**What's happening:**

`pnpm --package=@livekit/throws-transformer dlx throws-check ...` crashes while locating `tsconfig.json`, before analyzing any source:

```
TypeError: Cannot read properties of undefined (reading 'fileExists')
    at main (.../throws-transformer/dist/cli.js:61:66)  // ts.sys.fileExists
```

**Root cause:**

- `@livekit/throws-transformer` declares an open-ended peer range: `typescript: ">=4.7.0"`
- `pnpm dlx` resolves peers to the latest matching version on every run
- npm's `typescript` `latest` tag recently moved to **7.0.2** (the Go-based rewrite), whose JS API no longer exposes `ts.sys`

Evidence from the logs: the last green run (Jul 10) resolved 9 packages for the dlx install, while this run (Jul 11) resolved 29 — TS 7 pulls in per-platform native binaries — and the dlx cache path in the stack trace shows `@livekit+throws-transformer@0.1.8_typescript@7.0.2`.

**Possible fixes:**

- Quick: pin TypeScript in the `throws:check` script, e.g. `pnpm --package=typescript@^6 --package=@livekit/throws-transformer dlx throws-check ...`
- Proper: cap the peer range in `@livekit/throws-transformer` (e.g. `>=4.7.0 <7`), or migrate it to the TS 7 API

Happy to split the quick fix into a separate PR if that helps.

lukasIO (approved):
nice catch! thanks for the fix

PR #2010, opened 2026-07-13
Address data channel thundering herd due to concurrent sendFile(...) calls

### LINKED ISSUE #1995: Concurrent sendFile() calls can kill the reliable data channel
### What I'm expecting
Calling sendFile() for a few files without awaiting in between should just work (maybe slowly).

### What happens instead
In my app users can drop several files into chat at once. When they do, the reliable data channel dies mid-transfer and usually only one file out of N arrives, the rest stall forever. Console shows `publisher data channel 'RELIABLE' closed unexpectedly`.

I dug into it and I'm pretty sure it's a check-then-send race in `RTCEngine.sendDataPacket`:

    await this.waitForBufferStatusLow(kind);
    if (this.attemptingReconnect) return;
    dc.send(msg);

`waitForBufferStatusLow` resolves for every waiter once bufferedAmount drops below the threshold, but nothing reserves capacity — so N concurrent stream writers all see "buffer low" in the same tick and all call dc.send() together, blowing past the SCTP send buffer. The browser then aborts the channel. Your close handler comment even describes this signature ("A publisher DC closing while the session is up ... see livekit/rust-sdks#1137").

The per-writer Mutex in streamBytes only serializes chunks within one stream, not across streams. Sending files strictly one at a time never triggers it. 
In a minimal standalone repro (logs below) it shows up as the other failure mode: all sender promises resolve fulfilled while the receiver gets files 1-4 truncated, silent data loss instead of the channel dying.

### Reproduction

    // several files, even small ones (1-20 MB each), no await between calls:
    for (const file of files) {
      room.localParticipant.sendFile(file, { topic: 'files' });
    }

Two participants (sender and receiver with a registerByteStreamHandler).
Sender picks 3-5 files at once. With 5 x 15 MB: file 0 arrives complete,
files 1-4 truncated. Sending strictly one at a time never triggers it;
that's my workaround (FIFO, one open writer at a time). 

### Logs

```shell
receiver:
RECEIVED blob-0.bin 15728640
RECEIVE FAILED blob-1.bin DataStreamError: Not enough chunk(s) received - expected 15728640 bytes of data total, only received 12536488 bytes
RECEIVE FAILED blob-2.bin DataStreamError: Not enough chunk(s) received - expected 15728640 bytes of data total, only received 9920768 bytes
RECEIVE FAILED blob-3.bin DataStreamError: Not enough chunk(s) received - expected 15728640 bytes of data total, only received 7964336 bytes
RECEIVE FAILED blob-4.bin DataStreamError: Not enough chunk(s) received - expected 15728640 bytes of data total, only received 6419336 bytes

sender (same run):
send 0 fulfilled
send 1 fulfilled
send 2 fulfilled
send 3 fulfilled
send 4 fulfilled
```

### System Info

```shell
System:
  OS: Fedora Linux 44 (Workstation)
Browsers:
  Chrome: 150.0.7871.46 (Official Build) (64-bit)
npmPackages:
  livekit-client: 2.20.0
Server: self-hosted livekit/livekit-server v1.13.2 (reproduced browser-to-browser through the SFU)
```

### Severity

### Additional Information

Looked at the Data streams v2 branch (#1985), it doesn't touch RTCEngine.ts, and its write mutex in the new OutgoingDataStreamManager is still created per-writer inside streamBytes(), so the race should survive the rewrite as-is.

1egoman:
Fixes #1995

## Problem

Sending `n` files concurrently over a data stream can result in all of the `sendFile` calls attempting to send their first packet via the reliable data channel at the same time. When this happens, multiple writers can think the buffer status is low at the same time and all call `dc.send(...)`, exhausting the data channel buffer and causing the buffer to overflow (which causes data to get dropped).

## Pre-existing workaround
The reporter of the issue has had success with converting concurrent `sendFile` calls into sequential ones - so `await Promise.all([sendFile(...), sendFile(...), sendFile(...)])` vs `await sendFile(...); await sendFile(...); await sendFile(...);`. The latter works because there's only one data stream actively writing to the data channel at a time, so there's no contention.

## Solution
To address this, I've opted to sequentialize `waitForBufferStatusLow(...)` resolutions when the the data channel is not "low". So what will now happen:
1. `await waitForBufferStatusLow(...)` works as it did before as long as it is in "low status", and by default has no sequentializing semantics.
2. Once `await waitForBufferStatusLow(...)` gets called and the data channel is no longer in low status, then the `resolve` is cached in `waitForBufferedStatusLowResolves`. Once the data channel is no longer in "low status", instead of resolving all promises ASAP (ie, a "thundering herd"), instead resolve them in sequence and verify that the data channel buffer is still in "low status" between resolutions. This way events naturally resolve in the order they were called until the data channel buffer is full again, where they pause until the buffer is no longer full again and so on until the buffer exhausts.

## Demo

<details><summary>Code added to demo app to test this</summary>
<p>

```diff
diff --git a/examples/demo/demo.ts b/examples/demo/demo.ts
index 3471049a..002d744e 100644
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -168,6 +168,54 @@ const appActions = {
       onProgress: (progress) => console.log('sending file, progress', Math.ceil(progress * 100)),
     });
   },
+  // Reproduces https://github.com/livekit/client-sdk-js/issues/1995 — several large
+  // sendFile() calls fired in parallel without awaiting between them race on the
+  // reliable data channel's buffer-low signal, so the receiver sees truncated files
+  // or the channel closes with "publisher data channel 'RELIABLE' closed unexpectedly".
+  reproduceParallelSendFile: async () => {
+    const lp = currentRoom?.localParticipant;
+    if (!lp) {
+      appendLog('reproduce: not connected to a room');
+      return;
+    }
+    const fileCount = 5;
+    const fileSizeBytes = 15.7 * 1024 * 1024; // ~15.7 MB, matching the issue report
+    appendLog(
+      `reproduce #1995: firing ${fileCount} parallel sendFile() calls of ~${(
+        fileSizeBytes /
+        (1024 * 1024)
+      ).toFixed(1)} MB each (no await between calls)`,
+    );
+
+    const files = Array.from({ length: fileCount }, (_, i) => {
+      // Fill with a per-file byte pattern so the receiver can spot truncation/corruption.
+      const bytes = new Uint8Array(fileSizeBytes);
+      bytes.fill((i + 1) & 0xff);
+      return new File([bytes], `repro-${i + 1}.bin`, { type: 'application/octet-stream' });
+    });
+
+    // Intentionally do NOT await sequentially — this is the buggy usage pattern.
+    const sends = files.map((file, i) =>
+      lp
+        .sendFile(file, {
+          mimeType: file.type,
+          topic: 'files',
+          onProgress: (progress) =>
+            console.log(`reproduce: file ${i + 1} send progress`, Math.ceil(progress * 100)),
+        })
+        .then((info) => appendLog(`reproduce: file ${i + 1} send resolved (streamId ${info.id})`))
+        .catch((err) => appendLog(`reproduce: file ${i + 1} send FAILED: ${err}`)),
+    );
+
+    try {
+      await Promise.all(sends);
+      appendLog(
+        'reproduce #1995: all send promises resolved — check the receiver, files are likely truncated',
+      );
+    } catch (err) {
+      appendLog(`reproduce #1995: a send rejected: ${err}`);
+    }
+  },
   connectWithFormInput: async () => {
     const url = (<HTMLInputElement>$('url')).value;
     const token = (<HTMLInputElement>$('token')).value;
@@ -479,7 +527,17 @@ const appActions = {
         throw err;
       }
       const result = new Blob(byteContents, { type: info.mimeType });
-      appendLog(`Completely received file "${info.name}" from ${participant?.identity}`);
+      const receivedBytes = result.size;
+      const expectedBytes = info.size;
+      const sizeSuffix =
+        expectedBytes != null
+          ? ` (${receivedBytes} / ${expectedBytes} bytes${
+              receivedBytes !== expectedBytes ? ' — TRUNCATED!' : ''
+            })`
+          : ` (${receivedBytes} bytes)`;
+      appendLog(
+        `Completely received file "${info.name}" from ${participant?.identity}${sizeSuffix}`,
+      );
 
       streamReaderAbortController = undefined;
       (<HTMLButtonElement>$('cancel-chat-receive-button')).style.display = 'none';
diff --git a/examples/demo/index.html b/examples/demo/index.html
index 9ec439b5..d8d8a939 100644
--- a/examples/demo/index.html
+++ b/examples/demo/index.html
@@ -190,6 +190,12 @@
               </button>
               <input id="file" type="file" />
               <button onclick="appActions.sendFile()">Send file</button>
+              <button
+                onclick="appActions.reproduceParallelSendFile()"
+                title="Fires several large sendFile() calls in parallel to reproduce https://github.com/livekit/client-sdk-js/issues/1995"
+              >
+                Reproduce #1995 (parallel sendFile)
+              </button>
               <select
                 id="simulate-scenario"
                 class="custom-select"
```

</p>
</details> 

Before, here was the behavior:

https://github.com/user-attachments/assets/e49cf096-adca-42b2-b44e-7b33915b81d1

Now, with this pull request:

https://github.com/user-attachments/assets/82a9d970-f65c-49d1-acfb-7d3286f2f41a

Note the decreased upload speed, because the bandwidth is being split between all file uploads.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -1743,7 +1743,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
         this.waitForBufferedStatusLowResolves.push(resolve);
         if (!this.waitForBufferStatusLowAbortController) {
           this.waitForBufferStatusLowAbortController = new AbortController();
-          dc.addEventListener('bufferedamountlow', () => {
+          dc.addEventListener('bufferedamountlow', async () => {
             while (true) {
               const resolve = this.waitForBufferedStatusLowResolves[0];
               if (!resolve) {
@@ -1758,6 +1758,9 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
                 break;
               }
               resolve();
+              // Defer to the event loop so any `await dc.send(...)` calls can fire and fill back up
+              // the data channel.
+              await new Promise((r) => setTimeout(r, 0));
               this.waitForBufferedStatusLowResolves.shift();
             }
           }, { signal: this.waitForBufferStatusLowAbortController.signal });
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -1715,6 +1715,12 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
   };
 
+  /** Abort controller which when called will remove the `bufferedamountlow` event on the data channel. */
+  private waitForBufferStatusLowAbortController: AbortController | null = null;
+  /** List of resolve functions which are waiting to be called by {@link waitForBufferStatusLow}.
+    * Note that these will be called in order, validating that the buffer status is still low
+    * between each call. */
+  private waitForBufferedStatusLowResolves: Array<() => void> = [];
   async waitForBufferStatusLow(kind: DataChannelKind) {
     return new TypedPromise<void, UnexpectedConnectionState>(async (resolve, reject) => {
       if (this.isClosed) {
@@ -1728,10 +1734,34 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
           reject(new UnexpectedConnectionState(`DataChannel not found, kind: ${kind}`));
           return;
         }
+
+        // Proxy along any errors due to the engine closing
         this.bufferStatusLowClosingFuture.promise.catch((e) => reject(e));
-        dc.addEventListener('bufferedamountlow', () => resolve(), {
-          once: true,
-        });
+
+        // Store resolve so that when the bufferedamountlow event fires, all resolve calls can be
+        // fired in order until the data channel buffer fills up again.
+        this.waitForBufferedStatusLowResolves.push(resolve);
+        if (!this.waitForBufferStatusLowAbortController) {
+          this.waitForBufferStatusLowAbortController = new AbortController();
+          dc.addEventListener('bufferedamountlow', () => {
+            while (true) {
+              const resolve = this.waitForBufferedStatusLowResolves[0];
+              if (!resolve) {
+                this.waitForBufferedStatusLowResolves = [];
+                this.waitForBufferStatusLowAbortController?.abort();
+                this.waitForBufferStatusLowAbortController = null;
+                break;
+              }
+              if (!this.isBufferStatusLow(kind)) {
+                // Buffer status no longer low, bail out and resume once the next bufferedamountlow
+                // event fires again.
+                break;
+              }
+              resolve();
+              this.waitForBufferedStatusLowResolves.shift();
+            }
+          }, { signal: this.waitForBufferStatusLowAbortController.signal });
+        }
       }
     });
   }
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -1715,58 +1715,58 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     }
   };
 
-  /** Abort controller which when called will remove the `bufferedamountlow` event on the data channel. */
-  private waitForBufferStatusLowAbortController: AbortController | null = null;
-  /** List of resolve functions which are waiting to be called by {@link waitForBufferStatusLow}.
-    * Note that these will be called in order, validating that the buffer status is still low
-    * between each call. */
-  private waitForBufferedStatusLowResolves: Array<() => void> = [];
+  /** Per-kind lock serializing callers of {@link waitForBufferStatusLow} that have to wait. */
+  private waitForBufferStatusLowLocks = new Map<DataChannelKind, Mutex>();
+
   async waitForBufferStatusLow(kind: DataChannelKind) {
-    return new TypedPromise<void, UnexpectedConnectionState>(async (resolve, reject) => {
-      if (this.isClosed) {
-        reject(new UnexpectedConnectionState('engine closed'));
-      }
-      if (this.isBufferStatusLow(kind)) {
-        resolve();
-      } else {
+    if (this.isClosed) {
+      throw new UnexpectedConnectionState('engine closed');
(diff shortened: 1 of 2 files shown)


src/room/RTCEngine.ts
```diff
@@ -1715,25 +1715,58 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-      if (this.isClosed) {
-        reject(new UnexpectedConnectionState('engine closed'));
-      }
-      if (this.isBufferStatusLow(kind)) {
-        resolve();
-      } else {
+    if (this.isClosed) {
+      throw new UnexpectedConnectionState('engine closed');
+    }
+    if (this.isBufferStatusLow(kind)) {
```
lukasIO, line 1725: issue: this path here means ordering isn't guaranteed anymore. 
it needs to live _after_ acquiring the lock

src/room/RTCEngine.ts
```diff
@@ -1715,25 +1715,58 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
         });
-      }
-    });
+      });
+    } finally {
+      // Release only after the caller's synchronous dc.send() has run, so the next waiter reads the
+      // updated bufferedAmount rather than the stale (still-low) value. The caller's continuation is
+      // scheduled when this promise settles; deferring the unlock to a later task lets that send run
+      // first.
+      setTimeout(unlock, 0);
```
lukasIO, line 1768: how many packets will this affect worse case? 

I'm not a huge fan of the `setTimeout` here as timeouts get heavily throttled under certain circumstances, e.g. in backgrounded tabs

1egoman:
Lukas and I have been going back and forth on this in the background - it's fairly tricky to get right and come up with a solution that generalizes both to data streams and data tracks. Closing this in favor of https://github.com/livekit/client-sdk-js/pull/2013, as this approach seems to generalize better.

PR #2011, opened 2026-07-14
Remove stale data track subscriber handles

ladvoc:
Follow-up to #2000


1egoman (approved):
Makes sense, good catch!

For my own understanding, can you outline a scenario where a packet would be received from an old subscribe handle after a full reconnect? Maybe something like an old packet was delayed when being sent over the network and takes longer than normal to be delivered?

PR #2013, opened 2026-07-15
Add new data channel high / low water mark approach for packet congestion control

### LINKED ISSUE #1995: Concurrent sendFile() calls can kill the reliable data channel

1egoman:
(Started by Ryan -> Lukas took over this change and wrote the below)

closes https://github.com/livekit/client-sdk-js/issues/1995

the gist of this PR is to add a kind dependent mutex around data channel sends. 
Previously the `waitForBufferStatusLow` method could be stalling lots of data requests in parallel and once it resolved all of them got flushed to the data channel at once potentially overflowing its buffer. 

To generalise this solution without any performance/throughput hits the PR additionally introduces high and low watermarks for the data channels so that there's a designated "headroom" under which data packets are allowed to be pushed to the data channel buffer immediately (avoiding a constant back and forth between high and low DC buffer status)

### DIFF
--- a/src/room/RTCEngine.test.ts
+++ b/src/room/RTCEngine.test.ts
@@ -231,7 +231,7 @@ describe('RTCEngine', () => {
       const send = vi.fn();
       Object.assign(engine as unknown as Record<string, unknown>, {
         ensurePublisherConnected: vi.fn().mockResolvedValue(undefined),
-        waitForBufferHeadroom: vi.fn().mockResolvedValue(undefined),
+        waitForBufferHeadroomWithLock: vi.fn().mockResolvedValue(undefined),
         updateAndEmitDCBufferStatus: vi.fn(),
         dataChannelForKind: vi.fn(() => ({ send })),
         pcManager: {
@@ -322,8 +322,7 @@ describe('RTCEngine', () => {
         },
       });
 
-      // Two messages queued for replay, and a full buffer so the replay parks on
-      // waitForBufferHeadroom before its first send.
+      // Two messages queued for replay, and a full buffer so the XX its first send.
       const replayed1 = new Uint8Array([1]);
       const replayed2 = new Uint8Array([2]);
       const buffer = (
@@ -379,7 +378,7 @@ describe('RTCEngine', () => {
         .reliableMessageBuffer;
       const replayed = new Uint8Array([1]);
       buffer.push({ data: replayed, sequence: 1, sent: true });
-      // Full buffer so replay parks on waitForBufferHeadroom before its first send.
+      // Full buffer so replay parks on waitForBufferHeadroomWithLock before its first send.
       dc.bufferedAmount = 2 * 1024 * 1024;
 
       const replay = (
@@ -540,7 +539,7 @@ describe('RTCEngine', () => {
     });
   });
 
-  describe('waitForBufferHeadroom', () => {
+  describe('waitForBufferHeadroomWithLock', () => {
     it('rejects parked waiters and releases the lock when the data channels are invalidated', async () => {
       const engine = new RTCEngine(roomOptionDefaults);
       const dc = new FakeDataChannel();
@@ -551,7 +550,7 @@ describe('RTCEngine', () => {
 
       // Park a waiter: buffer above the reliable high-water mark, holding the headroom lock.
       dc.bufferedAmount = 2 * 1024 * 1024;
-      const parked = engine.waitForBufferHeadroom(DataChannelKind.RELIABLE);
+      const parked = engine.waitForBufferHeadroomWithLock(DataChannelKind.RELIABLE);
       // Swallow the expected rejection so it can't surface as unhandled before we assert on it.
       parked.catch(() => {});
       await tick();
@@ -566,7 +565,9 @@ describe('RTCEngine', () => {
       // The lock must be free again: a wait against the fresh, drained channel resolves instead
       // of queueing forever behind the stranded waiter.
       dc.bufferedAmount = 0;
-      await expect(engine.waitForBufferHeadroom(DataChannelKind.RELIABLE)).resolves.toBeUndefined();
+      await expect(
+        engine.waitForBufferHeadroomWithLock(DataChannelKind.RELIABLE),
+      ).resolves.toBeUndefined();
     });
   });
 
--- a/src/room/RTCEngine.test.ts
+++ b/src/room/RTCEngine.test.ts
@@ -1,6 +1,6 @@
 import { DataPacket, DataPacket_Kind, UserPacket } from '@livekit/protocol';
 import { afterEach, describe, expect, it, vi } from 'vitest';
-import type { DataPacketBuffer } from '../utils/dataPacketBuffer';
+import type { DataPacketBuffer, DataPacketItem } from '../utils/dataPacketBuffer';
 import RTCEngine, { DataChannelKind } from './RTCEngine';
 import { roomOptionDefaults } from './defaults';
 import { PublishDataError, UnexpectedConnectionState } from './errors';
@@ -326,8 +326,9 @@ describe('RTCEngine', () => {
       // waitForBufferHeadroom before its first send.
       const replayed1 = new Uint8Array([1]);
       const replayed2 = new Uint8Array([2]);
-      const buffer = (engine as unknown as { reliableMessageBuffer: { push: Function } })
-        .reliableMessageBuffer;
+      const buffer = (
+        engine as unknown as { reliableMessageBuffer: { push: (item: DataPacketItem) => void } }
+      ).reliableMessageBuffer;
       buffer.push({ data: replayed1, sequence: 1, sent: true });
       buffer.push({ data: replayed2, sequence: 2, sent: true });
       dc.bufferedAmount = 2 * 1024 * 1024; // above the reliable high-water mark
--- a/src/room/RTCEngine.test.ts
+++ b/src/room/RTCEngine.test.ts
@@ -230,7 +230,7 @@ describe('RTCEngine', () => {
       const send = vi.fn();
       Object.assign(engine as unknown as Record<string, unknown>, {
         ensurePublisherConnected: vi.fn().mockResolvedValue(undefined),
-        waitForBufferStatusLow: vi.fn().mockResolvedValue(undefined),
+        waitUntilBelowHighWaterMark: vi.fn().mockResolvedValue(undefined),
         updateAndEmitDCBufferStatus: vi.fn(),
         dataChannelForKind: vi.fn(() => ({ send })),
         pcManager: {
--- a/src/room/RTCEngine.test.ts
+++ b/src/room/RTCEngine.test.ts
@@ -298,6 +298,68 @@ describe('RTCEngine', () => {
     });
   });
 
+  describe('resendReliableMessagesForResume', () => {
+    class FakeDataChannel extends EventTarget {
+      bufferedAmount = 0;
+
+      bufferedAmountLowThreshold = 64 * 1024;
+
+      send = vi.fn();
+    }
+
+    const tick = () => new Promise<void>((resolve) => setTimeout(resolve, 0));
+
+    it('does not let a concurrent reliable send interleave into the resume replay', async () => {
+      const engine = new RTCEngine(roomOptionDefaults);
+      const dc = new FakeDataChannel();
+      Object.assign(engine as unknown as Record<string, unknown>, {
+        _isClosed: false,
+        ensurePublisherConnected: vi.fn().mockResolvedValue(undefined),
+        dataChannelForKind: vi.fn(() => dc),
+        pcManager: {
+          getMaxPublisherMessageSize: vi.fn(() => 64 * 1024 - 1),
+        },
+      });
+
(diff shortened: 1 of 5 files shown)


src/room/RTCEngine.ts
```diff
@@ -1645,19 +1660,18 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
     kind: Exclude<DataChannelKind, DataChannelKind.RELIABLE>,
     bufferStatusLowBehavior: 'drop' | 'wait' = 'drop',
   ) {
-    // make sure we do have a data connection
-    await this.ensurePublisherConnected(kind);
-
     const dc = this.dataChannelForKind(kind);
```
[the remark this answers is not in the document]
lukasIO: they are, the check lives in `sendDataPacket` already, so this one was just repeating the check

src/room/RTCEngine.test.ts
```diff
@@ -298,6 +299,277 @@ describe('RTCEngine', () => {
...
+        attemptingReconnect: true,
+        ensurePublisherConnected: vi.fn().mockResolvedValue(undefined),
+        dataChannelForKind: vi.fn(() => dc),
+        pcManager: {
+          getMaxPublisherMessageSize: vi.fn(() => 64 * 1024 - 1),
+        },
+      });
+
+      const buffer = (engine as unknown as { reliableMessageBuffer: DataPacketBuffer })
+        .reliableMessageBuffer;
```
1egoman, line 378: nitpick: it might be nice to have some test specific helpers to avoid all the casting to access these private members. Maybe something like `getReliableMessageBuffer(engine)`, `resendReliableMessagesForResume(engine)`, etc

src/room/RTCEngine.test.ts
```diff
@@ -298,6 +299,277 @@ describe('RTCEngine', () => {
...
+      await Promise.all([replay, concurrentSend]);
+
+      expect(dc.send).toHaveBeenCalledTimes(3);
+      expect(dc.send.mock.calls[0][0]).toBe(replayed1);
+      expect(dc.send.mock.calls[1][0]).toBe(replayed2);
+      expect(dc.send.mock.calls[2][0]).not.toBe(replayed1);
+      expect(dc.send.mock.calls[2][0]).not.toBe(replayed2);
+    });
+
+    it('transmits a packet deferred mid-replay instead of marking it sent without sending', async () => {
```
1egoman, line 363: question: To make sure I understand, this test is exercising the case where a new data packet is enqueued when the data channel's fullness is above the high water mark?

If so, a suggested better test name:

```suggestion
    it('transmits a packet deferred mid-replay instead of dropping it due to the data channel buffer being full', async () => {
```

src/room/RTCEngine.test.ts
```diff
@@ -298,6 +299,277 @@ describe('RTCEngine', () => {
...
+      const dc = new FakeDataChannel();
+      const buffer = stubEngine(engine, dc);
+
+      // Park the send on a full buffer, then invalidate the channel (reconnect/replacement).
+      dc.bufferedAmount = 2 * 1024 * 1024;
+      const send = engine.sendDataPacket(makePacket(1), DataChannelKind.RELIABLE);
+      await tick();
+      (
+        engine as unknown as { invalidateDataChannelWaiters: (reason: string) => void }
+      ).invalidateDataChannelWaiters('data channels recreated');
```
1egoman, line 448: question: To make sure I understand properly, this is simulating a full reconnect?

src/room/RTCEngine.test.ts
```diff
@@ -298,6 +299,277 @@ describe('RTCEngine', () => {
...
+        engine.sendDataPacket(makePacket(1), DataChannelKind.RELIABLE),
+      ).resolves.toBeUndefined();
+      expect(dc.send).not.toHaveBeenCalled();
+      expect(buffer.length).toBe(1);
+      expect(buffer.getAll()[0].sent).toBe(false);
+    });
+  });
+
+  describe('sendLossyBytes', () => {
+    it('ensures the publisher is connected before sending (direct data-track path)', async () => {
```
1egoman, line 500: question: I only see one place where `sendLossyBytes(..., ..., 'drop')` is exercised in here (`await engine.sendLossyBytes(new Uint8Array(100), DataChannelKind.LOSSY, 'drop');`), is there more worth adding? (ie, maybe verifying drops occur when the data channel isn't initialized, etc).

src/room/RTCEngine.ts
```diff
@@ -1700,39 +1783,156 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+  /** Rejects all parked headroom waiters (all kinds) and starts a fresh epoch. */
+  private invalidateDataChannelWaiters(reason: string) {
+    for (const controller of this.dataChannelEpochs.values()) {
+      controller.abort(reason);
+    }
+    this.dataChannelEpochs.clear();
+  }
+
+  /** Acquires the per-kind headroom lock, resolving with the unlock function. */
+  private lockBufferHeadroom(kind: DataChannelKind): Promise<() => void> {
```
1egoman: suggestion: Maybe call this `getBufferHeadroomLock` instead and make it return the `Mutex` instead of `Promise<() => void>`? Then make all callers do `.getBufferHeadroomLock().lock()`?

Earlier in the diff I found myself a little confused about what this did and how it differed with `waitForBufferHeadroom`, I think making this method clearly just "return the mutex" would have helped.

src/room/RTCEngine.ts
```diff
@@ -1700,39 +1783,156 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+    }
+  }
+
+  /**
+   * Core wait of {@link waitForBufferHeadroom}. The caller must hold the kind's headroom lock —
+   * batch senders (the resume replay) hold it across all of their sends so no other sender can
+   * interleave, and call this per message to respect flow control within the batch.
+   */
+  private async waitForBufferHeadroomLocked(
+    kind: DataChannelKind,
```
1egoman, line 1894: thought: These two method name I find really confusing - `waitForBufferHeadroomLocked` to me implies that it is acquiring the lock, and in comparison `waitForBufferHeadroom` implies that it isn't. Should these names be flipped around?

src/room/RTCEngine.ts
```diff
@@ -1700,39 +1783,156 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
+        );
+      };
+      const cleanup = () => {
+        dc.removeEventListener('bufferedamountlow', onBufferedAmountLow);
+        dc.removeEventListener('close', onDCClose);
+        epochSignal.removeEventListener('abort', onEpochAbort);
+      };
+      if (epochSignal.aborted) {
+        onEpochAbort();
+        return;
```
1egoman: question: Can you explain what `epoch` means here - is there something you could call this which makes it more clear this is like "signal which fires when data channel closes"? (I'm assuming that's all this is for, but maybe it's more broad?)

src/utils/dataPacketBuffer.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(buffer.getAll().map((i) => i.sequence)).toEqual([1, 2]);
+  });
+
+  it('popToSequence drops acked packets regardless of sent state', () => {
+    const buffer = new DataPacketBuffer();
+    buffer.push(item(1, 100, true));
+    buffer.push(item(2, 100, false));
+    buffer.push(item(3, 100, false));
+
+    buffer.popToSequence(2);
```
1egoman, line 96: nitpick (probably out of scope of this pull request since it was pre-existing): Should this be `unshiftToSequence`? `pop` usually implies the inverse of `push`, but this looks like it's offsetting from the front.

1egoman (commented):
I can't approve officially because I originally opened this pull request, but this looks good ✅ 

Probably as expected due to our fairly extensive conversation on slack about this most of my comments are naming / code clarity related.

src/room/RTCEngine.test.ts
lukasIO: At one point during the work on this PR I was already half way through changing those casts to ts-ignores when I realised that this pattern just emerged because it's how the other tests are doing it as well. 
I'm not convinced that we should make public accessors for all of them, but definitely aligned on changing this current pattern. 
Let's do it in a follow up with a clear concept of how to tackle it as a more general solution

src/room/RTCEngine.test.ts
lukasIO: the refactor PR makes this a bit obsolete, so I'll hold off on adding this here now, but happy to add something in that regard on the refactor PR if you still see a gap there

src/utils/dataPacketBuffer.test.ts
lukasIO: makes sense, yeah let's address this in a follow up

src/room/RTCEngine.test.ts
lukasIO: no, this tests that if another packet is enqueued while resuming (landing in our own intermediate buffer), but the replay has already started, the new packet should either _not_ be marked as sent when the replay finishes

src/room/RTCEngine.test.ts
lukasIO: any condition in which the data channels would get recreated, this can also happen on resume, see https://github.com/livekit/client-sdk-js/blob/cc79fd75e965e5ddc6ffa92c597cd3d4d138dfc9/src/room/RTCEngine.ts#L1511-L1515

src/room/RTCEngine.ts
lukasIO: done!

src/room/RTCEngine.ts
lukasIO: yeah, it's confusing. There's some (debated) convention that the `*locked` suffix indicates it needs to be called _within_ a lock. 

Changed this to be explicit `*withLock` and `*withoutLock`

src/room/RTCEngine.ts
lukasIO: just meant to indicate a generation of data channels. I dropped the `epoch` entirely and just called it `onAbort` now and `waiterAbortController`

lukasIO (approved):

PR #2014, opened 2026-07-16
chore: extract data channel handling into manager class

lukasIO:


src/room/data-channel/FlowControlledDataChannel.ts
```diff
@@ -0,0 +1,216 @@
...
+ * current — replaced or torn down — its events may never fire again, so the owner must call
+ * {@link invalidateWaiters}, which aborts parked waiters (releasing the gate); the next waiter
+ * gets a fresh controller.
+ */
+export class FlowControlledDataChannel {
+  readonly kind: DataChannelKind;
+
+  readonly lowWaterMark: number;
+
+  readonly highWaterMark: number;
```
1egoman, line 40: thought: I don't see any interface overlap between this class and `LossyDataChannel` which is a bit surprising - I would expect this to at least have a `send` method of some sort matching the lossy one? I bring it up because it might be nice to have an abstract class / interface all three are children of to help ensure future api homogeneity.

src/room/data-channel/ReliableDataChannel.ts
```diff
@@ -0,0 +1,153 @@
...
+   * whole replay — releasing it between messages would let a concurrent send (whose newer
+   * sequence was already assigned before it queued on the lock) hit the wire mid-replay, and
+   * receivers would then discard the remaining lower-sequence resent messages as duplicates.
+   */
+  async replay(lastMessageSeq: number) {
+    const dc = this.getChannel();
+    if (!dc) {
+      return;
+    }
+    this.messageBuffer.popToSequence(lastMessageSeq);
```
1egoman, line 103: nitpick: resurfacing this comment: https://github.com/livekit/client-sdk-js/pull/2013#discussion_r3597223480

src/room/data-channel/ReliableDataChannel.ts
```diff
@@ -0,0 +1,153 @@
...
+ * The reliable channel: flow control plus delivery-across-resume semantics.
+ *
+ * Every packet gets a monotonic sequence (stamped into the protobuf by the caller before
+ * serialization, via {@link nextSequence}) and is retained in a replay buffer until the channel's
+ * `bufferedAmount` confirms it has been handed to the transport. Sends that land in a reconnect
+ * window — or whose headroom wait is torn down transiently — are queued unsent and resolve;
+ * {@link replay} delivers them (plus any unacked packets) after a resume. Only an engine close
+ * rejects, because no replay is coming after that.
+ */
+export class ReliableDataChannel extends FlowControlledDataChannel {
```
1egoman, line 27: Very nice that you were able to make this superset of `FlowControlledDataChannel`!

src/room/data-channel/DataChannelManager.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  return { manager, opts, pcManager, created };
+}
+
+describe('DataChannelManager', () => {
+  it('creates and wires the three publisher channels', () => {
+    const { manager, opts, pcManager, created } = makeManager();
+
+    manager.createPublisherChannels(pcManager);
+
+    expect(created.map((dc) => dc.label)).toEqual(['_lossy', '_reliable', '_data_track']);
```
1egoman: nitpick: should this `.sort()` before asserting? Should the test be asserting the order or just that all three were made?

src/room/data-channel/DataChannelManager.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    isReconnecting: () => false,
+    onDataMessage: vi.fn(),
+    onDataTrackMessage: vi.fn(),
+    onDataError: vi.fn(),
+    onChannelClose: vi.fn(),
+    onBufferStatusChanged: vi.fn(),
+    ...overrides,
+  };
+  const manager = new DataChannelManager(opts);
+  const created: FakeDataChannel[] = [];
```
1egoman: suggestion: You might consider making `created` some sort of map / object with `dc.label` as the key. Then code like `created.find((dc) => dc.label === '_reliable')!.bufferedAmount = ...` could become `created['_reliable'].bufferedAmount = ...` or also the earlier nitpick I made regarding `.sort()` could maybe be done more elegantly by checking `Object.keys(...)`.

src/room/data-channel/FlowControlledDataChannel.test.ts
```diff
@@ -0,0 +1,150 @@
...
+class FakeDataChannel extends EventTarget {
+  bufferedAmount = 0;
+
+  bufferedAmountLowThreshold = 64;
+}
+
+const tick = () => new Promise<void>((resolve) => setTimeout(resolve, 0));
+
+function makeChannel(opts?: { dc?: FakeDataChannel | undefined; engineClosed?: boolean }) {
+  const dc = 'dc' in (opts ?? {}) ? opts!.dc : new FakeDataChannel();
```
1egoman: nitpick: This is a bit of a weird way do do this, couldn't this just be:


```suggestion
  const dc = opts?.dc ?? new FakeDataChannel();
```

src/room/data-channel/FlowControlledDataChannel.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    kind: DataChannelKind.RELIABLE,
+    lowWaterMark: 64,
+    highWaterMark: 1024,
+    isEngineClosed: () => state.engineClosed,
+    onBufferStatusChanged,
+  });
+  if (dc) {
+    channel.attach(dc as unknown as RTCDataChannel);
+  }
+  return { channel, dc: dc as FakeDataChannel, state, onBufferStatusChanged };
```
1egoman, line 32: nitpick: Also not exactly sure why you need this cast here since `dc` is already `FakeDataChannel`, maybe I'm missing something though:


```suggestion
  return { channel, dc, state, onBufferStatusChanged };
```

src/room/data-channel/FlowControlledDataChannel.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    dc.bufferedAmount = 512; // between low (64) and high (1024)
+    expect(channel.isBelowHighWaterMark(handle)).toBe(true);
+    expect(channel.isBelowLowWaterMark(handle)).toBe(false);
+
+    dc.bufferedAmount = 2048;
+    expect(channel.isBelowHighWaterMark(handle)).toBe(false);
+  });
+
+  it('waiting for headroom without a handle rejects with a connection error', async () => {
```
1egoman, line 78: These tests are really easy to understand, amazing job!

src/room/data-channel/LossyDataChannel.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    expect(stats.dropCount).toBe(1);
+    expect(stats.statCurrentBytes).toBe(0);
+  });
+
+  it('wait policy: parks above the high-water mark instead of dropping', async () => {
+    const { channel, dc } = makeChannel('wait');
+    dc.bufferedAmount = 2048;
+    const resolved = vi.fn();
+    const send = channel.send(new Uint8Array(10)).then(resolved);
+    await vi.advanceTimersByTimeAsync(0);
```
1egoman, line 70: question: What's the difference between this and the `await tick()` you were using in other tests in practice (I realize one is mocking timers and one isn't, but does that matter here? I'm assuming vitest is calling these "next tick" deferred handlers on the current tick to implement this?), and does it make sense to standardize on one or the other?

1egoman (approved):

src/room/data-channel/FlowControlledDataChannel.ts
lukasIO: it's a fair consideration. both lossy and reliable extend this one, so things like refreshBufferStatus/attach/detach are shared. `send` isn't because the signature for the reliable one is slightly different (we also pass a sequence number on the reliable channel) and it would arguably become a bit awkward if we had a shared base signature that looks like `send(bytes, sequence?)`. There's also no call site that acts on `FlowControlledDataChannel` as a base, so we always know in code which kind we're dealing with.

src/room/data-channel/FlowControlledDataChannel.test.ts
lukasIO: yeah, that's a lot nicer!

there's a slight difference though, where `opts: { dc: undefined }` would become identical to no opts being passed. solved it by using `null` as the explicit pass option.

src/room/data-channel/LossyDataChannel.test.ts
lukasIO: good question, dove a bit deeper into the _why_:

the main consideration is that `advanceTimersByTimeAsync` is only used in files where we use fake timers. Here we need those for the dynamic threshold scheduling. 

the `tick` method uses real timers and it would simply hang if we were using it in a file with fake timers. 
the other way around, using `advanceTimersByTimeAsync` in the other files would mean we have to do manual timer book keeping with fake timers.

src/room/data-channel/FlowControlledDataChannel.test.ts
lukasIO: the cast is only for type checking in the tests to work as expected. the `dc` could be passed as `null` for checking the no-handle cases. The return is cast to simply not-be-null

src/room/data-channel/DataChannelManager.test.ts
lukasIO: converted it into a record which should also address this

PR #2015, opened 2026-07-16
Add publish-side frame metadata user_data support

### LINKED ISSUE #2012: Frame metadata: no publish-side API to attach user_data from the JS SDK
### Describe the feature

`livekit-client` can publish frame-metadata trailers with `user_timestamp` and `frame_id`, and (since #1983, shipped in `livekit-client@2.20.1`) it can now *parse* a `user_data` trailer (tag `0x03`) on the receive side. However, there is still no way to **write** `user_data` on the **publish** side from the JS SDK.

- `FrameMetadataPublishOptions` only exposes `timestamp` and `frameId`.
- The encode transform (`appendPacketTrailerToEncodedFrame` in `src/frameMetadata/frameMetadata.ts`) only ever writes the timestamp + frameId TLVs; there is no plumbing to attach an arbitrary per-frame `user_data` payload.
- By contrast, the Rust SDK supports this today via `VideoFrame.frame_metadata.user_data` passed to `capture_frame` (livekit/rust-sdks#890), and the blog post (https://livekit.com/blog/frame-metadata) advertises `user_data` as a first-class capability across SDKs.

So JS clients can *read* `user_data` published by Rust/other SDKs, but a JS publisher cannot *emit* it.

### Use case

Frame-accurate correlation of application events with the published video: attach a small `user_data` payload (e.g. an event id / tag) to the frame that was live when the event fired, so a downstream recording/analysis pipeline can line up video with application state with no separate sync channel — exactly the "frame-accurate correlation" use case from the frame-metadata blog post.

Today this requires reimplementing the SDK's encoded-frame transform in a custom worker just to add the `user_data` TLV (reusing the exact `LKTS` XOR-TLV wire format), which duplicates SDK internals and risks drifting from the wire format across releases.

### Proposal

Expose a publish-side `user_data` mechanism, e.g.:
- a `userData` flag in `FrameMetadataPublishOptions` (to advertise the feature on the track), and
- an API to set the `user_data` for upcoming frame(s), e.g. `LocalVideoTrack.setFrameUserData(bytes: Uint8Array)` (attach to the next encoded frame, or valid until replaced), mirroring the Rust `capture_frame` model.

### Questions

- Was the publish-side `user_data` writer intentionally deferred, or is it planned as a follow-up to #1983?
- Is there an issue/PR already tracking this that I missed?

### SDK
`livekit-client@2.20.1`

mcuelenaere:
Fixes #2012

## What

Adds the missing publish-side counterpart to the receive-side `user_data` parsing introduced in #1983:

- **`FrameMetadataPublishOptions.userData?: boolean`** — opt-in flag that advertises `PTF_USER_DATA` on the published track.
- **`LocalVideoTrack.attachUserDataToNextFrame(userData?: Uint8Array)`** — attaches the bytes to the next encoded frame as a `0x03` user_data TLV in the `LKTS` frame trailer.
- **`getMaxFrameUserDataLength(options)`** (exported) — computes the payload cap implied by the single-byte trailer length (248 bytes for user_data alone, 232 alongside timestamp + frameId).

## Semantics

`attachUserDataToNextFrame` is **one-shot**: the value is written once to the next encoded frame of *each* active encode pipeline of the track (primary and backup-codec senders), then cleared. Calling again before a frame is encoded replaces the pending value (last write wins); `undefined`/empty clears it. Oversized payloads throw a `RangeError` at the API; the wire encoder additionally drops an overflowing TLV defensively since frame pipelines must never throw. Empty (DTX-like) frames never consume the pending value.

## How

- `appendPacketTrailer` gained a `userData` parameter emitting the `0x03` TLV byte-exactly matching the existing parser (tag/len/value all XOR `0xFF`).
- **Non-E2EE**: encode pipelines in the frame metadata worker were previously anonymous; they are now keyed by `track.mediaStreamID` in a new `EncodePipelineRegistry`, and a new `setFrameUserData` worker message routes pending bytes to every pipeline of the track. A small stash covers the `RTCRtpScriptTransform` registration race (`onrtctransform` arrives on a different channel than `postMessage`). `RTCEngine.setupFrameMetadataSender` registers a poster on the track that forwards the bytes.
- **E2EE**: `FrameCryptor` stores the pending value and writes it after encryption (trailer stays outside the ciphertext, matching extract-before-decrypt on receive). The `setFrameUserData` message uses `getTrackCryptor`'s create-if-missing semantics, so data posted before the encode transform registers is kept — and `setFrameMetadataOpts` deliberately does not clear it.
- `@livekit/protocol` bumped 1.46.6 → 1.49.0, the first 1.x release with `PTF_USER_DATA`.
- Demo: new "Attach User Data" checkbox + send box; the receive side already renders user_data since #1983.

## Notes for reviewers

- Older receivers gate extraction on `packetTrailerFeatures.length > 0`, which a userData-only track satisfies (proto3 open enums preserve the unknown value), and SDKs ≥ 2.20.1 already parse tag `0x03`. Forwarding of the new enum value by older SFU builds is worth a sanity check; combining `userData` with `timestamp`/`frameId` is the safe fallback.
- 35 new tests: encoder round-trips against the reference trailer builder, feature mapping, registry one-shot/fan-out/race/teardown behavior, FrameCryptor E2EE paths (including the passthrough path and pending-survives-setup regression), `LocalVideoTrack` validation, and `RTCEngine` poster wiring.
- Also smoke-tested the built `livekit-client.fm.worker.mjs` in a browser by transferring streams into it and driving the message protocol end-to-end.

### DIFF
--- a/.changeset/publish-frame-user-data.md
+++ b/.changeset/publish-frame-user-data.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
+Add publish-side frame metadata user_data support: enable it with `FrameMetadataPublishOptions.userData` and attach per-frame data via `LocalVideoTrack.attachUserDataToNextFrame()`
--- a/examples/demo/demo.ts
+++ b/examples/demo/demo.ts
@@ -106,8 +106,9 @@ function getFrameMetadataPublishOptions() {
   const enabled = (<HTMLInputElement>$('frame-metadata')).checked;
   const timestamp = enabled && (<HTMLInputElement>$('frame-metadata-timestamp')).checked;
   const frameId = enabled && (<HTMLInputElement>$('frame-metadata-frame-id')).checked;
+  const userData = enabled && (<HTMLInputElement>$('frame-metadata-user-data')).checked;
 
-  return timestamp || frameId ? { timestamp, frameId } : undefined;
+  return timestamp || frameId || userData ? { timestamp, frameId, userData } : undefined;
 }
 
 function syncFrameMetadataPublishOptions(room = currentRoom) {
@@ -134,14 +135,18 @@ function syncFrameMetadataFeatureControls() {
   const featureControls = $('frame-metadata-features');
   const timestamp = <HTMLInputElement>$('frame-metadata-timestamp');
   const frameId = <HTMLInputElement>$('frame-metadata-frame-id');
+  const userData = <HTMLInputElement>$('frame-metadata-user-data');
 
   featureControls.style.display = enabled ? 'block' : 'none';
   timestamp.disabled = !enabled;
   frameId.disabled = !enabled;
+  userData.disabled = !enabled;
   if (!enabled) {
     timestamp.checked = false;
     frameId.checked = false;
+    userData.checked = false;
   }
+  $('frame-user-data-controls').style.display = userData.checked ? 'flex' : 'none';
   syncFrameMetadataPublishOptions();
 }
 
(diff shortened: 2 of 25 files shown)


1egoman:
Closing per discussion in https://github.com/livekit/client-sdk-js/issues/2012

PR #2017, opened 2026-07-21
Add a drop-down option to choose back up codec.

boks1971:
Not the prettiest (in general, feels like the options area can use a re-work to categorise, actually the buttons area too). But, adding this as I was testing this bug report for server (https://github.com/livekit/livekit/issues/4695)


lukasIO (approved):
lgtm!

PR #2019, opened 2026-07-24
fix: cancel pending onTrack callbacks when subscription fails

lukasIO:


1egoman:
I think I get the general idea and it makes sense, I'll let @boks1971 provide the actual review unless I am tagged explicitly otherwise.

src/room/utils.ts
```diff
@@ -867,3 +867,16 @@ export function isPublisherOfferWithJoinSupported() {
...
+  stream: MediaStream,
+): Track.SID | undefined {
+  const [, streamId] = unpackStreamId(stream.id);
+  if (streamId?.startsWith('TR')) {
+    return streamId;
+  }
+  if (mediaTrack.id.startsWith('TR')) {
+    return mediaTrack.id;
+  }
+}
```
boks1971, line 882: Does this need single peer connection handling to get the track ID from the mapping?

boks1971 (approved):
lgtm! One comment on track SID.

I do not understand the test though. I just cannot wrap my head around what is definition what is implementation :-)

xianshijing-lk (approved):

src/room/utils.ts
lukasIO: this still happens later on at [this line](https://github.com/livekit/client-sdk-js/pull/2019/changes#diff-8f1faf7d794e3b6ef11baf7dc189141bc77dee7e0a1ab9fc8b3a09e22640060bR1666) unless I'm missing something?

lukasIO:
> I do not understand the test though. I just cannot wrap my head around what is definition what is implementation :-)

fair point. It's quite contrived due to synthesising a lot of state for this specific case.
The only goal of that test is to ensure that the `onTrackAdded` isn't invoked if we receive a subscription failure before the room is completely connected. 
I threw out a bunch of the synthetisation and let the test be driven by the media track added event on the engine now.

src/room/utils.ts
boks1971: Not able to see which line you are referring to. Guessing it is in this PR (that line), but for some reason GH is not showing the right line.

boks1971: oh, maybe `unpackStreamId` does it?

boks1971: Nah, that is a standalone function and does not have context.

boks1971: Was referring to this part in Room.ts - https://github.com/livekit/client-sdk-js/blob/eb728ef843cb83ad2553f0487bbaba3804f68360/src/room/Room.ts#L1646-L1659

lukasIO: yeah, sorry, that's the line I tried to link to, but it's outside of the regular diff view :/ 

that path is unchanged. `trackId` extraction still falls back to the `mediastreamtrack.id` so the path that gets the id from `this.engine.getTrackIdForReceiver(..)` should be unchanged, I think?

boks1971: a ha got it, my bad, I did not realise it is in the same function 🤦

lukasIO: my bad for not outlining the intent of the refactor changes in the PR description. Will do better next time

PR #2027, opened 2026-07-28
Downgrade sdp transform

lukasIO:
reverts #2023 due to a regex bug that stops parsing msids at `|` (which we use as a separator). 

Also adds a rule to the renovate config to not update sdp-transform

### DIFF
--- a/src/room/PCTransport.test.ts
+++ b/src/room/PCTransport.test.ts
@@ -1,6 +1,10 @@
 import { type MediaDescription, parse } from 'sdp-transform';
 import { describe, expect, it } from 'vitest';
-import { conformBundledCodecFmtp, placeholderMidsFromTransceivers } from './PCTransport';
+import {
+  applyVideoStartBitrate,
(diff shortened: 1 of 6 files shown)


src/room/PCTransport.test.ts
```diff
@@ -56,36 +52,6 @@ a=recvonly
 a=rtpmap:49 H265/90000
 a=fmtp:49 level-id=180;profile-id=1;tier-flag=0;tx-mode=SRST`;
 
-describe('video start bitrate', () => {
```
boks1971, line 59: is this a straight revert too? Feels like having this as a function + testable maybe useful?

boks1971 (approved):

src/room/PCTransport.test.ts
lukasIO: yeah, I intorduced that helper in the sdp-transform update PR, but good call out, will leave this in

PR #2036, opened 2026-08-04
fix: exact-token opus fmtp matching and case-insensitive opus lookup

lukasIO:
## What

Two small correctness fixes in the SDP munging code (`PCTransport.ts`), surfaced by comparing behavior against the Swift SDK's new SDP module ([client-sdk-swift#1078](https://github.com/livekit/client-sdk-swift/pull/1078)):

1. **Exact-token fmtp parameter matching.** Opus fmtp parameters were matched with a substring `.includes()` check. Because `stereo=1` is a substring of `sprop-stereo=1`, an offer that already carried `sprop-stereo=1` (but not `stereo=1`) would incorrectly be treated as already-stereo, and `stereo=1` would never be added. Introduced a `fmtpConfigHasParam(config, param)` helper that splits on `;` and compares whole tokens, and used it in both `ensureAudioNackAndStereo` and `extractStereoAndNackAudioFromOffer`.
2. **Case-insensitive opus lookup (RFC 4855).** The opus rtpmap lookup used `rtp.codec === 'opus'` (exact, case-sensitive), while every video/bitrate codec lookup in the same file already matches case-insensitively. rtpmap encoding names are case-insensitive per RFC 4855, so an `OPUS`/`Opus` codec name would be missed. Both opus lookups now lowercase before comparing.

## Why

These are latent correctness/spec-compliance bugs. Real-world impact is low today (browsers emit lowercase `opus` and typically both/neither stereo parameter together), but the matching logic was wrong and inconsistent with the rest of the file. This aligns the JS munging behavior with the Swift SDK's exact-token, uniformly-case-insensitive approach.

## Notes for reviewers

- The two internal audio munge functions (`ensureAudioNackAndStereo`, `extractStereoAndNackAudioFromOffer`) are now exported with `@internal`, consistent with the other tested helpers in this file, so the new behavior can be tested directly.
- Added 8 unit tests in `PCTransport.test.ts` covering the `sprop-stereo=1` vs `stereo=1` distinction, no-duplicate behavior, and uppercase `OPUS`.

### DIFF
--- a/src/room/PCTransport.ts
+++ b/src/room/PCTransport.ts
@@ -744,7 +744,19 @@ export default class PCTransport extends (EventEmitter as new () => TypedEmitter
   }
 }
 
-function ensureAudioNackAndStereo(
+/**
+ * Checks whether an fmtp config declares `param` as an exact, `;`-delimited
+ * token. A plain substring check conflates distinct opus parameters — e.g.
+ * `stereo=1` is a substring of `sprop-stereo=1` — so `param` must match a whole
+ * parameter, not appear anywhere within the config string.
+ * @internal
+ */
+export function fmtpConfigHasParam(config: string, param: string): boolean {
+  return config.split(';').some((entry) => entry.trim() === param);
+}
+
+/** @internal */
+export function ensureAudioNackAndStereo(
   media: {
     type: string;
     port: number;
@@ -760,7 +772,8 @@ function ensureAudioNackAndStereo(
   // found opus codec to add nack fb
   let opusPayload = 0;
   media.rtp.some((rtp): boolean => {
-    if (rtp.codec === 'opus') {
+    // rtpmap encoding names are case-insensitive (RFC 4855)
+    if (rtp.codec.toLowerCase() === 'opus') {
       opusPayload = rtp.payload;
       return true;
     }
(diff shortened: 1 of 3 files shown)


src/room/PCTransport.ts
```diff
@@ -744,7 +744,19 @@ export default class PCTransport extends (EventEmitter as new () => TypedEmitter
...
-function ensureAudioNackAndStereo(
+/**
+ * Checks whether an fmtp config declares `param` as an exact, `;`-delimited
+ * token. A plain substring check conflates distinct opus parameters — e.g.
+ * `stereo=1` is a substring of `sprop-stereo=1` — so `param` must match a whole
+ * parameter, not appear anywhere within the config string.
+ * @internal
+ */
+export function fmtpConfigHasParam(config: string, param: string): boolean {
+  return config.split(';').some((entry) => entry.trim() === param);
```
boks1971, line 755: just for bullet-proofing, should this be case insensitive too?

boks1971 (approved):

src/room/PCTransport.ts
lukasIO: hm, good question. I can add it for good measure, for codec names the spec says specifically they need to be case insensitive, but can't find anything like that for the params.

If we do this, we should do it consistently also in the swift SDK and in other SDKs. Let me know if you think that would be a good change in general

boks1971: No, we can leave it. 

We can add it if there is a spec that mentions it. I have never seen it mixed case.

PR #2039, opened 2026-08-06
fix: emit TrackStreamStateChanged again (regression since v2.16.0)

tomkail:
### Summary

`RoomEvent.TrackStreamStateChanged` and `ParticipantEvent.TrackStreamStateChanged` have not fired since v2.16.0.

`Room.handleStreamStateUpdate` writes the new stream state and *then* compares it against the field it just wrote:

```ts
const newStreamState = Track.streamStateFromProto(streamState.state);
pub.track.setStreamState(newStreamState);          // assigns Track._streamState
if (newStreamState !== pub.track.streamState) {    // getter returns what was just assigned
```

`Track.streamState` is a getter over `_streamState` and `setStreamState()` assigns that field, so the condition is always false and the body is unreachable.

### How it regressed

#1199 added the "only emit on an actual change" guard, and got the ordering right — it compared before assigning:

```ts
const newStreamState = Track.streamStateFromProto(streamState.state);
if (newStreamState !== pub.track.streamState) {
  pub.track.streamState = newStreamState;
  ...
}
```

#1625 later introduced `Track.setStreamState()` and swapped the bare field assignment for a call to it, but hoisted the call above the guard it was supposed to be inside. First released in v2.16.0 and still present on `main` (v2.21.0).

### Fix

Read the previous value before overwriting it. This restores the event while preserving #1199's intent: a redundant update (e.g. `ACTIVE` while already active) still emits nothing.

### Tests

Two cases added to `src/room/Room.test.ts`, driving `handleStreamStateUpdate` with a real `RemoteParticipant` / `RemoteTrackPublication` / `RemoteVideoTrack`:

- pausing a subscribed track emits on both the room and the participant — **fails on `main`**, passes with this change
- a redundant `ACTIVE` update emits nothing — passes either way, and guards against "fixing" this by deleting the check

`pnpm test`, `pnpm type:check` and `pnpm format:check` are clean; `pnpm lint` reports no new warnings.

### Impact

Consumers currently have no way to detect the SFU pausing a subscribed track under bandwidth pressure, which is the documented signal for diagnosing a frozen remote video.

### DIFF
--- a/.changeset/olive-moons-repeat.md
+++ b/.changeset/olive-moons-repeat.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+fix: emit TrackStreamStateChanged again by reading the previous stream state before overwriting it
--- a/src/room/Room.test.ts
+++ b/src/room/Room.test.ts
@@ -1,13 +1,22 @@
 import {
   ClientInfo_Capability,
   JoinResponse,
+  StreamState as ProtoStreamState,
+  StreamStateUpdate,
   SubscriptionError,
   SubscriptionResponse,
+  TrackInfo,
+  TrackType,
 } from '@livekit/protocol';
 import { afterEach, describe, expect, it, vi } from 'vitest';
+import MockMediaStreamTrack from '../test/MockMediaStreamTrack';
 import Room, { ConnectionState } from './Room';
 import { roomConnectOptionDefaults, roomOptionDefaults } from './defaults';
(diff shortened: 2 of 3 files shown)


1egoman:
@tomkail Thanks for the contribution! This generally makes sense to me, would you be able to click the link in the comment above this one and accept the CLA? Once you have I'll get this merged.

tomkail:
Done! Ta!

1egoman (approved):

PR #2043, opened 2026-08-13
fix: ensure buffered events are flushed on signal reconnection

lukasIO:


src/room/Room.ts
```diff
@@ -2438,7 +2443,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
       .on(
         ParticipantEvent.TrackUnsubscribed,
         (track: RemoteTrack, publication: RemoteTrackPublication) => {
-          this.emit(RoomEvent.TrackUnsubscribed, track, publication, participant);
+          this.emitWhenConnected(RoomEvent.TrackUnsubscribed, track, publication, participant);
```
lukasIO: this was asymmetric before wrt TrackSubscribed. Now both events are buffered when not connected

src/room/Room.ts
```diff
@@ -639,7 +639,6 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
         this.emitBufferedEvents();
       })
       .on(EngineEvent.SignalResumed, () => {
-        this.bufferedEvents = [];
```
lukasIO, line 642: this is the main bug. The buffered events should not have been cleared on SignalResumed. 

Instead we're now flushing them

src/room/Room.ts
xianshijing-lk: Nice, that makes sense, do we need to use emitWhenConnected for other events as well ?and does emitWhenConnected make sure all the events will be emitted in the right orders ?

lukasIO: had to revert this, as it was leading to the events getting lost during teardown. 
There's still a test in https://github.com/livekit/client-sdk-js/pull/2044 that ensures correct behaviour here

xianshijing-lk (approved):

PR #2044, opened 2026-08-13
fix: e2ee setup hardening and tests

lukasIO:
Follow up to #2043 

Fixes E2EE subscribers rendering a permanently black tile with nothing logged.

Three independent defects, each of which could leave a subscribed, encrypted track with no live decrypt transform:

- **Transceiver reuse** — a reused receiver keeps its already-transferred encoded streams (they can't be transferred twice, and createEncodedStreams() is once-per-receiver), so the new trackId used to spawn a fresh cryptor with no pipeline while the old one kept piping with no participant. Reuse now carries previousTrackId and re-points the existing cryptor, rebuilding its pipe if it died.
- **Decode failed open** — with unknown encryption state the cryptor handed raw ciphertext to the decoder. It now drops instead, and reports it when the track is genuinely live (routine teardown stays quiet).
- **Silent failure** — a watchdog reports an encrypted track left without a decrypt transform, so this class of bug is no longer invisible.

Also removes the now-dead isReuse flag and hardens pipeThrough against synchronous throws.


src/e2ee/worker/FrameCryptor.ts
```diff
@@ -167,10 +188,12 @@ export class FrameCryptor extends BaseFrameCryptor {
...
-    if (this.currentTransform) {
-      this.currentTransform = undefined;
-    }
-
+    // NOTE: deliberately does not clear `currentTransform`. Nothing here cancels
+    // the pipe, so the transform really is still running and clearing it would
+    // desync our bookkeeping from reality -- which previously let a later
+    // resubscribe skip setup while frames kept flowing through a cryptor with no
+    // participant assigned.
+    clearTimeout(this.undecryptedTrackTimeout);
```
1egoman, line 196: nitpick: Maybe clear `this.undecryptedTrackTimeout` explictly here so there's not a no longer active timeout id stored? That could be a source of more bugs in the future.

There's another place in `scheduleUndecryptedTrackWatchdog` where I think it would be good to clear it too, before the `if (operation !== 'decode') {` early bail out.

src/e2ee/E2eeManager.ts
```diff
@@ -544,18 +562,27 @@ export class E2EEManager
...
+            // @ts-ignore
+            previousTrackId: receiver[E2EE_TRACK_ID],
             codec,
             participantIdentity,
             hasPacketTrailer,
           },
         };
         this.worker.postMessage(msg);
+        // @ts-ignore
+        receiver[E2EE_TRACK_ID] = trackId;
```
1egoman: question: What's up with these `@ts-ignore`s? Is it worth making the if condition earlier on `if (E2EE_FLAG in receiver && E2EE_TRACK_ID in receiver) {` to drop them?

1egoman (approved):
Generally I can follow what's going on, but I've yet to dig really deep into e2ee so my review is fairly high level

src/e2ee/E2eeManager.ts
lukasIO: makes sense!

src/e2ee/worker/FrameCryptor.ts
lukasIO: good idea!

PR #2046, opened 2026-08-13
chore: initial state machine plumbing

lukasIO:


src/api/SignalClient.ts
```diff
@@ -118,6 +138,20 @@ export enum SignalConnectionState {
   DISCONNECTED,
 }
 
+/**
+ * Public projection of the lifecycle machine's states. `new`, `offline` and `closed` are all
+ * reported as `DISCONNECTED`: they differ in what may happen next
+ */
+const lifecycleToConnectionState: Record<SignalLifecycleState, SignalConnectionState> = {
```
1egoman: nitpick: should this constant be in `SCREAMING_SNAKE_CASE`? There's a few other constants as well like `signalStates` which I think this also applies to.


```suggestion
const LIFECYCLE_TO_CONNECTION_STATE: Record<SignalLifecycleState, SignalConnectionState> = {
```

src/api/SignalClientStateMachine.ts
```diff
@@ -0,0 +1,155 @@
...
+const signalStates = {
+  new: {
+    connect: startConnect,
+    reconnect: startReconnect,
+    close: requestClose,
+  },
+  connecting: {
+    connect: startConnect,
+    reconnect: startReconnect,
+    connectComplete: 'connected',
```
1egoman: nitpick: you may consider breaking out some of these repeated states instead into some more "component level" objects which can be spread in rather than breaking out hte functions individually. As is I find myself going back and forth a lot between function and state transition map when reasoning about this.


```suggestion
const common = {
    connect: on<'connect'>(({ ctx }) => {
      ctx.attemptId += 1;
      ctx.lastError = undefined;
      return 'connecting';
    }),
    reconnect: on<'reconnect'>(({ ctx }) => {
      ctx.attemptId += 1;
      ctx.lastError = undefined;
      return 'reconnecting';
    }),
    close: on<'close'>(({ ctx }, event) => {
      ctx.closeReason = event.reason;
      return 'disconnecting';
    }),
};

const signalStates = {
  new: common,
  connecting: {
    ...common, // etc
    connectComplete: 'connected',
```

1egoman (approved):
This generally looks like a great start to me!

One thing I would find really helpful would be some sort of way to visualize the state machine. If there isn't a tool already which can read the machina state machine and generate a mermaid / graphviz / etc type visualization from it, I suspect this would be pretty easy to add. I'm thinking some sort of small script which could import in the `SignalClient` (and in the future, other state machines all combined together), convert to mermaid or graphviz, then generate a big svg. This is also something a LLM could use to generate diagrams for describing potential changes.

The dream / probably a stretch goal would be some way for the example app to be able to show the state machine live so as you test out scenarios you can actually watch the "active" state propagate through the state machine interactively. At a past job I did this for a large distributed state machine which was not too dissimilar from livekit's domain and it proved invaluable both for engineering understanding and explaining failure modes to less technical folks.

src/api/SignalClient.ts
lukasIO: makes sense, I refactored it to be a function which avoids the awkwardness of the constant being a record and feels cleaner in general

src/api/SignalClientStateMachine.ts
lukasIO: not opposed to that, what I like more about the current way of defining it, is that all possible transitions are directly visible within the definition. Spreading a `common` set would make that opaque. I'll merge it like it is for now, but very open to discuss this more

lukasIO:
> One thing I would find really helpful would be some sort of way to visualize the state machine. 

added via `pnpm examples:machine`

examples/state-machine/README.md
```diff
@@ -0,0 +1,48 @@
+# machina inspector
+
+Renders a [machina](https://machina-js.org) FSM and lets you drive it by hand: fire inputs, watch
+the state move, and see what the machine does with an input it does not want.
+
+```bash
+pnpm examples:machine
+```
```
1egoman, line 8: thought: This is definitely useful and I think is a fine start if you want to leave it here for now!

But to clarify my previous suggestion - I think this would be a lot more valuable if it were actually hooked up to an actual client sdk. So instead of manually triggering `connect` / `connectComplete` / etc events, you could connect a client, test different scenarios (either via "simulate scenario", or by actually causing them), and watch the state machine to make sure it does the "right thing" in each case. Especially as more of the connection management logic moves into explicit state machines and the whole thing grows, I think a tool like this will be required for effective debugging.

src/api/SignalClient.ts
```diff
@@ -371,34 +475,42 @@ export class SignalClient {
...
+        // only the `opened` path can classify it (a 401 is known after asking the validate endpoint).
+        // Noting the failure *before* the close handler is registered means it runs first, so a close
+        // that merely reflects a failed upgrade stands down and lets the classified error reach the
+        // caller. Any other close still fails the attempt straight away: after a successful upgrade
+        // nothing else would, short of the first-message timeout.
+        let upgradeFailed = false;
+        this.ws.opened.catch(() => {
+          upgradeFailed = true;
+        });
+
```
1egoman, line 520: thought: Seeing this makes me wonder if it's worth there being distinct events for each type of "close", or maybe some sort of enum value on the event which captures a "close reason". It doesn't seem like it's required for a logic specific reason today but I think the two big advantages are 1) capturing context in events for debugging, and 2) building ahead a bit for when there inevitably will need to be different logic in different failure modes.

src/api/SignalClientStateMachine.test.ts
```diff
@@ -0,0 +1,501 @@
...
+function send(machine: SignalMachine, input: SignalMachineInput) {
+  machine.handle(input.type, input);
+}
+
+/** Drives a machine into `state` through legal inputs only. */
+function machineIn(state: SignalLifecycleState): SignalMachine {
+  const machine = createSignalMachine();
+  switch (state) {
+    case 'new':
+      break;
```
1egoman: question: Is this really the best way to initialize a state machine at an arbitrary state? I know with xstate there is a way that you can start a machine at any arbitrary state as a one-liner - a quick search points me at [machine.resolveState](https://xstate.js.org/api/interfaces/statemachine.html#resolvestate).

Is there something similar with machina? If not, I wonder if you can set some internal variable on the machina state machine class to emulate this. The concern I have with what you have here is that all these cases are something we are going to have to keep up to date, and especially as the state machine grows, that is going to become quite burdensome.

src/api/SignalClientStateMachine.test.ts
```diff
@@ -0,0 +1,501 @@
...
+ * Every state/input pair and what it resolves to. Asserted exhaustively below, and used as the
+ * oracle for the random walks, so adding a state or an input cannot silently leave a hole.
+ */
+const documentedTransitions: Record<string, Record<string, TransitionTarget>> = {
+  // Establishing a session is legal exactly where no transport and no attempt are in play: `new`,
+  // `offline`, `closed`. Elsewhere it is a caller error, and `SignalClient` refuses rather than
+  // opening a transport the lifecycle would not own — except from `disconnecting`, which it waits out.
+  new: {
+    connect: 'connecting',
+    close: 'disconnecting',
```
1egoman, line 38: question: How useful is this in the tests? Like my previous comment, I worry about keeping this up to date, especially since these are `Record<string, ..>`s (so there won't be a type error if new states are added). It looks like these are only used in a few tests which validate that the whole state machine exactly matches with all transitions lining up, but IMO, I'm not sure that's all that useful. I think the most valuable tests are ones which start at a state, and drive it through a series of steps and make sure that the state is correct at each step and that any side effects have been accomplished.

src/api/SignalClientStateMachine.test.ts
```diff
@@ -0,0 +1,501 @@
...
+    it('ignores a transport that has already been replaced', () => {
+      const machine = machineIn('connected');
+      const staleAttemptId = machine.context.attemptId;
+
+      // a newer attempt takes over, then the old socket reports its close
+      send(machine, { type: 'reconnect' });
+      send(machine, { type: 'transportFailed', attemptId: staleAttemptId, reason: 'late close' });
+
+      expect(machine.currentState()).toBe('reconnecting');
+    });
```
1egoman, line 220: This is I think the most useful type of test here!

One small thought though - I think something higher level than this (ie, more like an integration test where you are interacting with either `SignalClient` or the SDK's user facing interface) would be even more valuable. As is, I think it's more likely that some side effect code around the state machine is accidentally broken than removing a state transition itself. So ie, maybe add some assertions around the socket state here, etc.

Or maybe the thing to do here is to eventually add some internal state machine state assertions (probably by reaching into some private members in a class) to the e2e tests to make sure it's going down the right paths.

src/api/SignalClientStateMachine.test.ts
```diff
@@ -0,0 +1,501 @@
...
+
+      send(machine, { type: 'reconnectFailed', recoverable: true });
+      send(machine, { type: 'connect' });
+      expect(machine.context.attemptId).toBe(first + 2);
+    });
+  });
+
+  describe('close', () => {
+    it('is legal from every state that owns a transport or an attempt', () => {
+      for (const state of ['new', 'connecting', 'connected', 'offline', 'reconnecting'] as const) {
```
1egoman, line 320: thought: I wonder if there's some way you could make these lists exhaustive (is there some way to determine statically whether a state has a transport associated?), so that when a new entry is added, there's a type error that ensures that somebody also adds the new entry to this list.

src/api/SignalClientStateMachine.test.ts
```diff
@@ -0,0 +1,501 @@
...
+            documentedTransitions[previousState]?.[input],
+          ];
+          if (!allowed.includes(actual)) {
+            throw new Error(
+              `${previousState} --${input}--> ${state}, but the table documents ${allowed.join(' or ')}`,
+            );
+          }
+        },
+      });
+    });
```
1egoman, line 416: nitpick: Though these seem to have seeds, it does somewhat worry me that code doing something non deterministic is in a unit test.

src/api/SignalClientStateMachine.ts
```diff
@@ -0,0 +1,180 @@
...
+  | { type: 'connectFailed'; error?: unknown }
+  | { type: 'reconnectComplete'; attemptId: number }
+  /**
+   * A resume attempt ended. `recoverable` distinguishes "another resume may follow" (→ `offline`)
+   * from a terminal outcome such as a server leave or an expired token (→ `closed`).
+   */
+  | { type: 'reconnectFailed'; error?: unknown; recoverable: boolean }
+  /** The transport identified by `attemptId` was lost (unexpected ws close, ping timeout). */
+  | { type: 'transportFailed'; attemptId: number; reason: string }
+  | { type: 'close'; reason: string }
```
1egoman, line 52: Ah, per my previous comment - it looks like `close` does have a `reason`! I think making it an enum (along with other `reason` fields here as well) though would be even better than an arbitrary string.

1egoman (approved):

src/api/SignalClient.ts
lukasIO: do you mean on the ws/transport level?

src/api/SignalClientStateMachine.test.ts
lukasIO: yeah, it isn't pretty. changed it to pass the initial state to the create function with `new` being the default

src/api/SignalClientStateMachine.test.ts
lukasIO: I think the `randomWalks` test is where this is most useful. 
I'm not super concerned with keeping it up to date, as the tests would fail if we change something without updating the documented transitions here

src/api/SignalClientStateMachine.test.ts
lukasIO: yeah, that's what the test in `SignalClient.test.ts` and `SignalClient.test.ts` are supposed to do. 

> So ie, maybe add some assertions around the socket state here, etc.

The socket isn't part of the machine, so I'm not sure what exactly you're suggesting.

src/api/SignalClient.ts
1egoman: No, I mean at the state machine level. At the past job I've referred to a few times now, we had found it really useful to capture as much context on each event as possible about the current state as it was fairly useful for debugging. But also this was just a passing thought and if there isn't a concrete reason to add extra metadata to the events today feel free to disregard.

src/api/SignalClientStateMachine.test.ts
lukasIO: do you have an alternative suggestion or do you generally think this sort of test isn't useful?

src/api/SignalClientStateMachine.test.ts
1egoman: > The socket isn't part of the machine, so I'm not sure what exactly you're suggesting.

Here is my thought rephrased - checking to make sure the state machine goes through each state transition successfully is somewhat valuable but I suspect less so in practice.

What I think is more likely to break is all the code that runs as _part_ of a state transition - this likely isn't part of the state machine directly. What a test that exercises this would look like:
1. Create a new room, validate the socket is closed
2. Call `room.connect()`
3. Make sure the state machine is driven through `connect` -> `connecting` -> `connected`.
4. Validate the socket is open and in a valid state

So you are testing the state machine, but not as directly as you are here. And exactly how the state machine works could change (ie, you could add an intermediate state in between `connecting` and `connected`) and the test wouldn't immediately break (or if it would break, it would be very easy to fix), since it's checking user visible behavior is working as expected.

src/api/SignalClientStateMachine.test.ts
1egoman: Let's chat about it in our next 1:1, I want to hear your thoughts in higher fidelity. Right now I am skeptical but I think I could be convinced.

src/api/SignalClientStateMachine.ts
lukasIO: yeah, enum would be even better. Will leave it as a string for now as there's some sections where we extract the reason from thrown errors via their message. So "reason" here is more like "friendly error message".

PR #2050, opened 2026-08-13
Data track schema definition storage

ladvoc:
Adds support for defining/retrieving data track schema definitions through new methods on local participant. Under the hood, this is powered by [data blobs](https://github.com/livekit/livekit/pull/4619).

Mirrors the implementation in: https://github.com/livekit/rust-sdks/pull/1159.

Usage (publisher):
```ts
const schemaId: DataTrackSchemaId = { name: 'sensor-reading', encoding: 'jsonSchema' };

await room.localParticipant.defineSchema(
  schemaId,
  JSON.stringify({
    type: 'object',
    properties: { temperature: { type: 'number' } },
  }),
);
```

Usage (subscriber):
```ts
const schemaId: DataTrackSchemaId = { name: 'sensor-reading', encoding: 'jsonSchema' };

const definition = await room.localParticipant.getSchema(schemaId, 'publisher-identity');
console.log(JSON.parse(definition));
```

### DIFF
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -6,9 +6,13 @@ import {
   ClientInfo_Capability,
   ConnectionQualityUpdate,
   ConnectionSettings,
+  DataBlob,
+  DataBlobKey,
   DataTrackSubscriberHandles,
   DisconnectReason,
   Encryption_Type,
+  GetDataBlobRequest,
+  GetDataBlobResponse,
   JoinRequest,
   JoinResponse,
   LeaveRequest,
@@ -30,6 +34,8 @@ import {
   SignalTarget,
   SimulateScenario,
   SpeakerInfo,
+  StoreDataBlobRequest,
+  StoreDataBlobResponse,
   StreamStateUpdate,
   SubscribedQualityUpdate,
   SubscriptionPermission,
@@ -202,6 +208,10 @@ export class SignalClient {
   onUnPublishDataTrackResponse?: (event: UnpublishDataTrackResponse) => void;
 
   onDataTrackSubscriberHandles?: (event: DataTrackSubscriberHandles) => void;
+
+  onStoreDataBlobResponse?: (res: StoreDataBlobResponse) => void;
+
+  onGetDataBlobResponse?: (res: GetDataBlobResponse) => void;
 
   onJoined?: (event: JoinResponse) => void;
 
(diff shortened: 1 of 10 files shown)


src/api/SignalClient.ts
```diff
@@ -940,6 +968,14 @@ export class SignalClient {
       if (this.onDataTrackSubscriberHandles) {
         this.onDataTrackSubscriberHandles(msg.value);
       }
+    } else if (msg.case === 'storeDataBlobResponse') {
```
lukasIO, line 971: probably this is too late now, but curious to learn why the success path isn't using the existing `RequestResponse` format, too

src/room/participant/LocalParticipant.ts
```diff
@@ -2187,4 +2254,133 @@ export default class LocalParticipant extends Participant {
...
+   * in an error.
+   *
+   * @param id Identifies the schema; the same ID is provided when publishing a data track
+   *   that uses it.
+   * @param definition The schema definition, stored as-is. It is neither parsed nor
+   *   validated against its {@link DataTrackSchemaId.encoding | encoding}, so the caller
+   *   is responsible for ensuring it is well-formed.
+   * @param signal Optional abort signal to cancel the request.
+   */
+  async defineSchema(
```
lukasIO, line 2275: I know this doesn't really exist as a pattern in this SDK right now, but I'm wondering if we should scope these new methods to a dedicated `localParticipant.dataTracks.*` namespace so that it's clear what they relate to.

Consistency across SDKs is the biggest factor though, so it ultimately only makes sense if we want to commit to that pattern also for other SDKs. 

I find `localP.defineSchema` sounds a bit too generic to be a data-track only feature.

src/api/SignalClient.ts
ladvoc: Great question, this is because typically a `RequestResponse` contains a copy of the request. However, in this case, since the request's payload is potentially large, it didn't make sense to echo the whole thing back if there is an error.

src/room/participant/LocalParticipant.ts
ladvoc: Yes, this is currently what the API looks like in Rust and C++. However, I do like scoping this somehow. Another idea I had is having it under `localParticpant.schemaRegistry`. I think adoption of this feature is early enough that we could justify marking the current APIs as deprecated and scoping them. For additional context, data blobs support, which is the prerequisite to using this feature, isn't enabled for all projects in cloud yet.

lukasIO: Ah ok, if the feature isn't available generally (what's the status for OSS?), we should mark the methods as `experimental` for now to avoid confusion

src/room/participant/LocalParticipant.ts
```diff
@@ -179,6 +190,16 @@ export default class LocalParticipant extends Participant {
...
+  private pendingStoreDataBlobRequests: Map<
+    number,
+    Future<StoreDataBlobResponse, DataTrackSchemaStorageError>
+  >;
+
+  private pendingGetDataBlobRequests: Map<
+    number,
+    Future<GetDataBlobResponse, DataTrackSchemaStorageError>
+  >;
+
```
1egoman, line 202: @ladvoc We talked about this briefly in a 1:1, so I wanted to follow up here. I think this `pendingStoreDataBlobRequests` / `pendingGetDataBlobRequests` pattern you are using here is fairly decent. You are using that `Future` type we discussed, which is good. It's also fairly similar to how [pending rpc responses are tracked](https://github.com/livekit/client-sdk-js/blob/33aadf80f035885b06efac2f43269bddb1c76ae8/src/room/rpc/client/RpcClientManager.ts#L38-L44) in `RpcClientManager`. 

The only smaller, somewhat nitpicky thing that comes to mind is using a numerical id that is signalling defined here as the key does tightly couple this maybe more than would be ideal. So maybe consider making this some uuid or something if that doesn't make things significantly ergonomically worse. Or feel free to ignore this.

PR #2055, opened 2026-08-18
Forward DisconnectReason to ParticipantDisconnected event

seonghunYang:
## Summary

Forwards the server-provided `DisconnectReason` to the `ParticipantDisconnected` event.

## Motivation

The server already distinguishes an intentional leave (client sends `LeaveRequest` → `CLIENT_INITIATED`) from a network failure (no leave → `SIGNAL_CLOSE` / `CONNECTION_TIMEOUT` / `MEDIA_FAILURE`), and ships the result in `ParticipantInfo.disconnectReason`. The JS client receives it but drops it: `handleParticipantDisconnected` only takes the identity and the participant.

There is no way to read it afterwards either — `RemoteParticipant` does not expose it and `participantInfo` is `protected`.

Other SDKs already expose this (`@livekit/rtc-node` via a `disconnectReason` getter, `server-sdk-go` in livekit/server-sdk-go#805, which treated it as a bug fix), and this SDK applied the same pattern to the local `Disconnected` event in #304 — this PR does the same for remote participants.

## Changes

- `handleParticipantDisconnected` takes an optional `disconnectReason` and passes it to the event
- Only the server-update path supplies it; the two local-cleanup call sites (reconnect unwind, room move) pass nothing since they are not real departures
- `UNKNOWN_REASON` is normalized to `undefined`, matching `rtc-node`
- Adding an optional parameter keeps existing listeners working

## Testing

`lint`, `format:check`, `type:check`, `throws:check`, `build` and `publint` pass. Full unit suite passes (715 tests / 47 files). Verified the built `dist/src/room/Room.d.ts` declares `participantDisconnected: (participant: RemoteParticipant, disconnectReason?: DisconnectReason) => void`.

### DIFF
--- a/.changeset/forward-participant-disconnect-reason.md
+++ b/.changeset/forward-participant-disconnect-reason.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Forward `DisconnectReason` to the `ParticipantDisconnected` event
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -1913,7 +1913,13 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
 
       // when it's disconnected, send updates
       if (info.state === ParticipantInfo_State.DISCONNECTED) {
-        this.handleParticipantDisconnected(info.identity, remoteParticipant);
+        this.handleParticipantDisconnected(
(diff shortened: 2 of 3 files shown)


lukasIO (approved):
thanks for the PR, this makes sense

PR #2057, opened 2026-08-19
fix: treat TokenSource JWTs without exp as expired

SashaMIT:
## Problem

`isResponseTokenValid` used `if (!nbf || !exp) return true`. Missing either claim skipped the rest of the check, so `TokenSourceCached` kept returning the cached response.

`nbf` is optional in RFC 7519. A JWT with `exp` in the past and no `nbf` was treated as still valid. A JWT with no `exp` was treated as never expired.

This is the client cache counterpart of livekit/protocol#1706, livekit/python-sdks#779, livekit/node-sdks#710, livekit/server-sdk-ruby#97, and livekit/server-sdk-kotlin#170. Those require `exp` on verify. This helper decides whether to refetch.

## Fix

Require `exp`. Check `nbf` only when it is present. Still apply the existing one-minute `exp` skew.

Threat-model: the cached participant token is already in the client. The attacker does not need host or TokenSource authority. An expired token that omits `nbf` (common) never triggers a refetch. A token that omits `exp` is cached forever. The server may still reject the JWT; the client never goes back to the token source.

## Test plan

- `pnpm exec vitest run src/room/token-source/utils.test.ts` (13/13)
- `pnpm test` (718/718)
- lint, format, typecheck, throws-check, build, compat, publint green locally
- Revert-tested: restoring `if (!nbf || !exp) return true` fails `should treat a jwt without exp as expired` and `should honor exp when nbf is absent`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/token-source-require-exp.md
+++ b/.changeset/token-source-require-exp.md
@@ -1,0 +1,5 @@
+---
+'livekit-client': patch
+---
+
+Treat TokenSource JWTs without `exp` as expired, and still honor `exp` when `nbf` is absent
--- a/src/room/token-source/test-tokens.ts
+++ b/src/room/token-source/test-tokens.ts
@@ -1,3 +1,11 @@
+// Builds an unsigned (`alg: none`) JWT. These aren't signed at all, so they can only be used in
+// tests which don't care about the signature.
+function unsignedToken(payload: Record<string, unknown>) {
+  const encode = (value: Record<string, unknown>) =>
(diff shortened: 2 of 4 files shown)


1egoman (approved):
Looks good to me, thanks for the contribution! Just pushed a small update to move your `unsignedToken` function into `test-tokens.ts` so it works a little closer to how the existing tests work.

SashaMIT:
Thanks for the review and the merge.

PR #2073, opened 2026-08-27
fix: error on uncompressed datastreams when chunks are missing

### LINKED ISSUE #2072: Uncompressed data streams silently accept missing and out-of-order chunks
### What we saw
A client in production received an unparseable JSON payload over an uncompressed `ByteStream`. We have not been able to conclusively confirm the cause, and we're filing this as a suspected one: a missing or reordered chunk in the `ByteStream`.

### The bug
`ByteStreamReader` and `TextStreamReader` never look at `DataStream_Chunk.chunkIndex`. Chunks are concatenated in arrival order, so a missing chunk produces a payload with a hole in the middle and a reordered pair produces bytes in the wrong order. In both cases `readAll()` resolves successfully. Application code receives silently corrupted data from a call that reported success, with no error, no event, and nothing on the reader it could check to find out.

The compressed path already enforces this invariant. `ensureOrderedChunks` errors with `DataStreamErrorReason.Incomplete` on a gap and drops duplicates. Uncompressed streams carry the same chunks over the same channel and get no equivalent check: `handleStreamChunk` enqueues unconditionally, so nothing on the uncompressed path ever reads the index.

### Reachability
We went looking for how a chunk could go missing in the first place, and could not close the loop from the client side. Data stream packets go over `DataChannelKind.RELIABLE`, and both reconnect paths appear covered: a full reconnect errors in-flight streams, and a resume relies on the SFU replaying missed reliable packets from the `datachannelReceiveStates` cursors the client sends in sync state. The resume replay is the one link we can't check: whether it is ever short is a server-side question we have no visibility into.

Nothing would catch it if that happened. `handleDataMessage` drops packets with `sequence <= lastSeq` to absorb replayed duplicates, but never flags `sequence > lastSeq + 1`, so a short replay is ignored at the engine layer, which is the layer that knows a replay occurred.

### Reproduction

Deterministic at the manager level; `src/room/data-stream/incoming/IncomingDataStreamManager.test.ts` already has the harness. Register a stream handler, feed a `streamHeader`, feed chunks where chunk *i* carries byte/character `i`, feed a `streamTrailer`, then `await reader.readAll()`.

| | `totalLength` | chunks delivered | expected | actual |
|---|---|---|---|---|
| **A** | none | 0, 1, 3, 4 | reject | resolves `[0,1,3,4]` |
| **B** | none | 0, 1, 3, 2, 4 | reject | resolves `[0,1,3,2,4]` |
| **C** | 5 | 0, 1, 3, 4 | reject | rejects (`Incomplete`) ✓ |
| **D** | 5 | 0, 1, 3, 2, 4 | reject | resolves `[0,1,3,2,4]` |

Case C is the only one caught today, and it is caught by byte count rather than by index. Text streams behave identically to A and B (`"0134"` and `"01324"`).

### Logs

```shell

```

### System Info

```shell
Browsers:
  Electron 43.0.0 (Chromium), macOS
npmPackages:
  livekit-client: 2.18.6   (version of the production sighting)
Server: LiveKit Cloud, US East 1, UDP
```

### Severity

### Additional Information

Use `ensureOrderedChunks` on the uncompressed byte and text paths.

lukasIO:
closes https://github.com/livekit/client-sdk-js/issues/2072

### DIFF
--- a/src/room/data-stream/incoming/IncomingDataStreamManager.test.ts
+++ b/src/room/data-stream/incoming/IncomingDataStreamManager.test.ts
@@ -1371,6 +1371,37 @@ describe('IncomingDataStreamManager', () => {
       expect(concatChunks(await reader.readAll())).toStrictEqual(bytes);
     });
 
+    it('should drop a higher-version chunk resent at an already-received index', async () => {
+      const manager = new IncomingDataStreamManager();
+      manager.setConnected(true);
+
+      const readerPromise = new Promise<TextStreamReader>((resolve) => {
+        manager.registerTextStreamHandler('my-topic', (reader) => resolve(reader));
+      });
+
+      const streamId = crypto.randomUUID();
+      const text = 'hello world';
+      const textBytes = new TextEncoder().encode(text);
+
+      manager.handleDataStreamPacket(
+        headerPacket(streamId, 'textHeader', { totalLength: BigInt(textBytes.length) }),
+        Encryption_Type.NONE,
+      );
+      manager.handleDataStreamPacket(chunkPacket(streamId, 0, textBytes, 0), Encryption_Type.NONE);
+      // Chunk-level `version` retcon is not supported: a reader that has already yielded chunk 0 to
+      // its consumer cannot retract it, so a resend at the same index is dropped like any other
+      // duplicate rather than superseding the original. See the note on
+      // `TextStreamReader.handleChunkReceived`.
+      manager.handleDataStreamPacket(
+        chunkPacket(streamId, 0, new TextEncoder().encode('goodbye world'), 1),
+        Encryption_Type.NONE,
+      );
+      manager.handleDataStreamPacket(trailerPacket(streamId), Encryption_Type.NONE);
+
+      const reader = await readerPromise;
+      expect(await reader.readAll()).toStrictEqual(text);
+    });
+
     it('should reframe multibyte UTF-8 on chunk boundaries when decompressing a text stream', async () => {
       const manager = new IncomingDataStreamManager();
       manager.setConnected(true);
--- a/src/room/data-stream/incoming/IncomingDataStreamManager.test.ts
+++ b/src/room/data-stream/incoming/IncomingDataStreamManager.test.ts
@@ -1237,6 +1237,140 @@ describe('IncomingDataStreamManager', () => {
       await expect(reader.readAll()).rejects.toThrow('Missing chunk(s)');
     });
 
+    it('should error on a gap in chunk indices on an uncompressed text stream', async () => {
+      const manager = new IncomingDataStreamManager();
+      manager.setConnected(true);
+
+      const readerPromise = new Promise<TextStreamReader>((resolve) => {
+        manager.registerTextStreamHandler('my-topic', (reader) => resolve(reader));
+      });
+
+      const streamId = crypto.randomUUID();
+      const text = randomText(3_000);
+      const textBytes = new TextEncoder().encode(text);
+      const split = Math.floor(textBytes.length / 3);
+
+      manager.handleDataStreamPacket(
+        headerPacket(streamId, 'textHeader', {
+          totalLength: BigInt(textBytes.length),
+          compression: DataStream_CompressionType.NONE,
+        }),
+        Encryption_Type.NONE,
+      );
+      manager.handleDataStreamPacket(
+        chunkPacket(streamId, 0, textBytes.slice(0, split)),
+        Encryption_Type.NONE,
+      );
+      // Skip chunk index 1 entirely — a gap means the payload cannot be reassembled in order.
+      manager.handleDataStreamPacket(
+        chunkPacket(streamId, 2, textBytes.slice(split)),
+        Encryption_Type.NONE,
+      );
+
+      const reader = await readerPromise;
+      await expect(reader.readAll()).rejects.toThrow('Missing chunk(s)');
+    });
+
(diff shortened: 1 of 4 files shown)


src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -241,7 +241,7 @@ export default class IncomingDataStreamManager {
             info,
             compressed
               ? inflateRawByteChunkStream(stream, streamHeader.streamId, this.maxPayloadByteLength)
-              : stream,
+              : stream.pipeThrough(ensureOrderedChunks(streamHeader.streamId)),
             // `totalLength` is the pre-compression size, and the reader counts decompressed bytes,
```
1egoman, line 245: suggestion: Could you add a test for this? I think something like [this one](https://github.com/livekit/client-sdk-js/blob/9769282ccf1d4473f4be46c047737a1d96173316/src/room/data-stream/incoming/IncomingDataStreamManager.test.ts#L1204-L1238) only set `compression: DataStream_CompressionType.NONE` for the packets (instead of `DEFLATE_RAW`) would be sufficient.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -241,7 +241,7 @@ export default class IncomingDataStreamManager {
             info,
             compressed
               ? inflateRawByteChunkStream(stream, streamHeader.streamId, this.maxPayloadByteLength)
-              : stream,
+              : stream.pipeThrough(ensureOrderedChunks(streamHeader.streamId)),
```
1egoman, line 244: suggestion: I think the same issue also exists with text streams here: https://github.com/livekit/client-sdk-js/blob/9769282ccf1d4473f4be46c047737a1d96173316/src/room/data-stream/incoming/IncomingDataStreamManager.ts#L342-L351

Similarly also a test which exercises this per my other comment would also be a good idea.

1egoman (commented):
Generally looks good, I think before merging it would be good to get this also fixed for text streams and add test(s) to make sure that this doesn't regress.

@lukasIO If you'd like me to take this over let me know. I will also see if this is an issue in the rust implementation, and if so, open a PR fixing it over there.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -344,7 +344,7 @@ export default class IncomingDataStreamManager {
             info,
             compressed
               ? inflateRawChunkStream(stream, streamHeader.streamId, this.maxPayloadByteLength)
-              : stream,
+              : stream.pipeThrough(ensureOrderedChunks(streamHeader.streamId)),
```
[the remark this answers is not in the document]
lukasIO: the version based update has never been used. 
retcon was explicitly removed in https://github.com/livekit/client-sdk-js/pull/1410

src/room/data-stream/incoming/IncomingDataStreamManager.test.ts
```diff
@@ -1237,6 +1237,171 @@ describe('IncomingDataStreamManager', () => {
...
+      manager.handleDataStreamPacket(chunkPacket(streamId, 0, textBytes, 0), Encryption_Type.NONE);
+      // Chunk-level `version` retcon is not supported: a reader that has already yielded chunk 0 to
+      // its consumer cannot retract it, so a resend at the same index is dropped like any other
+      // duplicate rather than superseding the original. See the note on
+      // `TextStreamReader.handleChunkReceived`.
+      manager.handleDataStreamPacket(
+        chunkPacket(streamId, 0, new TextEncoder().encode('goodbye world'), 1),
+        Encryption_Type.NONE,
+      );
+      manager.handleDataStreamPacket(trailerPacket(streamId), Encryption_Type.NONE);
```
1egoman, line 1405: nitpick: It also might be good to add a test here which sends chunk version 1 and then 0. From our conversation on slack, this is now "first one wins", so assuming I have that right I think it would be good to test both permutations.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -484,17 +480,20 @@ function ensureOrderedChunks(
...
-          `Missing chunk(s) ${lastChunkIndex + 1}..${index - 1} for compressed data stream ${streamId} - cannot continue decompressing`,
+          `Missing chunk(s) ${lastChunkIndex + 1}..${index - 1} for data stream ${streamId} - cannot reassemble payload`,
           DataStreamErrorReason.Incomplete,
         );
       }
       lastChunkIndex = index;
+      if (value.content.length === 0) {
+        return;
+      }
       controller.enqueue(value);
```
1egoman, line 497: (Just wanted to say explicitly that I think moving this to the send path instead of the receive path is probably the better place for this to live!)

1egoman (approved):
Looks good to me!

src/room/data-stream/incoming/IncomingDataStreamManager.ts
lukasIO: we can also add a guard on the send path, but the receive path has additional value in so far, that we'd account for older clients interop

src/room/data-stream/incoming/IncomingDataStreamManager.test.ts
lukasIO: good idea, added this within the same test

PR #2077, opened 2026-09-01
handle late reconnect responses

lukasIO:
ensures that during a reconnect a ReconnectResponse arrives after some other signalling message, the ReconnectResponse still gets applied retroactively.


1egoman (approved):

lukasIO:
superseded by #2082

PR #2079, opened 2026-09-02
finalization strategy for worker listeners

lukasIO:


src/e2ee/E2eeManager.ts
```diff
@@ -87,6 +88,20 @@ export class E2EEManager
...
+   * Runs a cleanup callback once this manager is garbage collected. Lets the
+   * log-level listener (held in a module-global Set on the main-thread logger)
+   * fall out of scope even when the consumer forgets to call `dispose()`.
+   */
+  private static disposeRegistry =
+    typeof FinalizationRegistry !== 'undefined' &&
+    typeof WeakRef !== 'undefined' &&
+    new FinalizationRegistry((cleanup: () => void) => {
+      cleanup();
+    });
```
1egoman, line 104: question: Have you been able to actually verify this runs? I think there's a reference cycle between E2eeManager and Room currently (`E2eeManager.room` and `Room.e2eeManager`) so I'm not convinced without breaking this cycle that `E2eeManager` would ever get garbage collected.

src/e2ee/E2eeManager.ts
```diff
@@ -129,13 +144,62 @@ export class E2EEManager
...
+  dispose() {
+    this.unsubscribeLogLevel?.();
+    this.unsubscribeLogLevel = undefined;
+    if (E2EEManager.disposeRegistry) {
+      E2EEManager.disposeRegistry.unregister(this);
+    }
+    if (this.worker) {
+      this.worker.onmessage = null;
+      this.worker.onerror = null;
+    }
```
1egoman, line 218: question: Does there need to be something here which waits for pending buffers being e2ee encrypted before disposal can be completed? Otherwise given this is a public method that a user could call (not marked `@internal` / etc) it could be possible to break a room by calling it preemptively.

src/e2ee/E2eeManager.ts
lukasIO: you're right, that this won't work on its own, or rather won't address the issues around the cyclical Room references. 
My intention was to set up everything in a way that would allow for it to get dropped once the cyclical room references are taken care of. 
Addressing this is out of scope for this PR however.

lukasIO: I've added tests to verify the cleanup works without room references as expected, the value of that is somewhat questionable within the greater picture, but it allows us to verify the cleanup logic within the module works as expected.

src/e2ee/E2eeManager.ts
lukasIO: good point. I marked it as `internal` and added rejections for pending data encryption futures

src/e2ee/E2eeManager.ts
```diff
@@ -184,15 +184,30 @@ export class E2EEManager
...
     this.unsubscribeLogLevel = undefined;
     if (E2EEManager.disposeRegistry) {
       E2EEManager.disposeRegistry.unregister(this);
     }
+
+    // Reject pending futures BEFORE detaching worker handlers, so any late
+    // response can't resolve one after we've cut the pipe. Each future's
+    // `onFinally` deletes its own map entry, so both maps drain themselves.
+    // Snapshot before iterating in case a rejection handler mutates the map.
+    const disposalError = new Error('E2EEManager disposed');
```
1egoman: nitpick: Maybe this should be a custom error type?

1egoman (approved):

src/e2ee/E2eeManager.ts
1egoman: Sounds good. I suppose this does set us up well to have this all working once we can break that reference cycle. It doesn't have to be part of this PR, but I wonder if it could be prudent to introduce that `room.dispose()` type method we talked about in a 1:1 sooner rather than later and that could be used as a way to trigger this in the more near term.

PR #2083, opened 2026-09-03
Support simulcast for svc codecs (vp9/av1)

cnderrauber:
Conditions to enable:
1. The server supports it (version > 1.13.6).
2. `ScalabilityMode: L1tx` and `simulcast: True`
3. Chrome only, version >= M113

### DIFF
--- a/src/room/participant/LocalParticipant.ts
+++ b/src/room/participant/LocalParticipant.ts
@@ -1150,9 +1150,6 @@ export default class LocalParticipant extends Participant {
           opts.simulcast = false;
         }
 
-        // when the caller asked for VP9/AV1 as rid based simulcast (simulcast + an L1Tx
-        // scalability mode) the SVC defaults below must not apply: they would replace the
-        // requested mode and force a contentHint that only makes sense for SVC screenshare.
         const svcSimulcast = isSVCSimulcast(videoCodec, opts);
         if (isSVCCodec(videoCodec) && !svcSimulcast) {
           if (track.source === Track.Source.ScreenShare) {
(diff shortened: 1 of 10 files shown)


src/room/participant/LocalParticipant.ts
```diff
@@ -240,6 +245,11 @@ export default class LocalParticipant extends Participant {
     }
   }
 
+  private getServerVersion(): string | undefined {
+    const joinResponse = this.engine?.latestJoinResponse;
+    return joinResponse?.serverInfo?.version || joinResponse?.serverVersion || undefined;
+  }
+
```
1egoman: nitpick: with the introduction of this, could you also update https://github.com/livekit/client-sdk-js/blob/ae32e316171e481bfef09cb46636f2c3186aeca3/src/room/Room.ts#L342 to use this method? I think that's the only other place in the sdk where there's an explicit SFU version check currently that I know about.

src/room/utils.ts
```diff
@@ -185,6 +185,65 @@ export function negotiateDependencyDescriptor(transceiver: RTCRtpTransceiver): b
...
+/**
+ * VP9 and AV1 are published as SVC (a single RTP stream carrying every spatial layer)
+ * by default. They can instead be published as real, rid based simulcast — one
+ * independent stream per rid, each carrying a single spatial layer — when the caller
+ * opts in with `simulcast: true` and a single spatial layer scalability mode (`L1Tx`).
+ *
+ * The SFU has to be told about this: without an explicit
+ * `SimulcastCodec.videoLayerMode` it assumes `MULTIPLE_SPATIAL_LAYERS_PER_STREAM` for
+ * any SVC capable codec.
+ */
```
1egoman, line 197: suggestion: It might be worth adding some of this context to a user facing docstring in `RoomConnectOptions` (and eventually docs.livekit.io). As is I don't think it is very discoverable.

Also maybe worth considering as well - some sort of debug log that prints when `isSVCSimulcast(...)` is true in the connection path, to be able to confirm it is enabled properly.

1egoman (approved):
High level makes sense to me, but I'm probably not the best person to give a lower level review and defer to @xianshijing-lk or @lukasIO

xianshijing-lk (approved):

src/room/utils.ts
cnderrauber: Have added a comment to TrackPublishDefaults.The doc will be generated from it.

Issue #1915, opened 2026-04-24
Coalesce burst negotiations and avoid full-reconnect on single NegotiationError under concurrent signaling


U-OK:
### Describe the problem

In rooms with many participants, a burst of server-initiated renegotiations (e.g. several participants leaving within a few seconds — each leave triggers `onMediaSectionsRequirement`) causes severe client-side degradation:

1. `RTCEngine.negotiate()` is invoked repeatedly in rapid succession. Each invocation registers `Closing` and `Restarting` listeners on the engine emitter (`this.on(EngineEvent.Closing, handleClosed)` / `this.on(EngineEvent.Restarting, handleClosed)`). With ~11 concurrent calls, Node's EventEmitter raises `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 restarting/closing listeners added`.

2. When one of the concurrent negotiations fails, `fullReconnectOnNext = true` is set unconditionally, triggering a full engine restart (`cleanupPeerConnections` + `cleanupClient` + rejoin signal + rejoin engine).

3. Any track publish in flight at that moment (`publishOrRepublishTrack`) awaits `reconnectFuture.promise` and stalls for the full duration of the restart.

Observed impact in production (12-participant room, 7 leaves within 4s, Chrome 146 / macOS, `livekit-client@2.18.6`, stable network, signal WS never disconnected):

- `setMicrophoneEnabled(true)` stalled for **5.4 seconds** when the user first enabled their mic during the mass-leave moment.
- Noise-suppression processor initialization (`AudioWorklet.addModule`) failed immediately after the forced restart with `AbortError: Unable to load a worklet's module`, forcing passthrough fallback.
- Two `MaxListenersExceededWarning` entries appeared in Sentry (one for `restarting`, one for `closing`).

Relevant breadcrumbs:

```
07:23:17.395Z publishing track {kind:“audio”, source:“microphone”}
07:23:17-19 7 × video_room.member.status_updated user_state:“left”
07:23:19.730Z MaxListenersExceededWarning: 11 restarting listeners added
(stack: RTCEngine.negotiate → client.onMediaSectionsRequirement → SignalClient.handleSignalResponse)
07:23:19.939Z MaxListenersExceededWarning: 11 closing listeners added (same stack)
07:23:21.421Z [mic] slow toggle — setMicMs=5399 totalMs=5433 trackExists=false
```

[...]

PR #2084, opened 2026-09-03
Fix RTCEngine.negotiate EngineEvent.Closing / EngineEvent.Restarting listeners to not subscribe to a new listener every call

1egoman:
Previously, every `negotiate` call would add a new listener for `EngineEvent.Closing` and `EngineEvent.Restarting`. If this path was hit at too frequent of a rate, many listeners would get added and a `MaxListenersExceededWarning ` would get logged.

Instead, aim for an approach similar to https://github.com/livekit/client-sdk-js/pull/1896 - attach the listener in one place, and fan out this one listener entry to all `negotiate` calls dynamically.

Fixes the first part (point 1) of https://github.com/livekit/client-sdk-js/issues/1915. The other two points need some further discussion / thought but this one makes a lot of sense to me and was fairly unambiguous.

### DIFF
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -260,6 +260,14 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
   /** timestamp (ms) the primary transport entered `CONNECTING`, used to bound how long we tolerate it */
   private transportConnectingSince?: number;
 
+  /**
+   * Abort handlers for the in-flight `negotiate()` calls, which a single central
+   * Closing/Restarting listener pair fans out to. Registering a listener pair per call instead
+   * accumulates them on the engine emitter and trips its max-listener warning under
+   * renegotiation bursts.
+   */
+  private pendingNegotiationAborts = new Set<() => void>();
+
   constructor(private options: InternalRoomOptions) {
     super();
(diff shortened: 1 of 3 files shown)


src/room/RTCEngine.ts
```diff
@@ -300,6 +308,15 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
     this.client.onJoined = (joinResponse) => this.emit(EngineEvent.Joined, joinResponse);
+
+    const abortPendingNegotiations = () => {
+      // Iterate a copy: each handler removes itself from the set as its negotiation settles.
+      for (const abort of Array.from(this.pendingNegotiationAborts)) {
+        abort();
+      }
+    };
+    this.on(EngineEvent.Closing, abortPendingNegotiations);
+    this.on(EngineEvent.Restarting, abortPendingNegotiations);
```
1egoman, line 319: Note to reviewers - an important deviation from the [example pull request linked in the description](https://github.com/livekit/client-sdk-js/pull/1896,) is I opted not to use a `Future` because the pending list needs to be resettable. There also isn't a super elegant way to remove a listener in the `finally` block in the same way with this solution - I am calling `this.pendingNegotiationAborts.delete` to do this.

lukasIO (approved):

PR #2085, opened 2026-09-03
feat: add disableIceLite connect option

subham2006:
Adds a `disableIceLite` option to `RoomConnectOptions` that sets `ConnectionSettings.disable_ice_lite` on the join request (and `disable_ice_lite=1` on the v0 path), so a client can ask a server running `use_ice_lite: true` for a full ICE agent, which the server already honors per participant since livekit/livekit#2862 and which livekit/protocol#1151 carried into the join request proto. livekit-server reads `disable_ice_lite` from the signal URL query string and this SDK preserves caller query params on the `/rtc` URL, so `wss://host?disable_ice_lite=1` already works as an undocumented workaround; no SDK exposes it as an option. libwebrtc clients behind some NATs nominate noticeably slower against a lite remote, so a browser or React Native app had no way to opt out.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -166,6 +166,12 @@ export interface InternalRoomConnectOptions {
    */
   rtcConfig?: RTCConfiguration;
 
+  /**
+   * when the server runs ICE lite, asks it to use a full ICE agent for this participant.
+   * this is a server-side setting that the server may ignore, defaults to false
+   */
+  disableIceLite?: boolean;
+
   /** specifies how often an initial join connection is allowed to retry (only applicable if server is not reachable) */
   maxRetries: number;
 
--- a/src/options.ts
+++ b/src/options.ts
@@ -170,7 +170,7 @@ export interface InternalRoomConnectOptions {
    * when the server runs ICE lite, asks it to use a full ICE agent for this participant.
    * this is a server-side setting that the server may ignore, defaults to false
    */
-  disableIceLite?: boolean;
+  disableIceLite: boolean;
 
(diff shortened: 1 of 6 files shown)


1egoman:
@subham2006 Hi! This pull request generally makes sense though it is worth mentioning that this flag doesn't seem to exist on any other sdks (I also don't know of any sdks that allow passing raw query parameters?).

I'm going to reach out to the SFU team and make sure they don't have any concerns with adding this but unless they bring anything up I'll probably aim to get this merged early to mid next week!

Also, prior to merging, would you be able to click the "CLA not signed yet" button and sign the CLA? Thanks!

subham2006:
Correct, no SDK exposes this, fixed the description. CLA signed, thanks!

src/options.ts
```diff
@@ -166,6 +166,12 @@ export interface InternalRoomConnectOptions {
    */
   rtcConfig?: RTCConfiguration;
 
+  /**
+   * when the server runs ICE lite, asks it to use a full ICE agent for this participant.
+   * this is a server-side setting that the server may ignore, defaults to false
+   */
+  disableIceLite?: boolean;
```
lukasIO: (with rtcConfig above being the exception) the way default values are handled within the SDK is to make the field in `InternalRoomConnectOptions` non optional and declare a default in `defaults.ts`.

1egoman: Fixed this, thanks @lukasIO!

1egoman:
Talked with the SFU folks and this is fine on their end. This should be ready to merge once the ci build passes.

1egoman (approved):

PR #2086, opened 2026-09-05
fix: add code 1000 to onclose handler in WebSocketStream

phaitonican:
React Native's WebSocket never sets wasClean (react-native:267), so livekit-client treats every close as unexpected — even the normal code-1000 close that happens on disconnect. On web wasClean is set, so it's silent; on mobile it logs WS closed unexpectedly with code 1000 + error reading from signal stream.

### DIFF
--- a/.changeset/short-dragons-look.md
+++ b/.changeset/short-dragons-look.md
@@ -1,0 +1,5 @@
+---
+"livekit-client": patch
+---
+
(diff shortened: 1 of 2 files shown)


lukasIO:
This makes sense, thanks for the PR!

lukasIO (approved):

PR #2092, opened 2026-09-09
fix: expose participantIdentity on RoomEvent.DataReceived

mariusgassen:
- `handleDataPacket` had `packet.participantIdentity` but dropped it before passing to `handleUserPacket`
- added as required param internally, optional trailing arg on `RoomEventCallbacks.dataReceived`
- allows callers to buffer and replay messages when the sender hasn't joined yet (data channel / participant join event race)

### DIFF
--- a/src/room/Room.ts
+++ b/src/room/Room.ts
@@ -2098,7 +2098,13 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     // find the participant
     const participant = this.remoteParticipants.get(packet.participantIdentity);
     if (packet.value.case === 'user') {
-      this.handleUserPacket(participant, packet.value.value, packet.kind, encryptionType);
+      this.handleUserPacket(
+        participant,
+        packet.value.value,
+        packet.kind,
+        encryptionType,
+        packet.participantIdentity,
+      );
     } else if (packet.value.case === 'transcription') {
       this.handleTranscription(participant, packet.value.value);
     } else if (packet.value.case === 'sipDtmf') {
@@ -2148,6 +2154,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
     userPacket: UserPacket,
     kind: DataPacket_Kind,
     encryptionType: Encryption_Type,
+    participantIdentity: string,
   ) => {
     this.emit(
       RoomEvent.DataReceived,
@@ -2156,6 +2163,7 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>) 
       kind,
       userPacket.topic,
       encryptionType,
+      participantIdentity,
     );
 
     // also emit on the participant
@@ -3061,6 +3069,7 @@ export type RoomEventCallbacks = {
     kind?: DataPacket_Kind,
     topic?: string,
     encryptionType?: Encryption_Type,
+    participantIdentity?: string,
   ) => void;
   sipDTMFReceived: (dtmf: SipDTMF, participant?: RemoteParticipant) => void;
   transcriptionReceived: (
--- a/src/room/events.ts
+++ b/src/room/events.ts
@@ -225,7 +225,7 @@ export enum RoomEvent {
    * Data packets provides the ability to use LiveKit to send/receive arbitrary payloads.
    * All participants in the room will receive the messages sent to the room.
    *
-   * args: (payload: NonSharedUint8Array, participant: [[Participant]], kind: [[DataPacket_Kind]], topic?: string)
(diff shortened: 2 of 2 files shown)


lukasIO:
thanks for the PR. This race shouldn't occur anymore in practice as the server is buffering messages for some time until a participant becomes active. 
In what scenarios do you see this happening? 
Is there a way to consistently reproduce it?

mariusgassen:
> In what scenarios do you see this happening? Is there a way to consistently reproduce it?

Are you refering to this commit?
https://github.com/livekit/livekit/commit/6a64df20df7a0cef6cbe17d305347387659f180c

I was running livekit-server 1.13.2 when implementing a client side buffering of messages with unknown remote participant senders.
I will update to the latest server version and verify this makes our downstream implementation obsolete.

The original scenario was caused by our application's internal protocol. Participants would only accept messages from certain other participants. Without being able to check the sender (message sent directly after join, received before the sender was known as a remote participant), the messages were just discarded.

**Edit**:
I don't think the above mentioned commit solves our issue.
We use the Go server SDK to publish a data packet on the connected local participant `LocalParticipant.PublishDataPacket`, not on the `RoomService.SendData` API.
The message is delivered fine.

The race we hit is purely on the client side: the data channel message (WebRTC) arrives before the participant update (join) signal (WebSocket) and the `RemoteParticipant` is created. The `RTCEngine` event loop processes them in either order. We observe both events more or less consistently within a ~30ms window: data received (and dropped) - then participant join.

We can reproduce it by having a participant send a data message immediately on connected, before the receiving participant has had a chance to process the join signal.

Our solution is a client side buffering of messages with no known remote participant - requiring this PRs change.

We already observed and implemented something similar on the server component side:
The Go server SDK provides both - the Sender (`*RemoteParticipant`) and SenderIdentity (`string`) in the [`DataReceiveParams`](https://github.com/livekit/server-sdk-go/blob/ea6ed3488fbda18b78e2dfd06f041963c67299b8/data.go#L65) for processing incoming messages.
This way we can buffer messages within a timed window, in which the server-side participant awaits the `ParticipantConnected` event for the sender before discarding them completly.

mariusgassen:
I quickly vibe-coded two clients connecting to a room where one immediately sends a message.

Running against a local LiveKit server 1.13.6 it's reliably reproducible - the data channel message arrives before the participant connected event - up to 2500ms late.

```
async function makeToken(apiKey: string, apiSecret: string, roomName: string, identity: string) {
  const t = new AccessToken(apiKey, apiSecret, { identity });
  t.addGrant({ room: roomName, roomJoin: true, canPublishData: true });
  return t.toJwt();
}

function setupReceiver(room: Room) {
  let dataReceivedAt: number | undefined;
  let raceDetected = false;

  room.on(RoomEvent.DataReceived, (_payload, participant: RemoteParticipant | undefined) => {
    dataReceivedAt = performance.now();
    if (!participant) {
      raceDetected = true;
      log('[receiver] DataReceived — participant: UNDEFINED (race hit!) — sender identity unknown to SDK', 'race');
    } else {
      log(`[receiver] DataReceived — participant: "${participant.identity}" (no race this time)`, 'ok');
    }
  });

  room.on(RoomEvent.ParticipantConnected, (p) => {
    const delta = dataReceivedAt !== undefined
      ? `${(performance.now() - dataReceivedAt).toFixed(1)}ms after DataReceived`
      : 'before DataReceived';
    log(`[receiver] ParticipantConnected — "${p.identity}" arrived ${delta}`);
  });

  room.on(RoomEvent.Connected, () => {
    log('[receiver] Connected event received');
  });

  return { raceDetected: () => raceDetected };
}

function setupSender(room: Room, sendDelayMs: number) {
  room.on(RoomEvent.Connected, () => {
    log(`[sender]   Connected event received — publishing data in ${sendDelayMs}ms`);
    setTimeout(
      () => room.localParticipant.publishData(new TextEncoder().encode('hi'), { reliable: true }),
      sendDelayMs,
    );
  });
}

async function runTest(serverUrl: string, apiKey: string, apiSecret: string, sendDelayMs: number) {
  const roomName = `event-race-${Date.now()}`;
  const receiver = new Room();
  const sender   = new Room();

  const { raceDetected } = setupReceiver(receiver);
  setupSender(sender, sendDelayMs);

  log('[receiver] connecting…');
  await receiver.connect(serverUrl, await makeToken(apiKey, apiSecret, roomName, 'receiver'));
  log('[receiver] connected');

  log('[sender] connecting…');
  await sender.connect(serverUrl, await makeToken(apiKey, apiSecret, roomName, 'sender'));

  await new Promise((r) => setTimeout(r, 5000));

  log('');
  log(
    raceDetected()
      ? '✗ Race detected: RemoteParticipant was undefined at DataReceived time.'
      : '✓ No race this run (timing-dependent — try again).',
    raceDetected() ? 'race' : 'ok',
  );

  await Promise.all([receiver.disconnect(), sender.disconnect()]);
}
```

image

lukasIO (approved):
thanks for the reproduction. 

This makes sense to add and your implementation looks good

PR #2093, opened 2026-09-10
Initial implementation for transcription deduplication

1egoman:
This pull request contains the web implementation for the transcription deduplication project.

A high level summary of the project:
> Today all transcriptions are sent twice over the reliable data channel, once in legacy format, and once in modern data streams format. Because they are sent twice, the reliable data channel is clogged with data and when an end user is on a poor bandwidth network connection (ie, mobile) this can result in a very poor user experience - RPCs get missed, data stream are only partially delivered, etc.

The way this has implemented is there is a new client protocol version that has been introduced, `3`. Any client which advertises this client protocol must:
- NOT require the agents sdk to send it legacy transcriptions. Only modern, data stream transcriptions will need to be sent over the data channel. If any legacy data stream transcriptions are sent, they will be ignored.
- Back-convert any modern data stream transcriptions into legacy `RoomEvent.TranscriptionReceived` events. This is the majority of the change - doing this back conversion ended up being more involved than initially expected.

In parallel to this, a change has been [made to agents](https://github.com/livekit/agents/pull/7240) and (soon) to agents-js which stops sending legacy transcriptions to clients with client protocol of 3 or higher.

## TODO
- [ ] Fix this out of date comment: https://github.com/livekit/client-sdk-js/pull/2093#discussion_r4009399663

### DIFF
--- a/docs/superpowers/plans/2026-09-04-transcription-back-conversion.md
+++ b/docs/superpowers/plans/2026-09-04-transcription-back-conversion.md
@@ -1,0 +1,1885 @@
+# Transcription Back-Conversion Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Rebuild `RoomEvent.TranscriptionReceived` (and its `ParticipantEvent` / `TrackEvent`
+siblings) from `lk.transcription` data streams instead of legacy `Transcription` data packets, and
+advertise client protocol 3 so agents can stop publishing the legacy copy.
+
+**Architecture:** A new pure `TranscriptionStreamConverter` accumulates `lk.transcription` text
+streams into synthesized `Transcription` protobuf messages and feeds them to the existing
+`Room.handleTranscription` emit path. `IncomingDataStreamManager` becomes a typed event emitter and
+taps the reserved topic with a `transcriptionStreamArrived` event, so the SDK can read
+`lk.transcription` without stealing the topic from an application handler. Legacy `Transcription`
+packets are ignored unconditionally — there is no gate.
+
+**Tech Stack:** TypeScript, vitest, `@livekit/protocol` (protobuf-es), Web Streams API.
+
+**Spec:** `docs/superpowers/specs/2026-09-04-transcription-back-conversion-design.md`
+
+## Global Constraints
+
+- Type check with `npx tsc --noEmit`; run tests with `npx vitest run`; format with `npm run format`.
+- Attribute keys are fixed by the wire protocol and must be used verbatim: `lk.segment_id`,
+  `lk.transcribed_track_id`, `lk.transcription_final`, `lk.publish_on_behalf`. They are documented
+  in `src/room/attribute-typings.ts` (a generated file — do not edit it).
+- Reserved topic, verbatim: `lk.transcription`.
+- `RoomEvent.TranscriptionReceived` keeps its exact existing signature:
+  `(segments: TranscriptionSegment[], participant?: Participant, publication?: TrackPublication)`.
+  Do not change it, and do not add a new room event.
+- Managers take external state as constructor callbacks and never import `RTCEngine`
+  (see "Manager pattern" in `CLAUDE.md`).
+- `startTime`, `endTime` and `language` on synthesized segments stay `0n`, `0n` and `''` — exactly
+  what agents write into legacy packets today.
+- Every task ends with a commit.
+
+---
+
+### Task 1: Tap transcription streams with a manager event
+
+Today a text stream is delivered to at most one consumer: the application handler registered for its
+topic. The SDK needs to read `lk.transcription` itself **while still delivering it** to an
+application handler for the same topic, because components-js reads that topic directly for
+`lk.expression`.
+
+So `IncomingDataStreamManager` becomes a typed event emitter (the manager pattern in `CLAUDE.md`)
+and emits `transcriptionStreamArrived` whenever a text stream opens on the reserved transcription
+topic. Internally the text path generalizes to N consumers, each with its own independent
+`TextStreamReader`: the application handler, plus a synthetic consumer that feeds the event.
+
+The event is named for the *stream* arriving, not a transcription: it fires when the stream opens,
+before any text has been read, and deliberately avoids the legacy `transcriptionReceived` name.
+
+Byte streams are deliberately left single-consumer — nothing internal reads them.
+
+**Files:**
+- Create: `src/room/data-stream/incoming/events.ts`
+- Modify: `src/room/data-stream/constants.ts`
+- Modify: `src/room/data-stream/incoming/IncomingDataStreamManager.ts`
+- Test: `src/room/data-stream/incoming/IncomingDataStreamManager.test.ts`
+
+**Interfaces:**
+- Consumes: nothing.
+- Produces:
+  - `export const TRANSCRIPTION_TOPIC = 'lk.transcription'` from `src/room/data-stream/constants.ts`
+  - `export type EventTranscriptionStreamArrived = { reader: TextStreamReader; participantIdentity: string }`
+  - `export type IncomingDataStreamManagerCallbacks = { transcriptionStreamArrived: (event: EventTranscriptionStreamArrived) => void }`
+  - `IncomingDataStreamManager` extends `TypedEmitter<IncomingDataStreamManagerCallbacks>`, so
+    `manager.on('transcriptionStreamArrived', handler)` is available.
+
+- [ ] **Step 1: Write the failing tests**
+
+Add these imports to `src/room/data-stream/incoming/IncomingDataStreamManager.test.ts`:
+
+```typescript
+import { subscribeToEvents } from '../../../utils/subscribeToEvents';
+import { TRANSCRIPTION_TOPIC } from '../constants';
+import type { IncomingDataStreamManagerCallbacks } from './events';
+```
+
+Then add this block at the end of the top-level `describe('IncomingDataStreamManager', ...)`. The
+helpers `headerPacket`, `chunkPacket` and `trailerPacket` already exist at the top of that file —
+reuse them, do not redefine them. They send from `'alice'` and default to topic `'my-topic'`, which
+the `fields` argument overrides.
+
+```typescript
+  describe('Transcription stream events', () => {
+    it('should emit transcriptionStreamArrived alongside an application handler', async () => {
+      const manager = new IncomingDataStreamManager();
+      manager.setConnected(true);
+      const managerEvents = subscribeToEvents<IncomingDataStreamManagerCallbacks>(manager, [
+        'transcriptionStreamArrived',
+      ]);
+
+      const appReaders: Array<TextStreamReader> = [];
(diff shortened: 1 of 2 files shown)


src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -371,45 +402,56 @@ export default class IncomingDataStreamManager {
...
-          new DataStreamError(
-            `Encryption type mismatch for stream ${trailer.streamId}. Expected ${encryptionType}, got ${textBuffer.info.encryptionType}`,
-            DataStreamErrorReason.EncryptionTypeMismatch,
-          ),
+    const textBuffers = this.textStreamControllers.get(trailer.streamId);
+    if (textBuffers && textBuffers.length > 0) {
+      const [{ info: textInfo }] = textBuffers;
+      if (textInfo.encryptionType !== encryptionType) {
+        const error = new DataStreamError(
+          `Encryption type mismatch for stream ${trailer.streamId}. Expected ${encryptionType}, got ${textInfo.encryptionType}`,
```
1egoman: Should this be doing a `textBuffers.every((buffer) => buffer.info.encryptionType !== encryptionType)` in the if block?


```suggestion
    const textBuffers = this.textStreamControllers.get(trailer.streamId);
    if (textBuffers && textBuffers.length > 0) {
      if (textBuffers.every((buffer) => buffer.info.encryptionType !== encryptionType)) {
        const error = new DataStreamError(
          `Encryption type mismatch for stream ${trailer.streamId}. Expected ${encryptionType}, got ${textInfo.encryptionType}`,
```

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -371,45 +402,56 @@ export default class IncomingDataStreamManager {
...
+          DataStreamErrorReason.EncryptionTypeMismatch,
         );
+        for (const { controller } of textBuffers) {
+          controller.error(error);
+        }
       } else {
-        textBuffer.info.attributes = { ...textBuffer.info.attributes, ...trailer.attributes };
+        // Shared `info`, so this merge is visible to every consumer's reader.
+        textInfo.attributes = { ...textInfo.attributes, ...trailer.attributes };
```
1egoman: Same here, shouldn't this apply to all `textInfo`s across all `textBuffers`?

src/room/transcription/TranscriptionStreamConverter.ts
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * Attribute keys carried on `lk.transcription` stream headers and trailers. Declared in
+ * `src/room/attribute-typings.ts` (generated), repeated here as the literals the wire uses.
+ */
+const SEGMENT_ID_ATTRIBUTE = 'lk.segment_id';
+const TRANSCRIBED_TRACK_ID_ATTRIBUTE = 'lk.transcribed_track_id';
+const TRANSCRIPTION_FINAL_ATTRIBUTE = 'lk.transcription_final';
+
+/** Set on a participant that publishes on another participant's behalf (e.g. an avatar worker). */
+export const PUBLISH_ON_BEHALF_ATTRIBUTE = 'lk.publish_on_behalf';
```
1egoman: Move these constants to this file, maybe make them an enum? Also make them used in `TranscriptionAttributes` and `AgentAttributes` for keys: https://github.com/livekit/client-sdk-js/blob/e724996365baea029527b79b60a3d19758d558ff/src/room/attribute-typings.ts#L28

src/room/transcription/TranscriptionStreamConverter.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  private emitSegment(
+    partial: PartialTranscription,
+    segmentId: string,
+    senderIdentity: string,
+    reader: TextStreamReader,
+    final: boolean,
+  ) {
+    if (partial.emittedText === undefined && partial.text === '') {
```
1egoman: nitpick: Use `typeof partial.emittedText === 'undefined`

src/room/Room.ts
```diff
@@ -283,6 +288,17 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     this.incomingDataStreamManager = new IncomingDataStreamManager(
       this.options.dataStream?.maxPayloadByteLength,
     );
+    this.transcriptionStreamConverter = new TranscriptionStreamConverter({
+      onTranscription: (transcription) => this.handleTranscription(undefined, transcription),
```
1egoman: question: should this be `undefined` here? Where this was called previously it was `this.handleTranscription(participant, packet.value.value);` - do you need to inject in the participant value here too? If not, then consider removing the participant param all together.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -253,17 +263,31 @@ export default class IncomingDataStreamManager {
...
+        ) {
+          streamHandlerCallbacks.push((reader, { identity }) => {
+            this.emit('transcriptionStreamArrived', { reader, participantIdentity: identity });
+          });
+        }
+        const applicationCallback = this.textStreamHandlers.get(streamHeader.topic);
+        if (applicationCallback) {
+          streamHandlerCallbacks.push(applicationCallback);
+        }
+        if (streamHandlerCallbacks.length === 0) {
```
1egoman, line 293: I think it would be better if this worked differently, instead of `streamHandlerCallbacks` being an array, keep it how it was before and make a new private method like `emitSideChannelTranscriptionStreamEvent` or similar. Then everywhere that `streamHandlerCallbacks` is looped around can be dropped and kept how it is today plus a call after `streamHandlerCallback(...)` like `this.emitSideChannelTranscriptionStreamEvent(streamHeader, reader, identity)`. And then this new private method can contain:
```
if (
  streamHeader.topic === TRANSCRIPTION_TOPIC // Don't check listener count here
) {
  this.emit('transcriptionStreamArrived', { reader, participantIdentity: identity });
}
```

1egoman:
I've tested this with livekit/agents#7240 and have confirmed it works end to end - no packets which match `packet.value.case === "transcription"` are received when the agents framework is using the code in the pull request, and when running on `main` matching legacy transcription packets are received.

src/room/data-stream/incoming/IncomingDataStreamManager.ts
```diff
@@ -253,17 +273,31 @@ export default class IncomingDataStreamManager {
...
+        // drop it" behavior intact when no one has subscribed.
+        const streamHandlerCallbacks: Array<TextStreamHandler> = [];
+        if (
+          streamHeader.topic === TRANSCRIPTION_TOPIC &&
+          this.listenerCount('transcriptionStreamArrived') > 0
+        ) {
+          streamHandlerCallbacks.push((reader, { identity }) => {
+            this.emit('transcriptionStreamArrived', { reader, participantIdentity: identity });
+          });
+        }
```
1egoman, line 288: Note to reviewers - this "side channel" `transcriptionStreamArrived` event is how other subsystems tap off of `IncomingDataStreamManager` without requiring a whole separate subscription management system for internal subscribers.

src/room/transcription/TranscriptionStreamConverter.ts
```diff
@@ -0,0 +1,221 @@
...
+  constructor(options: TranscriptionStreamConverterOptions) {
+    this.options = options;
+  }
+
+  /** Drops all in-flight segment state. */
+  reset() {
+    this.partials.clear();
+  }
+
+  handleTextStream = async (reader: TextStreamReader, senderIdentity: string) => {
```
1egoman, line 64: Note to reviewers - this net new `TranscriptionStreamConverter` class which does the modern -> legacy transcription conversion is really the high risk part of this change, and is where the most scrutiny would be appreciated, especially from those who are familiar with the peculiars of how the legacy transcription system works.

src/room/Room.ts
```diff
@@ -284,6 +287,17 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
...
+      onTranscription: (transcription) => this.handleTranscription(transcription),
+      getMicrophoneTrackSid: this.getMicrophoneTrackSid,
+      getDelegatingPublisherIdentity: this.getDelegatingPublisherIdentity,
+    });
+    this.incomingDataStreamManager.on(
+      'transcriptionStreamArrived',
+      ({ reader, participantIdentity }) => {
+        this.transcriptionStreamConverter.handleTextStream(reader, participantIdentity);
+      },
+    );
```
1egoman, line 300: Note to reviewers - here is where everything is hooked together:
- The new `IncomingDataStreamManager` event (`transcriptionStreamArrived`) is wired up to `TranscriptionStreamConverter` via `this.transcriptionStreamConverter.handleTextStream(...)`
- `this.handleTranscription(...)` is called with the newly generated legacy transcriptions.

src/room/Room.ts
```diff
@@ -2100,7 +2115,11 @@ class Room extends (EventEmitter as new () => TypedEmitter<RoomEventCallbacks>)
     if (packet.value.case === 'user') {
       this.handleUserPacket(participant, packet.value.value, packet.kind, encryptionType);
     } else if (packet.value.case === 'transcription') {
-      this.handleTranscription(participant, packet.value.value);
+      // Legacy `Transcription` packets are ignored: transcription events are rebuilt from the
+      // `lk.transcription` data stream channel instead, which this client advertises support for
+      // via client protocol 3. See
+      // docs/superpowers/specs/2026-09-04-transcription-back-conversion-design.md
+      this.log.debug('ignoring legacy transcription data packet', this.logContext);
```
1egoman, line 2122: Note to self - drop `docs/superpowers/specs/2026-09-04-transcription-back-conversion-design.md` from this comment before merging

src/room/transcription/TranscriptionStreamConverter.ts
```diff
@@ -0,0 +1,221 @@
...
+        }
+        partial.text += unwrapTimedString(chunk);
+        this.emitSegment(partial, segmentId, senderIdentity, reader, this.isFinal(reader) ?? false);
+      }
+    } catch (err) {
+      // The stream ended abnormally (sender disconnected mid-segment, decode failure). Whatever
+      // was accumulated is all there will ever be, so close the segment out.
+      this.log.debug('lk.transcription stream ended abnormally', err);
+      this.closeSegment(key, segmentId, senderIdentity, reader, true);
+      return;
```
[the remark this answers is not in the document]
1egoman: I think this is fine to accept for now, fixing this is going to be hard and in practice this will just result in a surious message being sent in this one case so it's not going to cause like a fatal error.

src/room/Room.ts
lukasIO: I'm wondering whether it would make sense to gate all the conversion work on listener count of the legacy transcription event. 
Basically skipping all the conversion/work if there are no listeners registered for it 🤔

lukasIO (approved):

src/room/Room.ts
1egoman: Yea, good idea. I think this would be fairly easy to do!

1egoman: I did it locally and it surfaced a small edge case I hadn't considered - if you start listening for transcriptions half way through an in flight transcription being delivered, then the state within `TranscriptionStreamConverter` may not be fully up to date and it could lead to inconsistent output.

Because of this I think it makes sense to opt not to do this. @lukasIO if you disagree and think this is a worthy tradeoff, let me know and I can push up my work.

PR #2095, opened 2026-09-11
fix: cancel in-flight full reconnects when the engine closes

horvatz:
Calling `Room.disconnect()` while a full reconnect is in progress can leave the reconnect coroutine running after the engine has closed. When paused cleanup or region selection finishes, it can call `join` with the saved credentials. A late signal join response can also set `_isClosed` back to `false`. By then engine close has removed its listeners, so applications cannot reliably catch the late attempt through room-state events.

This change gives each reconnect attempt an abort controller, aborts it synchronously on engine close, and passes its signal through full reconnect joins and region selection. Cancelled continuations stop without retrying, and late join responses cannot reopen the engine. SignalClient also checks cancellation after waiting for its connection lock/connection parameters and after previous-transport teardown, before creating another transport.

### Reproduction and regression coverage

The original sequence is: start a full reconnect, pause peer-connection cleanup, await `Room.disconnect()`, then release cleanup. Before this change a late join is attempted; after it no join is attempted and the room/engine remain disconnected/closed.

Six new regression cases failed before the corresponding fixes and pass afterward:

- close during reconnect cleanup;
- close during region selection;
- a signal join response settling after close;
- an already-aborted signal before join;
- cancellation while waiting for the signal connection lock;
- cancellation during previous-transport teardown.

A successful full reconnect is covered as well. Tests exercise the actual engine reconnect/join and signal-client paths with network/peer-connection work stubbed; no live server is required. The original reproduction was also rerun through real `Room.disconnect()`.

### Validation

- Full unit suite: 819 passed, 1 skipped (51 files).
- Type check, ESLint, Prettier, throws check, build, browser compatibility check, and package validation commands pass. ESLint reports 13 existing warnings outside the changed files; compatibility and package validation also report warnings.
- Includes a patch changeset.

### DIFF
--- a/src/api/SignalClient.ts
+++ b/src/api/SignalClient.ts
@@ -443,6 +443,11 @@ export class SignalClient {
     const rtcUrl = createRtcUrl(url, params, useV0Path).toString();
     const validateUrl = createValidateUrl(rtcUrl).toString();
 
+    if (abortSignal?.aborted) {
+      unlock();
+      throw ConnectionError.cancelled('Connection aborted');
+    }
+
     return new Promise<JoinResponse | ReconnectResponse | undefined>(async (resolve, reject) => {
       try {
         let alreadyAborted = false;
@@ -497,6 +502,7 @@ export class SignalClient {
           await this.teardownTransport('replaced by a new connection attempt');
           this.log.debug(`closed previous ws connection in ${performance.now() - startClose}ms`);
         }
+        if (alreadyAborted) return;
 
         // the transport created below belongs to this attempt; events arriving from it after a
         // newer attempt has started are dropped by the machine
--- a/src/room/RTCEngine.ts
+++ b/src/room/RTCEngine.ts
@@ -227,6 +227,8 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
 
   private attemptingReconnect: boolean = false;
 
+  private reconnectAbortController?: AbortController;
+
   private reconnectPolicy: ReconnectPolicy;
 
   private reconnectTimeout?: ReturnType<typeof setTimeout>;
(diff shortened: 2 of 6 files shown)


src/api/SignalClient.ts
```diff
@@ -497,6 +502,7 @@ export class SignalClient {
           await this.teardownTransport('replaced by a new connection attempt');
           this.log.debug(`closed previous ws connection in ${performance.now() - startClose}ms`);
         }
+        if (alreadyAborted) return;
```
1egoman, line 506: question: Are there other places further down in this method where something like this could be needed as well? What happens if an abort gets triggered midway through the signaling websocket connection process?

src/room/RTCEngine.ts
```diff
@@ -463,6 +471,7 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
    * having given up on reconnecting.
    */
   async close(reason?: string) {
+    this.reconnectAbortController?.abort();
```
1egoman, line 474: After aborting this here, I think it might be a good idea to reset `this.reconnectAbortController` back to its initial `undefined` value (like you are doing in the `finally` in `attemptReconnect`). While in practice I think this isn't critical for the change today, it means that any future code will be able to use the existence of `this.reconnectAbortController` as a meaningful signal.

src/room/RTCEngine.ts
```diff
@@ -1335,16 +1344,20 @@ export default class RTCEngine extends (EventEmitter as new () => TypedEventEmit
...
-        await this.restartConnection();
+        await this.restartConnection(abortController.signal);
       } else {
         await this.resumeConnection(reason);
       }
+      if (abortController.signal.aborted) return;
       this.clearPendingReconnect();
       succeeded = true;
     } catch (e) {
+      if (abortController.signal.aborted) return;
```
1egoman, line 1360: For all of these types of bail outs throughout the change, I wonder if it would be a good idea to add some sort of log for each one so that a user debugging a connection failure would be able to tell a) there was a purposeful bailout midway through, and b) which one bailout was taken. Thoughts on this?

1egoman (commented):
Thanks for the PR - this looks like a really promising new addition. Left a few small comments, other than those this looks good to me.

1egoman:
Ah @horvatz would you be able to click the button in [this comment](https://github.com/livekit/client-sdk-js/pull/2095#issuecomment-5635641372) and sign the CLA? That is a prerequisite before this could be merged.

PR #2098, opened 2026-09-14
fix: fall back to default camera on TrackEnded, matching audio behavior

### LINKED ISSUE #2035: Video track does not fall back to default device on TrackEnded, unlike audio
### Describe the bug

In `LocalParticipant.handleTrackEnded`, when a device disconnects (e.g. an external webcam is unplugged) and the corresponding `MediaStreamTrack` fires `ended`, the SDK explicitly falls back to the default device for **audio** tracks, but not for **video** tracks:

https://github.com/livekit/client-sdk-js/blob/v2.21.0/src/room/participant/LocalParticipant.ts#L2126-L2132

```ts
if (isLocalAudioTrack(track)) {
  // fall back to default device if available
  await track.restartTrack({ deviceId: 'default' });
} else {
  await track.restartTrack();
}
```

For video, `restartTrack()` is called with no override, so it retries against the *same* (now-disconnected) `deviceId` constraint, fails again, and the surrounding `catch` mutes the track instead of falling back:

```ts
} catch (e) {
  this.log.warn(`could not restart track, muting instead`, getLogContextFromTrack(track));
  await track.mute();
}
```

The result: unplugging an external microphone falls back seamlessly to the default mic, but unplugging an external webcam just turns the camera off, even though a built-in/default camera is available.

I confirmed this is still present, unchanged, on the latest release (v2.21.0) — same `if (isLocalAudioTrack(track))` check, no equivalent branch for `isLocalVideoTrack(track)`.

### Reproduction

1. Join a room with an external webcam selected as the camera source.
2. Physically disconnect the webcam.
3. Observe: the local video track ends and gets muted (camera turns off), instead of falling back to the default/built-in camera the way an external microphone falls back to the default mic in the same scenario.

### Expected behavior

When a video track ends due to a device disconnect (and the track is not user-muted), it should retry `restartTrack({ deviceId: 'default' })`, mirroring the existing audio behavior, so the camera falls back to the default device instead of going dark.

### Suggested fix

```ts
if (isLocalAudioTrack(track) || isLocalVideoTrack(track)) {
  // fall back to default device if available
  await track.restartTrack({ deviceId: 'default' });
}
```

(Since this branch is only reached when `isLocalAudioTrack(track) || isLocalVideoTrack(track)` is already true per the outer `else if`, this can also simply be unconditional.)

### Reproduction repo/live example

N/A — traced directly in the SDK source while debugging camera-off behavior in our app; happy to share more details if useful.

### SDK version

`livekit-client` 2.15.14 through 2.21.0 (latest at time of filing)

hardope:
LocalParticipant.handleTrackEnded already falls back to the default microphone
(deviceId: 'default') when an audio track's underlying MediaStreamTrack ends
unexpectedly (e.g. the device is unplugged). The video branch instead called
restartTrack() with no options, which reuses the previous constraints,
including the deviceId of the device that just disappeared. getUserMedia then
fails again with the same error and the track is muted instead of switching
to another camera.

This applies the same deviceId: 'default' fallback to video, so camera swaps
behave the same as microphone swaps.

Fixes #2035

Test plan:
- Added a regression test in LocalParticipant.test.ts asserting restartTrack
  is called with { deviceId: 'default' } for both video and audio tracks on
  handleTrackEnded. Verified it fails without the fix.
- npx tsc --noEmit
- npx eslint src
- npx vitest run (821 passed, 1 pre-existing unrelated failure in
  smoke-tests due to a missing separate install, confirmed present on
  main before this change)


src/room/participant/LocalParticipant.ts
```diff
@@ -2139,11 +2139,16 @@ export default class LocalParticipant extends Participant {
...
-            // fall back to default device if available
-            await track.restartTrack({ deviceId: 'default' });
+          // fall back to default device if available
+          if (isLocalVideoTrack(track)) {
+            // keep the previously requested resolution/frame rate/facing mode,
+            // only swap out the (now gone) deviceId
+            await track.restartTrack({
+              ...track.constraints,
+              deviceId: 'default',
+            } as VideoCaptureOptions);
```
1egoman, line 2149: I will defer to @lukasIO and @davidzhao here (as I am not as familiar with the historical context), but my understanding is the decision was purposefully made to NOT switch back to video in the same way as audio, because it could startle a user and cause a camera they didn't expect to be live to suddenly be published.

Because of this, I'm not sure that this change makes sense in its current state.

1egoman (changes requested):

PR #2102, opened 2026-09-14
Write the video start bitrate hint as one connection-level value, once

xianshijing-lk:
Follow-up to #1987, bringing the JS start-bitrate munging in line with what came out of the review on [client-sdk-android#973](https://github.com/livekit/client-sdk-android/pull/973). No change to the formula: still 90% of the target, capped at 1 Mbps for camera, uncapped for screen share.

## `x-google-start-bitrate` is connection-scoped, not per-section

libwebrtc *reads* it per m-section but doesn't *apply* it per m-section:

```cpp
bitrate_config_ = GetBitrateConfigForCodec(send_codec()->codec);   // webrtc_video_engine.cc:1337
...
call_->GetTransportControllerSend()->SetSdpBitrateParameters(bitrate_config_);  // :1358
```

That lands in `RtpBitrateConfigurator`, which holds **one** `BitrateConstraints` for the entire `RTCPeerConnection`. Every video m-section writes the same slot, last writer wins, and the order is just m-section order.

Today each track writes its own value. The camera/screen-share split is what makes that bite: camera capped at 1000, screen share uncapped at e.g. 2700, both in the same offer. Whichever section applies last seeds the single estimator — so either the screen share overrides the camera cap for the whole connection, or the camera clobbers the screen share and the exemption does nothing. It depends on SDP layout, so it looks stable in testing and flips in the field.

**Fix:** `computeConnectionStartBitrate` takes the largest hint among the video m-sections that map to a published track, and every video section gets that same number. Only sections present in the current SDP count, since `trackBitrates` is append-only and outlives an unpublish.

## Written once per connection, not on every offer

libwebrtc guards re-application three ways: it only re-reads the fmtp when the **send codec changes** (`webrtc_video_engine.cc:1338-1342`), ignores a value equal to the stored one (`rtp_bitrate_configurator.cc:66-73` — *"setting the same remote description twice shouldn't restart bandwidth estimation"*), and returns `-1` for "no change" while retaining the real value.

The gap: when the value **changes** between offers and the codec changed too, it really does restart a converged estimator. With per-track values, publishing a screen share mid-session changes the last-writer value.

The reason once is *sufficient*, not just safe: the seed persists. `RtpBitrateConfigurator` keeps `start_bitrate_bps`, and `RtpTransportControllerSend::OnNetworkRouteChanged` re-applies it from `GetConfig()` (`rtp_transport_controller_send.cc:390`). A WiFi→cellular handover re-seeds the estimator from this hint automatically, with no renegotiation. A full reconnect builds a new peer connection and seeds the new estimator again.

`hasAppliedVideoStartBitrate` latches only after `setMungedSDP` accepts the offer, so a rejected munge retries on the next one.

## 300 kbps target floor

Matches the Rust SDK. Below that, seeding above the real capacity costs more than the ramp it saves.

## Refactor note

`applyVideoStartBitrate` was doing double duty as section-matcher, and the DD-extension munging depended on that match. Matching moved to `findTrackCodecPayload`, so DD extension still runs on **every** offer while the bitrate write happens once.

## Tests

`PCTransport.test.ts`: 28 passing (6 new — the floor, the cap and its screen-share exemption, the connection-level max, stale-entry filtering, and section matching).

All libwebrtc references are against `m144_release`.

> [!NOTE]
> 13 test files plus `lint` and `tsc` fail on this branch from a missing `machina` package — reproduced on a clean checkout of `main`, unrelated to this change. All 4 `tsc` errors are in `SignalClient*.ts`.


1egoman (approved):
High level makes sense, I think @lukasIO once you are back it would be good for you to take a look at this one post merge as I am less familiar with this part of the sdk.

PR #2108, opened 2026-09-18
Catch the negotiate() rejection on a media-sections requirement

SergeAx:
`RTCEngine` has three `this.negotiate()` call sites. Two of them attach the same handler:

```ts
this.negotiate().catch((err) => {
  this.log.error(err);
});
```

The third — the one wired to `client.onMediaSectionsRequirement` — does not:

https://github.com/livekit/client-sdk-js/blob/main/src/room/RTCEngine.ts#L756-L759

So when a renegotiation triggered by the server's media-sections requirement fails, the rejection goes nowhere. Instead of a logged error the application gets an unhandled promise rejection, and because `negotiate()` rejects from a detached promise there is no stack and no context attached to it.

We hit this on React Native (`livekit-client` 2.20.0). It reaches Sentry as a bare `onunhandledrejection` — in our case `Error: Session error code: ERROR_CONTENT. Session error description: Failed to apply the description for m= section with mid='0': Local fingerprint does not match identity…` — grouped as its own issue, disconnected from the `NegotiationError` that the same failure also produces through the caught paths. That made it look like two unrelated faults during triage.

This does not change behaviour: `negotiate()` already calls `handleDisconnect('negotiation', ReconnectReason.RR_UNKNOWN)` before it rejects, so the engine's own recovery runs either way. The patch only routes the error into the logger the other two call sites already use.

Still present on `main` and in 2.22.3.

### Checks

- `npx tsc --noEmit` — clean
- `npx vitest run src/room` — 593 passed
- `npx prettier --check` — clean
- Changeset included (patch)


SergeAx:
This patch is written by an agent (Claude Code @ Opus 5.1 xhigh) and verified by me, a human. Claude's vanity trailer in the commit message is omitted per my local settings.

1egoman (approved):
Makes sense to me, thanks for the contribution!

Issue #2107, opened 2026-09-17
waitForDimensions skips the first-frame wait on iPadOS because getBrowser() reports macOS (follow-up to #2099 / #2100)


lagudafuadtosin:
### What I'm expecting

After #2100, `LocalTrack.waitForDimensions` waits for the first video frame before trusting `getSettings()`, so a portrait front-camera capture on a WebKit device reports portrait dimensions. I would expect that to hold on an iPad as well as an iPhone.

### What happens instead

The wait is gated on `getBrowser()?.os === 'iOS'`. `browserParser.ts` (line 66) decides `iOS` versus `macOS` for Safari by whether the user agent contains `mobile/`. iPadOS has sent the Macintosh desktop user agent by default since iPadOS 13, so on an iPad Safari `getBrowser()` returns `os: 'macOS'` and `waitForDimensions` takes the old path: it reads `getSettings()` immediately, with no wait at all (the 10 ms sleep that #2100 replaced was also iOS-gated, so iPads never had even that).

The front camera on an iPad is mounted the same way as on an iPhone and runs the same WebKit capture pipeline, so I would expect the same window where the track reports the sensor frame until the first painted frame. I raised this in a review comment on #2100 but it was outside that PR's scope, so that is why I am opening it here.

I do not have an iPad to measure the window on, so the size of it on iPad is an expectation, not a reading. The gap in the gate is not an expectation, it is in the parser.

### Two shapes, and I will send a pull request for whichever you prefer

1. **Drop the OS gate.** Call `waitForFirstVideoFrame` on every platform. It is best effort and never throws, it reuses an attached preview element when one exists, and on desktop the first frame arrives within a frame or two, so the cost where the bug does not exist is a few tens of milliseconds on publish. Simplest, and it covers any other WebKit-based capture path that reports the same way.

2. **Detect iPadOS.** Treat Safari on a Macintosh user agent with `navigator.maxTouchPoints > 1` as iOS for this check, either inside `browserParser` (which would change `getBrowser().os` for every other iOS check in the SDK, so probably not) or as a local helper used only by `waitForDimensions`. Narrower, no change on desktop.

My preference is 1, because the wait is cheap wherever frames arrive fast and it removes a UA sniff that will drift again. Happy to do 2 if you would rather keep desktop publish timing untouched.

Either way the PR would come with a unit test in `utils.test.ts` covering the new path.

### Reproduction

Code path, at merge commit e55c21f (main after #2100):

[...]

Issue #1788, opened 2026-01-06
iPad Not Detected as iOS/iPadOS When Using Safari


TomFanella4:
### Describe the bug

LiveKit’s `isSafari()` does not detect iPad in some cases since modern iPads can present as a Mac. To get around this, we identify Safari-based browsers using feature detection instead of relying only on UA based checks.

What we do:
```javascript
export function getIsIOS() {
  const ua = navigator.userAgent;

  // Check for iOS devices (iPhone, iPod touch)
  const isIOS = /iPhone|iPod/i.test(ua);

  // Check for iPad (including iPadOS 13+ which identifies as Mac)
  const isIPad =
    /iPad/i.test(ua) ||
    (/\bMacintosh\b/i.test(ua) && navigator.maxTouchPoints > 1);

  return isIOS || isIPad;
}
```

What LiveKit checks for:
https://github.com/livekit/client-sdk-js/blob/4f126419a1f0c96262cacb74dd8d9d611f4839f5/src/utils/browserParser.ts#L59-L72

https://github.com/livekit/client-sdk-js/blob/4f126419a1f0c96262cacb74dd8d9d611f4839f5/src/room/utils.ts#L181-L188

### Reproduction

Run the example functions and view the expected results.

### Logs

```shell

```

### System Info

```shell
iPad, Safari
```

### Severity

annoyance

lukasIO:
hi @TomFanella4 thanks for the report. 

I believe 
```
  (/\bMacintosh\b/i.test(ua) && navigator.maxTouchPoints > 1);
```

would also be true for MacOS with a touch screen connected to it?

1egoman:
I did some investigation for an unrelated issue (https://github.com/livekit/client-sdk-js/issues/2107) and I think determined that MacOS Safari never in practice returns a `navigator.maxTouchPoints` value over 1, since MacOS Safari (as of September 2026) [doesn't implement the touch events api](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events).

I think given this `navigator.maxTouchPoints > 1` seems to be a fairly good discriminator (along with verifying a mac user agent) that the device is an ipad.

PR #2110, opened 2026-09-18
Wait for the first video frame on iPadOS in waitForDimensions

1egoman:
Updates the check from #2100 to also wait for the first video frame to go through the capture pipeline on iPads (as well as iPhones) as iPads suffer from the same issue.

More info / context can be found in https://github.com/livekit/client-sdk-js/issues/2107#issuecomment-5731779105.


src/room/utils.ts
```diff
@@ -300,6 +300,33 @@ export function isSafariBased(): boolean {
...
+}
+
+/**
+ * iPhone, iPad and iOS-hosted browsers — everything running the Apple camera capture pipeline,
+ * including iPads that present themselves as a Mac.
+ */
+export function isAppleMobile(): boolean {
+  return getBrowser()?.os === 'iOS' || isIPadOS();
+}
+
```
1egoman, line 329: This is the highest risk part of this change.

As a datapoint, I have an older ipad 6 and iphone 11 kicking around along with my mac - I ran the above functions on all three platforms and `isIPadOS()` returned true only for the ipad, and `isAppleMobile()` returned true for both the ipad and iphone.

Related issue: https://github.com/livekit/client-sdk-js/issues/1788. Not 100% sure this handles all cases right though and would like another perspective before merging.

xianshijing-lk (approved):

src/room/utils.ts
xianshijing-lk: btw, from some quick chatgpt prompt (I didn't verify it) : 
The tricky part is iPadOS 13+: Safari can deliberately identify itself as macOS, so checking only os === 'iOS' misses modern iPads. This heuristic:

b?.os === 'macOS' && navigator.maxTouchPoints > 1

is the standard workaround. Your additional b?.name === 'Safari' restriction, however, may be unnecessarily narrow depending on what you're trying to detect.

If your actual goal is “anything on iPhone/iPad that uses Apple's iOS/iPadOS browser/media stack”, I'd prefer:

```
export function isIPadOS(): boolean {
  if (!isWeb()) {
    return false;
  }

  const b = getBrowser();
  return b?.os === 'macOS' && navigator.maxTouchPoints > 1;
}

/**
 * iPhone/iPad browsers, including iPads that identify themselves as macOS.
 */
export function isAppleMobile(): boolean {
  const os = getBrowser()?.os;
  return os === 'iOS' || isIPadOS();
}
```

The reason I'd consider removing b?.name === 'Safari' is important: Chrome, Firefox, Edge, etc. on iOS/iPadOS are also using Apple's browser engine/capture stack. If you're making a WebRTC/camera workaround, you probably care about the underlying Apple platform rather than whether the browser calls itself Safari.


what do you think about it ?

davidzhao (approved):
helpful context. lgtm!

1egoman:
Thanks folks, I will review with lukas once he's back as well.

PR #2111, opened 2026-09-21
Fix track volume losing effect when set to 0

robintown:
If you call `RemoteAudioTrack.setVolume` with a volume of 0, the track mutes initially but becomes unmuted if it is ever reattached to a new audio element. Looks like this is just a truthiness slip-up.


1egoman (approved):
Makes sense, thanks for the contribution!

PR #2112, opened 2026-09-21
Remove fatal option from TextDecoder

JoelTowell:
## Summary
Last month I removed the `fatal: true` option passed to `TextDecoder` in [this PR](https://github.com/livekit/client-sdk-js/pull/2037), due to issues with React Native that I outlined in that description. 

_Update: Turns out I had a stale copy and only forgot to remove it from the test_

## Changes
I removed the `fatal: true` option being passed to `TextDecoder` in `src/room/data-stream/outgoing/OutgoingDataStreamManager.test.ts`.


1egoman:
Hi @JoelTowell - from what I can tell, there is only one instance of `fatal: true` on `main` of the upstream repo, which is in the tests: https://github.com/search?q=repo%3Alivekit%2Fclient-sdk-js+%22fatal%3A+true%22&type=code.

From what I can tell, it looks like your PR was based on an outdated version of the upstream `main` (f2e37acf258c348f59d6d1650a5f48bedff83700, from mid august 2026, from before your previous PR) and the merge this would result in would be a clean three way merge which is why it looks like a valid diff in github. In practice, I think it would be largely no-op.

The one meaningful change this does have is that one entry which shows up in the github search from the first link - this line in the tests:
```diff
-      const decode = (bytes: Uint8Array) => new TextDecoder('utf-8', { fatal: true }).decode(bytes);
+      const decode = (bytes: Uint8Array) => new TextDecoder('utf-8').decode(bytes);
```

I think that might have been missed in the original PR and probably wouldn't be a bad idea to also migrate, but this code should never be running as part of the released SDK so I don't think it should matter in your case.

Let me know if I'm missing something here. If `fatal: true` was indeed re-added, that was an oversight and not intended.

JoelTowell:
> Hi @JoelTowell - from what I can tell, there is only one instance of `fatal: true` on `main` of the upstream repo, which is in the tests: https://github.com/search?q=repo%3Alivekit%2Fclient-sdk-js+%22fatal%3A+true%22&type=code.
> 
> From what I can tell, it looks like your PR was based on an outdated version of the upstream `main` ([f2e37ac](https://github.com/livekit/client-sdk-js/commit/f2e37acf258c348f59d6d1650a5f48bedff83700), from mid august 2026, from before your previous PR) and the merge this would result in would be a clean three way merge which is why it looks like a valid diff in github. In practice, I think it would be largely no-op.
> 
> The one meaningful change this does have is that one entry which shows up in the github search from the first link - this line in the tests:
> 
> ```diff
> -      const decode = (bytes: Uint8Array) => new TextDecoder('utf-8', { fatal: true }).decode(bytes);
> +      const decode = (bytes: Uint8Array) => new TextDecoder('utf-8').decode(bytes);
> ```
> 
> I think that might have been missed in the original PR and probably wouldn't be a bad idea to also migrate, but this code should never be running as part of the released SDK so I don't think it should matter in your case.
> 
> Let me know if I'm missing something here. If `fatal: true` was indeed re-added, that was an oversight and not intended.

Yep, you're right. I did hit this in an app, but looking back I was evidently running an older version. I then managed to confuse myself further by fetching upstream but branching from my stale local `main`, so I thought the fix had somehow been reverted from `main`. 🤦

Sorry for the noise. I'll trim this down to just the test change.

1egoman (approved):
No problem, thanks for updating the test, still definitely valuable!

Issue #2116, opened 2026-09-23
Firefox 155+ subscribers receive AV1 SVC they cannot decode — subscriber side needs the same AV1 gate as publishers (black/frozen high spatial layers)


ham-c4b:
### Describe the bug

Firefox enabled AV1 for WebRTC by default in stable about a week ago (156.0, 2026-09-15;
  https://bugzilla.mozilla.org/show_bug.cgi?id=1944878), but still does not support SVC
  (https://bugzilla.mozilla.org/show_bug.cgi?id=1571470).

  The SDK already gates AV1 for Firefox *publishers*: supportsAV1() in src/room/utils.ts returns
  false for Firefox ("Firefox does support AV1, but SVC publishing is not supported", #1605 /
  #1631, v2.15.6). There is no equivalent on the *subscriber* side — verified against current
  main: supportsAV1() is only consumed on the publish path in LocalParticipant, and the
  subscriber answer path (PCTransport) does no codec filtering (no setCodecPreferences anywhere
  in src/). That was fine while Firefox never offered AV1 in the receive direction — the SFU
  served such subscribers the backup codec.

  Since Firefox 155/156 that assumption no longer holds: when a Chrome participant publishes
  videoCodec 'av1' (auto-injected scalabilityMode 'L3T3_KEY') with backupCodec { codec: 'vp8' },
  a Firefox 156 subscriber now negotiates AV1, so the SFU forwards the AV1 SVC stream instead of
  the VP8 backup. SVC is not visible in SDP, so Firefox cannot refuse it. Firefox then only
  decodes the base spatial layer reliably: small tiles play, any subscription selecting a higher
  spatial layer (large elements with adaptiveStream) renders black/frozen.

  Bisected on the receiver:
  - independent of hardware decoding (persists with media.hardware-video-decoding.enabled=false)
  - disabling Firefox's WebRTC AV1 pref (forcing the VP8 backup) fully fixes it; re-enabling
    immediately reproduces the black tile.

 **Expected behavior:**
  Firefox subscribers should receive the backup codec (as they effectively did before FF 155) —
  or generally: AV1 SVC should not be forwarded to a subscriber that cannot decode SVC.

**Suggested fix:**
  Mirror the publish-side gate on the receive path: exclude AV1 from Firefox subscribers'
  negotiated codecs (filter the subscriber answer / setCodecPreferences when isFireFox()), so
[...]

PR #2118, opened 2026-09-23
feat: add videoReceiveCodecFilter room option

### LINKED ISSUE #2117: Add a receive-side video codec filter so subscribers can exclude codecs and get the backup codec
### Describe the problem

LiveKit currently exposes publish-side codec configuration, including `videoCodec`, `backupCodec`, and `backupCodecPolicy`. However, applications do not appear to have a public API to control which video codecs are allowed on the receiving side of the WebRTC connection. This creates a problem for browsers that advertise support for a codec even when decoding it is only software-based or otherwise undesirable for the application.

For example, a browser may report:

```ts
{
  supported: true,
  smooth: true,
  powerEfficient: false,
}
```

for AV1. WebRTC may still advertise AV1 as a supported receive codec, so LiveKit considers the subscriber compatible with AV1 and may select AV1 instead of the configured H264 backup codec.

An application may want to apply a receive-side codec policy such as:

- allow AV1 only when it is efficiently decoded;
- exclude AV1 on specific browsers or devices;
- prefer H264 on battery-powered devices;
- exclude a codec because of a known browser or hardware issue.

The application should be able to make this decision without modifying `livekit-client` or intercepting/manipulating SDP.

Assume that a publisher provides:

```text
Primary codec: AV1
Backup codec: H264
```

If a subscriber excludes AV1 from its receive codecs, LiveKit should treat that subscriber as incompatible with AV1 and select the H264 backup codec when available.

Expected flow:

```text
Application determines that AV1 should not be received
        ↓
AV1 is excluded from the subscriber's receive capabilities
        ↓
LiveKit detects that the subscriber cannot receive AV1
        ↓
The configured backup codec can be selected
```

### Proposed API

A callback-based filter on `RoomOptions`:

```ts
new Room({
  videoReceiveCodecFilter: (codec: RTCRtpCodec) => {
    return codec.mimeType.toLowerCase() !== 'video/av1';
  },
});
```

A callback receives the full `RTCRtpCodec` (including `sdpFmtpLine`), so it can also target a specific H264 profile. A simple allow-list such as
`['video/H264', 'video/VP8']` is a one-liner on top of it.

The important requirements are:

- let the application decide which codecs are allowed;
- apply the policy to receive-side video transceivers;
- preserve the policy across renegotiations and reconnects;
- work with the existing backup codec mechanism;
- avoid requiring application-level SDP manipulation.

Only "allowed" is exposed, not "preferred": the SFU picks among the publisher's codecs in the publisher's order (primary, then backup), so receive-side ordering would have no effect.

### Notes

This feature should not necessarily make `MediaCapabilities.powerEfficient` a LiveKit-level rule. `powerEfficient` is only one possible input to an
application-specific policy.

The SDK could expose a generic receive-codec filtering mechanism, while the application remains responsible for deciding which codecs should be allowed.

### Alternatives considered

- Changing `publishDefaults.videoCodec`: this only affects the local participant's published video and does not control codecs received from
  other participants.
- Configuring `backupCodec`: this provides a fallback published by the publisher, but does not allow the subscriber to exclude a codec that the
  browser reports as supported.
- Application-level SDP interception: technically possible in some setups, but fragile and dependent on internal SDK behavior.

### Importance

would make my life easier

### Additional Information

- #2116: Firefox 156+ subscribers negotiate AV1 but cannot decode SVC. The root cause is the same: the subscriber side does no codec filtering. A   receive-side codec filter would give applications an immediate workaround, and could be reused internally for a Firefox default.

I have a working implementation on my fork: M-Blobby/client-sdk-js@feat/video-receive-codec-filter

---

Jitsi exposes this as configuration: `videoQuality.codecPreferenceOrder` and `mobileCodecPreferenceOrder` (e.g. `['VP8', 'VP9', 'H264', 'AV1']` on mobile). In lib-jitsi-meet, the effective codec list is built from the configured list filtered by browser support, so a codec left out of it is not used, and senders pick among the codecs supported by all remote endpoints. The separate mobile list covers the battery-powered device use case above. Jitsi's model is close to LiveKit's default `backupCodecPolicy`: one endpoint that does not accept a codec moves the sender to another codec for everyone. 

Maybe their approach is a better fit than a callback: a declarative list is simpler to configure, while a callback is more flexible (it can use `MediaCapabilities` or target a specific H264 profile). I'm open to either direction, or to offering both.

M-Blobby:
Description :
Closes #2117 
Related: #2116

## Summary

Adds a `videoReceiveCodecFilter` room option that lets applications exclude video codecs on the receive side. An excluded codec is left out of the subscriber's negotiation, so the server treats the subscriber as unable to receive it and delivers the publisher's backup codec instead.

```ts
new Room({
  videoReceiveCodecFilter: (codec) => codec.mimeType.toLowerCase() !== 'video/av1',
});
```

## How it works

- The filter restricts `RTCRtpTransceiver.setCodecPreferences` on receiving video transceivers. There is no SDP munging.
- It is applied in both negotiation directions:
  - **single peer connection** (default): on the recvonly video sections of the publisher connection, including those added later from `MediaSectionsRequirement`;
  - **dual peer connection**: between applying the server offer and creating
    the answer.
- `PCTransportManager` owns it. It computes the allowed codec list once and applies it once per transceiver (tracked in a `WeakSet`), since codec preferences persist across renegotiations. A full reconnect builds a new manager, so the filter carries over.
- Retransmission and FEC codecs (rtx, red, ulpfec, flexfec) are always kept.
- A filter that would exclude every codec is ignored with a warning, because `setCodecPreferences([])` resets to the browser defaults.
- It is a no-op where `setCodecPreferences` is unavailable, and a rejected call is caught and logged instead of failing the connection.
- Nothing changes on the publish side: publishers already publish their backup codec on demand from `subscribedCodecs`.

## Interaction with `backupCodecPolicy`

With the publisher's default policy (`PREFER_REGRESSION`), a single subscriber excluding the primary codec regresses the track to the backup codec for every subscriber. With `BackupCodecPolicy.SIMULCAST`, each subscriber keeps its best allowed codec. This is documented in the option's JSDoc.

## Testing

- Unit tests for the allowed-list computation and for both injection points (order relative to `setRemoteDescription` / `createAndSetAnswer`, applied once per transceiver, publisher-only vs dual mode, rejected preferences).
- Manual test on LiveKit Cloud with Chrome, publisher on AV1 with H264 backup, one subscriber excluding AV1 and one without a filter:
  - default policy: both subscribers received H264;
  - `SIMULCAST`: the excluding subscriber received H264, the other kept AV1.
- The demo app has a new "Exclude AV1 on receive" toggle to reproduce this.


1egoman:
@lukasIO I would be curious your take on this in general. Is this you think the right layer for something like this to exist, or would it make more sense for the SFU to be configurable to not advertise support for certain codecs in the offer / answer SDP?

Also, if this is a path we end up going down, I would think it would be good for this to work with audio as well as video.

M-Blobby:
@1egoman Thanks for the review!

I put this on the client because the decision is per-receiver: a device without hardware AV1 decoding should get H264 while others keep AV1, and only the client knows its decoding capabilities. A SFU-level setting would apply to the whole room. The two could coexist though.

On audio: the backup codec mechanism is video-only, so there's no fallback to trigger. I could rename it to `receiveCodecFilter: (codec, kind) => boolean` to keep the door open, if you prefer.

